PR #1, opened 2017-10-16
Add Existing wc-api-dev Code

timmyc:
This is the initial commit of sorts which copies over existing code from `wc-api-dev`. I have added checks for the prior function names from `wc-api-dev` in hopes of not doubling up on hooks/filter logic while we migrate the hotfixes and such to this new plugin.

Ideally I'd like to keep changes here to a minimum, ship a 0.1.0, then circle back with additional PRs to update other items we have mentioned in the original issue.

For https://github.com/Automattic/wp-calypso/issues/18603


hotfixes/wc-calypso-bridge-mailchimp-no-redirect.php
```diff
@@ -0,0 +1,26 @@
+<?php
+/**
+ * add_settings_error is only available in UI contecext
```
justinshreve: I think this brought over the wrong file: https://github.com/woocommerce/wc-api-dev/blob/master/hotfixes/wc-api-dev-mailchimp-no-redirect.php

timmyc:
> Do we want to consider calling these something else while we have the chance?

Interesting to hear the background on the name! I would be fine with dropping all the hotfix files into the `inc` folder if that works for you.

I will bring over the MailChimp API code too.

justinshreve:
>  I would be fine with dropping all the hotfix files into the inc folder if that works for you.

Yeah, let's do that.

timmyc:
@justinshreve I moved all the hotfixes to `inc` and added the MailChimp API files as well. On my local vvv instance, with `MailChimp_Woocommerce` installed/activated - I can see all the endpoints as expected.

justinshreve:
@timmyc I think that was the eventually plan for it. I know we also have some feedback about how they do a few things WP wise as well. Something we'll loop back to once it launches.

PR #4, opened 2017-10-23
Hide install / setup notice.

timmyc:
For https://github.com/Automattic/wp-calypso/issues/16627

I am still seeing the setup your store notice on a few test sites - seems like it may be shown again after an upgrade possibly - regardless, adding in some logic to ensure it is not shown to Store users.

__To Test__
- Setup a new store
- Visit `wp-admin`, and verify the following notice is not shown:

edit_plugins_ _save_the_deschutes_ _wordpress

### DIFF
--- a/inc/wc-calypso-bridge-hide-alerts.php
+++ b/inc/wc-calypso-bridge-hide-alerts.php
@@ -14,12 +14,12 @@ add_filter( 'pre_option_wc_stripe_show_apple_pay_notice', '__return_true' );
 add_filter( 'pre_option_wc_stripe_show_request_api_notice', '__return_true' );
 
 // Hide setup store notice.
-add_filter( 'woocommerce_show_admin_notice', 'wc_calypso_bridge_hide_admin_notice' );
+add_filter( 'woocommerce_show_admin_notice', 'wc_calypso_bridge_hide_admin_notice', 10, 2 );
 
-function wc_calypso_bridge_hide_admin_notice( $notice ) {
-	if ( 'install' == $notice ) {
+function wc_calypso_bridge_hide_admin_notice( $bool, $notice ) {
+	if ( 'install' === $notice ) {
 		return false;
 	}
 
-	return true;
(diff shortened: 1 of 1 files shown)


inc/wc-calypso-bridge-hide-alerts.php
```diff
@@ -12,3 +12,14 @@
 // Hide Apple Pay and Google Payment notices
 add_filter( 'pre_option_wc_stripe_show_apple_pay_notice', '__return_true' );
 add_filter( 'pre_option_wc_stripe_show_request_api_notice', '__return_true' );
+
+// Hide setup store notice.
+add_filter( 'woocommerce_show_admin_notice', 'wc_calypso_bridge_hide_admin_notice' );
+
+function wc_calypso_bridge_hide_admin_notice( $notice ) {
+	if ( 'install' == $notice ) {
```
justinshreve: `===`

inc/wc-calypso-bridge-hide-alerts.php
```diff
@@ -12,3 +12,14 @@
 // Hide Apple Pay and Google Payment notices
 add_filter( 'pre_option_wc_stripe_show_apple_pay_notice', '__return_true' );
 add_filter( 'pre_option_wc_stripe_show_request_api_notice', '__return_true' );
+
+// Hide setup store notice.
+add_filter( 'woocommerce_show_admin_notice', 'wc_calypso_bridge_hide_admin_notice' );
+
+function wc_calypso_bridge_hide_admin_notice( $notice ) {
```
justinshreve: I think we need to capture [both arguments here ($bool, $notice)](https://github.com/woocommerce/woocommerce/blob/7eedf1c5134607c3d38b085fc88721fc5d2d14af/includes/admin/class-wc-admin-notices.php#L154) so that we can return the correct value, instead of always returning `true`.

timmyc:
Fixed up, thanks for the review @justinshreve

PR #6, opened 2017-11-06
Bump version to 0.1.1

timmyc:
Updating version numbers so we can tag & update wpcomsh

### DIFF
--- a/readme.txt
+++ b/readme.txt
@@ -24,3 +24,6 @@ This section describes how to install the plugin and get it working.
 
 = 0.1.0 =
 * Initial release
+
+= 0.1.1 =
+* Adding back the forgotten email templates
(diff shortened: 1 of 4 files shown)


readme.txt
```diff
@@ -23,4 +23,7 @@ This section describes how to install the plugin and get it working.
 == Changelog ==
 
 = 0.1.0 =
-* Initial release
\ No newline at end of file
+* Initial release
```
allendav: We usually order the changelog with the newest releases towards the top

allendav:
I see a 0.8.9 in Stable tag in readme.txt - that needs updating too, right?

justinshreve:
> I see a 0.8.9 in Stable tag in readme.txt - that needs updating too, right?

Ah good catch.

It's not the end of the world since a lot of those headers (and readme.txt) are used if the plugin was on WP.org. We are mainly using it as a changelog. We could probably just remove the Stable tag.

PR #10, opened 2017-11-16
Add BACS accounts to API response

timmyc:
This PR is for https://github.com/Automattic/wp-calypso/issues/16630 and adds BACS account data to the current `/wc/v3/payment_gateways/*` API responses. This will allow us to expose this data within Store on WPCOM. 

In this PR I have also added a test suite akin to what we have in `wc-api-dev` for adding unit test coverage for the various API bits and pieces added to this project.

__To Test__
- The easiest way to test this PR is to use Postman, and install this branch as a plugin on a test site.
- You can either call the endpoints directly on your test site via `http://testsitedomain.com/wp-json/wc/v3/payment_gateways/` and `http://testsitedomain.com/wp-json/wc/v3/payment_gateways/bacs` or by passing them through the wpcom REST API like `https://public-api.wordpress.com/rest/v1.1/jetpack-blogs/137330147/rest-api/?http_envelope=1&path=%2Fwc%2Fv3%2Fpayment_gateways%26_method%3Dget%26json%3Dtrue%26context%3Dedit` but be sure to replace the site ID with that of your Jetpack connected blog ID.
- Regardless of method chosen, you will need to perform the API requests via an authenticated means ( i.e. a bearer token for WPCOM ).
- Verify that a new `accounts` key is present in the `settings` for the BACS payment type on both all payments response, and the singular response for only `bacs`:

postman

If you do not have any BACS accounts configured the `value` should be set to an empty array. After you verify the empty array is returned by default, visit https://yourtestdomain,.com/wp-admin/admin.php?page=wc-settings&tab=checkout&section=bacs, add some Account details, and verify they are returned as expected in the API requests too.

### DIFF
--- a/inc/wc-calypso-bridge-add-bacs-accounts.php
+++ b/inc/wc-calypso-bridge-add-bacs-accounts.php
@@ -1,0 +1,27 @@
+<?php
+/**
+ * Adds BACS Accounts to /wc/v3/payment_gateways response when applicable
+ *
+ * @since 0.1.0
+ */
+
+if ( ! defined( 'ABSPATH' ) ) {
+	exit;
+}
+
+
+function wc_calypso_bridge_add_bacs_accounts( $response, $gateway, $request ) {
+	if ( is_wp_error( $response ) ) {
+		return $response;
+	}
+
+	if ( 'bacs' == $gateway->id ) {
+		$response->data[ 'settings' ][ 'accounts' ] = array(
+			'id'    => 'accounts',
+			'value' => get_option( 'woocommerce_bacs_accounts', array() ),
+		);
+	}
+	return $response;
+}
+
+add_filter( 'woocommerce_rest_prepare_payment_gateway', 'wc_calypso_bridge_add_bacs_accounts', 10, 3 );
--- a/inc/wc-calypso-bridge-add-bacs-accounts.php
+++ b/inc/wc-calypso-bridge-add-bacs-accounts.php
@@ -15,7 +15,7 @@ function wc_calypso_bridge_add_bacs_accounts( $response, $gateway, $request ) {
 		return $response;
 	}
 
-	if ( 'bacs' == $gateway->id ) {
+	if ( 'bacs' === $gateway->id ) {
 		$response->data[ 'settings' ][ 'accounts' ] = array(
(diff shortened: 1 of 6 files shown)


tests/unit-tests/bacs-accounts.php
```diff
@@ -0,0 +1,151 @@
...
+	}
+
+	/**
+	 * Test actual account data is returned if set.
+	 */
+	public function test_get_all_payment_gateways_includes_actual_bacs_accounts() {
+		wp_set_current_user( $this->user );
+		$account_data = array(
+			array(
+				'account_name'   => 'chicken and ribs',
```
budzanowski: lol

budzanowski:
I understand that the test suite does autodiscovery of test in testfiles?

timmyc:
> I understand that the test suite does autodiscovery of test in testfiles?

Correct, I believe the following in `phpunit.xml` grabs all the test files for running:

`<directory suffix=".php">./tests/unit-tests</directory>`

inc/wc-calypso-bridge-add-bacs-accounts.php
```diff
@@ -0,0 +1,27 @@
...
+	exit;
+}
+
+
+function wc_calypso_bridge_add_bacs_accounts( $response, $gateway, $request ) {
+	if ( is_wp_error( $response ) ) {
+		return $response;
+	}
+
+	if ( 'bacs' == $gateway->id ) {
```
justinshreve: Should this be `===`?

tests/bootstrap.php
```diff
@@ -0,0 +1,158 @@
...
+	 */
+	public function load_wc() {
+		require_once( $this->wc_dir . '/woocommerce.php' );
+	}
+
+	/**
+	 * Load WC API Dev.
+	 */
+	public function load_wc_api_dev() {
+		define( 'WC_API_DEV_ENABLE_HOTFIXES', false );
```
justinshreve: I think we don't need this anymore (here or in wc-api-dev). From a quick search, I don't think it's used anywhere.

timmyc: Indeed you are correct.

PR #15, opened 2017-12-12
Add Endpoint to return currency data.

timmyc:
This branch is for https://github.com/Automattic/wp-calypso/issues/20675 - which calls for adding in a currency selector to the address page in the setup wizard. While there is some [currency data](https://github.com/Automattic/wp-calypso/blob/master/client/lib/format-currency/currencies.js) available in Calypso, it is lacking the descriptive name that is shown in the Woo Setup wizard.

__To Test__
Install this branch as a plugin on a test site and using something like Postman App, visit `/wc/v3/currencies` with a valid auth token. Verify currency data is returned. Perform the same request without an auth token and verify no data is returned.

### DIFF
--- a/api/class-wc-calypso-bridge-currencies-controller.php
+++ b/api/class-wc-calypso-bridge-currencies-controller.php
@@ -1,0 +1,81 @@
+<?php
+/**
+ * REST API WC Calypso Bridge Currencies
+ *
+ * Adds an endpoint for returning WooCommerce Currency Data
+ *
+ * @author   Automattic
+ * @category API
+ */
+
+if ( ! defined( 'ABSPATH' ) ) {
+	exit;
+}
+
+/**
+ * @package WooCommerce/API
+ */
+class WC_Calypso_Bridge_Currencies_Controller extends WC_REST_Controller {
+
+	/**
+	 * Endpoint namespace.
+	 *
+	 * @var string
+	 */
+	protected $namespace = 'wc/v3';
+
+	/**
+	 * Route base.
+	 *
+	 * @var string
+	 */
+	protected $rest_base = 'currencies';
+
+	/**
+	 * Register Currency route
+	 */
+	public function register_routes() {
+		register_rest_route( $this->namespace, '/' . $this->rest_base, array(
+			array(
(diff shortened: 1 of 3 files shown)


api/class-wc-calypso-bridge-currencies-controller.php
```diff
@@ -0,0 +1,81 @@
...
+
+	/**
+	 * Get WooCommerce Currencies.
+	 *
+	 * @param WP_REST_Request
+	 * @return WP_REST_Response
+	 *
+	 */
+	public function get_currencies( $request ) {
+        $currencies = array();
```
budzanowski: Indentation looks a bit off.

tests/unit-tests/currencies-controller.php
```diff
@@ -0,0 +1,51 @@
...
+	 * Test getting currencies when authed
+	 * which defaults to an empty array.
+	 *
+	 * @since 3.0.0
+	 */
+	public function test_get_currencies() {
+		wp_set_current_user( $this->user );
+
+		$response = $this->server->dispatch( new WP_REST_Request( 'GET', '/wc/v3/currencies' ) );
+        $data = $response->get_data();
```
budzanowski: Indentation.

tests/unit-tests/currencies-controller.php
```diff
@@ -0,0 +1,51 @@
...
+		parent::setUp();
+		$this->endpoint = new WC_Calypso_Bridge_Currencies_Controller();
+		$this->user = $this->factory->user->create( array(
+			'role' => 'administrator',
+		) );
+	}
+
+	/**
+	 * Test getting currencies when authed
+	 * which defaults to an empty array.
```
budzanowski: empty array?

budzanowski:
Left small nits, but otherwise looks good.

tests/unit-tests/currencies-controller.php
timmyc: copy paste 💯

budzanowski:
Good :+1: :ship: :it:

PR #20, opened 2017-12-19
WooCommerce Analytics: Add module for testing

psealock:
Add WooCommerce Analytics, which is slated for a February 2018 release of Jetpack. Inclusion in this repo is for testing on a select pool of stores and is an exact copy of the code introduced to Jetpack in https://github.com/Automattic/jetpack/pull/8296

### Local Testing instructions
* Compress this directory into a zipfile
* Manually upload it a test site
* View a product page
* In the console, see the `_wca` object is defined

### ToDo's 
- [ ] Enable https://stats.wp.com/s.js by deploying `s.js`. 
- [ ] Identify a pool of suitable test sites

cc @greenafrican

### DIFF
--- a/inc/woocommerce-analytics/classes/wp-woocommerce-analytics-universal.php
+++ b/inc/woocommerce-analytics/classes/wp-woocommerce-analytics-universal.php
@@ -1,0 +1,322 @@
+<?php
+/**
+ * Jetpack_WooCommerce_Analytics_Universal
+ *
+ * @package Jetpack
+ * @author Automattic
+ */
+
+/**
+ * Bail if accessed directly
+ */
+if ( ! defined( 'ABSPATH' ) ) {
+	exit;
+}
+
+/**
+ * Class Jetpack_WooCommerce_Analytics_Universal
+ * Filters and Actions added to Store pages to perform analytics
+ */
+class Jetpack_WooCommerce_Analytics_Universal {
+	/**
+	 * Jetpack_WooCommerce_Analytics_Universal constructor.
+	 */
+	public function __construct() {
+		// loading _wca
+		add_action( 'wp_head', array( $this, 'wp_head_top' ), 1 );
+
+		// loading s.js
+		add_action( 'wp_head', array( $this, 'wp_head_bottom' ), 999999 );
+
+		// single product page view
+		add_action( 'woocommerce_after_single_product', array( $this, 'product_detail' ) );
+
+		// add to cart on single product page
+		add_action( 'woocommerce_after_add_to_cart_button', array( $this, 'add_to_cart' ) );
+
+		// add to carts from list views (search, store etc.)
+		add_action( 'wp_footer', array( $this, 'loop_add_to_cart' ) );
+
+
+		add_action( 'woocommerce_after_cart', array( $this, 'remove_from_cart' ) );
+		add_action( 'woocommerce_after_mini_cart', array( $this, 'remove_from_cart' ) );
+		add_action( 'wcct_before_cart_widget', array( $this, 'remove_from_cart' ) );
+		add_filter( 'woocommerce_cart_item_remove_link', array( $this, 'remove_from_cart_attributes' ), 10, 2 );
+
+		// cart checkout
+		add_action( 'woocommerce_after_checkout_form', array( $this, 'checkout_process' ) );
+
+		// order confirmed
+		add_action( 'woocommerce_thankyou', array( $this, 'order_process' ), 10, 1 );
+		add_action( 'woocommerce_after_cart', array( $this, 'remove_from_cart_via_quantity' ), 10, 1 );
+	}
+
+	/**
+	 * Make _wca available to queue events
+	 */
+	public function wp_head_top() {
+		$wca_code = "<script>window._wca = window._wca || [];</script>";
+		echo "$wca_code\r\n";
+	}
+
+
+	/**
+	 * Place script to call s.js, Store Analytics
+	 */
+	public function wp_head_bottom() {
+		$filename = 's-' . gmdate( 'YWd' ) . '.js';
+		$async_code = "<script async src='https://stats.wp.com/" . $filename . "'></script>";
+		echo "$async_code\r\n";
+	}
+
+	/**
+	 * On a product page, add a click event listener to "Add to Cart" button click
+	 */
+	public function add_to_cart() {
+
+		if ( ! is_single() ) {
+			return;
+		}
+
+		$blogid = Jetpack::get_option( 'id' );
+		global $product;
+
+		wc_enqueue_js(
+			"jQuery( '" . esc_js( '.single_add_to_cart_button' ) . "' ).click( function() {
+				_wca.push( {
+					'_en': 'woocommerceanalytics_add_to_cart',
+					'blog_id': " . esc_js( $blogid ) . ",
+					'pi': '" . esc_js( $product->get_id() ) . "',
+					'pn' : '" . esc_js( $product->get_title() ) . "',
+					'pq': jQuery( 'input.qty' ).val() ? jQuery( 'input.qty' ).val() : '1',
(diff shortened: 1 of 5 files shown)


timmyc:
> Identify a pool of suitable test sites

I'll offer up oregonpaws.blog - I can even try to get some more clicks on there by running a promotion on some local facebook groups.

inc/woocommerce-analytics/classes/wp-woocommerce-analytics-universal.php
```diff
@@ -0,0 +1,321 @@
...
+ * Jetpack_WooCommerce_Analytics_Universal
+ *
+ * @package Jetpack
+ * @author Automattic
+ */
+
+/**
+ * Bail if accessed directly
+ */
+if ( ! defined( 'ABSPATH' ) ) {
```
timmyc: Just to be safe, maybe we should also add `class_exists( 'Jetpack_WooCommerce_Analytics_Universal' )` to this `if` block and exit out if it already does... that way if we do deploy this out to AT sites, and there is also a Jetpack release, we won't have duplicate logic.

inc/woocommerce-analytics/classes/wp-woocommerce-analytics-utils.php
```diff
@@ -0,0 +1,51 @@
...
+ * Jetpack_WooCommerce_Analytics_Options provides a single interface to module options
+ *
+ * @package Jetpack
+ * @author Automattic
+ */
+
+/**
+ * Bail if accessed directly
+ */
+if ( ! defined( 'ABSPATH' ) ) {
```
timmyc: likewise on the `class_exists` here.

inc/woocommerce-analytics/wp-woocommerce-analytics.php
```diff
@@ -0,0 +1,92 @@
+<?php
+/**
+ * Jetpack_WooCommerce_Analytics is ported from the Jetpack_Google_Analytics code.
+ *
+ * @package Jetpack
+ */
+
+if ( ! defined( 'ABSPATH' ) ) {
```
timmyc: Or I suppose just do it once here and call it good 😆

wc-calypso-bridge-class.php
```diff
@@ -79,10 +79,16 @@ public function includes() {
...
+
 		if ( class_exists( 'MailChimp_Woocommerce' ) ) {
 			include_once( dirname( __FILE__ ) . '/api/class-wc-calypso-bridge-mailchimp-settings-controller.php' );
 		}
+
+		/**
+		 * Store Stats Testing. woocommerce-analytics will eventually land in Jetpack.
+		 * See https://github.com/Automattic/jetpack/pull/8296 for more details
+		 */
+		include_once( dirname( __FILE__ ) . '/inc/woocommerce-analytics/wp-woocommerce-analytics.php' );
```
timmyc: do we want to add a blog ID check here, or in `isActiveStore` maybe?

inc/woocommerce-analytics/classes/wp-woocommerce-analytics-universal.php
psealock: Nice... I ran into this problem locally

wc-calypso-bridge-class.php
psealock: Lets add it here. I'm copy/pasting the entire folder from the original PR (https://github.com/Automattic/jetpack/pull/8296), so it would be good to keep Calypso Bridge specific things outside of `/inc/woocommerce-analytics`.

inc/woocommerce-analytics/classes/wp-woocommerce-analytics-universal.php
psealock: I addressed this in https://github.com/Automattic/wc-calypso-bridge/pull/20/commits/4fd56b7008312750512405a2d4841d63eddba939

`class_exists( 'Jetpack_WooCommerce_Analytics' )` on the page where `Jetpack_WooCommerce_Analytics` is defined results in `true` even for a singular instance, so I've placed the check before any of the `wca` code is included.

inc/woocommerce-analytics/classes/wp-woocommerce-analytics-universal.php
```diff
@@ -64,7 +64,8 @@ public function wp_head_top() {
 	 * Place script to call s.js, Store Analytics
 	 */
 	public function wp_head_bottom() {
-		$async_code = "<script async src='https://stats.wp.com/s.js'></script>";
+		$filename = 's-' . gmdate( 'YWd' ) . '.js';
```
timmyc: Nice! One less spot to manually bump for a cache bust 🌟 !

timmyc:
Random question - do we want to exclude site admins from the tracking here? I know for general page view stats when site owner / admins are logged in, we do not count their page views:

> Visits you make to your own publicly available blog (when logged in to your account)

via https://en.support.wordpress.com/stats/#views-and-visitors

timmyc:
Okay I loaded up the branch on my site and saw the `s.js` failures as expected, along with data being added to `_wca` too

psealock:
>do we want to exclude site admins from the tracking here?

Makes sense to me. Introduced a check for logged in site admins in 
https://github.com/Automattic/jetpack/pull/8296/commits/6ad85c871e8cc8b85eb39fc2bb7e9061a2ed763a

psealock:
https://github.com/Automattic/wc-calypso-bridge/pull/20/commits/47f54e108e1da78f6946deae9c52009353e34cde introduces an array of stores to Track.

https://github.com/Automattic/wc-calypso-bridge/blob/47f54e108e1da78f6946deae9c52009353e34cde/wc-calypso-bridge-class.php#L92-L97

@greenafrican Any others?

greenafrican:
Having a look at this now and have a quick question: should I run the `master` version of Jetpack before manually uploading and testing it?

@psealock

psealock:
>should I run the master version of Jetpack before manually uploading and testing it?

Yes, this code checks for an existing instance of the WCA class
https://github.com/Automattic/wc-calypso-bridge/blob/47f54e108e1da78f6946deae9c52009353e34cde/wc-calypso-bridge-class.php#L91

wc-calypso-bridge-class.php
```diff
@@ -89,7 +89,15 @@ public function includes() {
 		 * See https://github.com/Automattic/jetpack/pull/8296 for more details
 		 */
 		if ( ! class_exists( 'Jetpack_WooCommerce_Analytics' ) ) {
```
timmyc: Should we ensure the class_exists( 'Jetpack' ) too? Granted it should always be loaded on AT sites, might not hurt to add the check in this repo.

psealock: Check added in https://github.com/Automattic/wc-calypso-bridge/pull/20/commits/020adb2dfb1e6d49e98a83da3fe6314253b54c9a

psealock:
https://github.com/Automattic/wc-calypso-bridge/pull/20/commits/178fcbc926d3fdfd36fcbf97ea9c065c356d1c20 removes a check for test sites. This one is ready to be shipped to all Atomic sites.

@timmyc Can you give this last commit an OK and I will rebase this and merge? What is the deployment process, by the way?

timmyc:
178fcbc looks great to me @psealock - but yes @justinshreve if you could give this one more review/test that would be great. I've been running the logic stand-alone on oregonpaws.blog for about a month now so feeling pretty good about it.

timmyc:
> did we still want to roll this out for a selection of sites?

I'm thinking it is fine to roll to all AT Woo sites. Considering we are pushing to have this released in Jetpack, which will be over 700k Woo sites, testing it at AT-scale first seems like a good thing to do.

As long as @psealock and @greenafrican are good with it I think we should be fine.

PR #21, opened 2018-01-05
Add ability to send an invoice email via API.

timmyc:
This branch introduces a new endpoint for https://github.com/Automattic/wp-calypso/issues/15679 - which will allow triggering an order invoice email to be send via the API.

__Background__
I'm not 💯 % sold on the path name I have used in this branch, so please feel free to offer suggestions. Additionally, what to return as a response in this action was kind of odd to figure out. I opted to return the note that is created when this action is taken in wp-admin - since that is in-fact an object created by the request... but if so, should this be a `POST`!?

__To Test__
Good news is, the functionality works great in my testing - you too can check it out by:

- running `phpunit` to run the new tests for this endpoint
- I setup an email logging plugin on my test site, and verified the order invoice email was indeed sent when I requested the url via Postman via GET `http://domain/index.php?rest_route=/wc/v3/orders/send_invoice/ORDERID`

### DIFF
--- a/api/class-wc-calypso-bridge-send-invoice-controller.php
+++ b/api/class-wc-calypso-bridge-send-invoice-controller.php
@@ -1,0 +1,127 @@
+<?php
+/**
+ * REST API WC Calypso Bridge Send Invoice
+ *
+ * Adds an endpoint to trigger sending an order invoice email
+ *
+ * @author   Automattic
+ * @category API
+ */
+
+if ( ! defined( 'ABSPATH' ) ) {
+	exit;
+}
+
+/**
+ * @package WooCommerce/API
+ */
+class WC_Calypso_Bridge_Send_Invoice_Controller extends WC_REST_Controller {
+
+	/**
+	 * Endpoint namespace.
+	 *
+	 * @var string
+	 */
+	protected $namespace = 'wc/v3';
+
+	/**
+	 * Route base.
+	 *
+	 * @var string
+	 */
+	protected $rest_base = 'orders/send_invoice';
+
+	/**
+	 * Post type.
+	 *
+	 * @var string
+	 */
+	protected $post_type = 'shop_order';
+
+	/**
+	 * Register Currency route
+	 */
+	public function register_routes() {
+		register_rest_route( $this->namespace, '/' . $this->rest_base . '/(?P<id>[\d]+)' , array(
+			'args' => array(
+				'id' => array(
+					'description' => __( 'Unique identifier for the resource.', 'woocommerce' ),
+					'type'        => 'integer',
+				)
+			),
+			array(
+				'methods'             => WP_REST_Server::READABLE,
+				'callback'            => array( $this, 'send_invoice' ),
+				'permission_callback' => array( $this, 'permissions_check' ),
+			)
+		) );
+	}
+
+	/**
+	 * Makes sure the current user has permissions.
+	 *
+	 * @param WP_REST_Request $request Full data about the request.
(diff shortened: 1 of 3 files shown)


timmyc:
@ryelle wanted to make sure this looked good to you to / has all the requisite bits needed for triggering the emails from within Calypso.

🙌  thanks for the review @justinshreve - changed to `POST` in ac58261

api/class-wc-calypso-bridge-send-invoice-controller.php
```diff
@@ -0,0 +1,127 @@
...
+	 *
+	 * @var string
+	 */
+	protected $post_type = 'shop_order';
+
+	/**
+	 * Register Currency route
+	 */
+	public function register_routes() {
+		register_rest_route( $this->namespace, '/' . $this->rest_base . '/(?P<id>[\d]+)' , array(
```
ryelle: I think this endpoint should follow the other order resources, and be `/orders/<id>/send_invoice` (see [notes](https://github.com/woocommerce/woocommerce/blob/master/includes/api/v1/class-wc-rest-order-notes-controller.php#L37), for example)

api/class-wc-calypso-bridge-send-invoice-controller.php
```diff
@@ -0,0 +1,127 @@
...
+		WC()->mailer()->customer_invoice( $order );
+		// Note the event.
+		$note_id = $order->add_order_note( __( 'Order details manually sent to customer.', 'woocommerce' ), false, true );
+		if ( ! $note_id ) {
+			return new WP_Error( 'woocommerce_api_cannot_create_order_note', __( 'Cannot create order note, please try again.', 'woocommerce' ), array( 'status' => 500 ) );
+		}
+		$note = get_comment( $note_id );
+
+		do_action( 'woocommerce_after_resend_order_email', $order, 'customer_invoice' );
+		$response = $this->prepare_item_for_response( $note, $request );
```
ryelle: I'm not sure the note is a good response, since it might happen regardless of whether the email goes through, but since `WC()->mailer()->customer_invoice( $order );` doesn't return anything I guess this is the best option.

justinshreve:
> Also, what email logging plugin did you use? I'm not sure my local site can send emails

I know you are asking Timmy, but it worked for me by testing it on my wpsandbox by installing it as a separate plugin.

ryelle:
> I know you are asking Timmy, but it worked for me by testing it on my wpsandbox by installing it as a separate plugin.

Good point, I'll try my sandbox 🙂  Looks like that worked, though the email sends to spam. I bet that has more to do with me constantly creating/changing orders. I also get the payment link in the email when the order is payment-pending, so this is great 🎉

api/class-wc-calypso-bridge-send-invoice-controller.php
timmyc: Yeah - i know in some wpcom endpoints we return something simple like `success: true` - but from my cursory tour of the WC api - i don't see any current endpoints that don't return objects/arrays

timmyc:
I installed [this plugin](https://wordpress.org/plugins/email-log/) which logs all emails sent and makes them viewable in wp-admin... have seen that offered as a tip a few times when diagnosing email issues ( like that one time when I broke some emails for atomic+woo sites 😱 )

timmyc:
@ryelle updated to have the endpoint at `wc/v3/orders/<id>/send_invoice` in 05485fc

ryelle:
Looks good! Successfully sent an invoice with the new format 👍

PR #29, opened 2018-02-13
Prevent referrers getting set during the checkout process 

greenafrican:
To ensure payment gateways aren't set as referrers. This corresponds to a proposed code fix in Jetpack in the parallel branch -  https://github.com/Automattic/jetpack/pull/8296/commits/f91e76483119c398d711821ac93eeab262284f03

This will require an update to `s.js` to skip setting the referrer.

### DIFF
--- a/inc/woocommerce-analytics/classes/wp-woocommerce-analytics-universal.php
+++ b/inc/woocommerce-analytics/classes/wp-woocommerce-analytics-universal.php
@@ -55,6 +55,10 @@ class Jetpack_WooCommerce_Analytics_Universal {
 	 * Make _wca available to queue events
 	 */
 	public function wp_head_top() {
+		if ( is_cart() || is_checkout() || is_checkout_pay_page() || is_order_received_page() || is_add_payment_method_page() ) {
+			$prevent_referrer_code = "<script>window._wca_prevent_referrer = true;</script>";
+			echo "$prevent_referrer_code\r\n";
+		}
 		$wca_code = "<script>window._wca = window._wca || [];</script>";
 		echo "$wca_code\r\n";
 	}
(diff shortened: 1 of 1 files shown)


psealock:
You'll need to bump the version in 2 places (the [class](https://github.com/Automattic/wc-calypso-bridge/blob/master/wc-calypso-bridge-class.php#L10) and [composer.json](https://github.com/Automattic/wc-calypso-bridge/blob/master/composer.json#L3) and update the [release notes](https://github.com/Automattic/wc-calypso-bridge/blob/master/readme.txt#L23)).

More on that here, https://teamisotope.wordpress.com/field-notes/

timmyc:
> You'll need to bump the version in 2 places

@psealock I have one other PR that landed yesterday so we can get this merged in and do a release PR after.

psealock:
@greenafrican I tested this with COD and Stripe flows, but not paypal (they don't make it as easy to set up test accounts/credentials). Works as expected.

Are you able to determine from the data where the bad referrers are set? I just want to make sure we're not missing a page.

greenafrican:
@psealock 

> Are you able to determine from the data where the bad referrers are set? I just want to make sure we're not missing a page.

Great question! I've run a query to try and determine when the first event occurs with a payment gateway set as the referrer and it seems to be mostly from the `product_purchase` event, which we only capture on the `woocommerce_thankyou` order confirmation hook.

I advocate we get this in and keep our eyes on it to ensure we catch other edge cases.

greenafrican:
@timmyc 
> One quick suggestion, and does the s.js update have to happen before or at the same time?

I'd say it should happen at the same time, but it won't break anything in `s.js`, or vice-versa if they are deployed at slightly different times.

PR #32, opened 2018-02-16
Add created_via to products and jetpack sync for products

### LINKED ISSUE #31: Add and sync _created_via to product post meta to record how a product was added
Rationale: Add a new meta that can be synced back ala Jetpack so we can better understand the distribution of product creation methods

Dictionary:

- `post-new` : product was created via wp-admin/post-new
- `importer` : product was created via importer
- `api-v2` : product was created via api v2
- `api-v3` : product was created via api v3
- `unknown` : product was created via any other means

allendav:
Fixes #31 

Makes it possible to understand the origin of products sync-ed to Jetpack (importer, wp-admin, REST API v3, etc) by adding and syncing that post meta to WordPress.com

To test: See D10144

### DIFF
--- a/inc/wc-calypso-bridge-products.php
+++ b/inc/wc-calypso-bridge-products.php
@@ -29,6 +29,7 @@ function wc_calypso_bridge_products_get_context() {
 
 	$request_uri = isset( $_SERVER['REQUEST_URI'] ) ? $_SERVER['REQUEST_URI'] : '';
 	$known_fragments = array(
+		'&_via_calypso'           => 'calypso',
 		'/wp-json/wc/v2'          => 'rest-api-v2',
 		'/?rest_route=%2Fwc%2Fv2' => 'rest-api-v2-rest-route',
 		'/wp-json/wc/v3'          => 'rest-api-v3',
@@ -59,4 +60,4 @@ function wc_calypso_bridge_products_wp_insert_post( $post_ID, $post, $update ) {
 	update_post_meta( $post_ID, '_created_via', $created_via );
 }
 
-add_action( 'wp_insert_post', 'wc_calypso_bridge_products_wp_insert_post', 10, 4 );
+add_action( 'wp_insert_post', 'wc_calypso_bridge_products_wp_insert_post', 10, 3 );
--- a/inc/wc-calypso-bridge-products.php
+++ b/inc/wc-calypso-bridge-products.php
@@ -28,7 +28,7 @@ function wc_calypso_bridge_products_get_context() {
 	}
 
 	$request_uri = isset( $_SERVER['REQUEST_URI'] ) ? $_SERVER['REQUEST_URI'] : '';
-	$triggers = array(
+	$known_fragments = array(
 		'/wp-json/wc/v2'          => 'rest-api-v2',
 		'/?rest_route=%2Fwc%2Fv2' => 'rest-api-v2-rest-route',
 		'/wp-json/wc/v3'          => 'rest-api-v3',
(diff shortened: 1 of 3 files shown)


justinshreve:
I'll give everything more of a look & test on Monday, but one suggestion. Maybe we could add a query argument to our REST API call in Calypso, so that we could also break out Calypso specific creates from `/v3`? It's _probably_ safe to assume right now `/v3` calls are from Calypso, but this would future proof it.

allendav:
@justinshreve wrote

> Maybe we could add a query argument to our REST API call in Calypso, so that we could also break out Calypso specific creates from /v3? It's probably safe to assume right now /v3 calls are from Calypso, but this would future proof it.

I like that - will add that here and open an issue there as well

inc/wc-calypso-bridge-products.php
```diff
@@ -0,0 +1,62 @@
...
+ */
+
+if ( ! defined( 'ABSPATH' ) ) {
+	exit;
+}
+
+function wc_calypso_bridge_products_get_context() {
+	if ( defined( 'DOING_AJAX' ) && DOING_AJAX ) {
+		$referrer = isset( $_SERVER['HTTP_REFERER'] ) ? $_SERVER['HTTP_REFERER'] : '';
+		$known_referrers = array(
```
justinshreve: These equal signs should be aligned per the WP PHP coding standards.

inc/wc-calypso-bridge-products.php
```diff
@@ -0,0 +1,62 @@
...
+				return $context;
+			}
+		}
+
+		return 'ajax-unknown';
+	}
+
+	$request_uri = isset( $_SERVER['REQUEST_URI'] ) ? $_SERVER['REQUEST_URI'] : '';
+	$known_fragments = array(
+		'/wp-json/wc/v2'          => 'rest-api-v2',
```
justinshreve: We should probably support v1 as well. There are also legacy endpoint versions before that, though I think it's more likely v1 is still in use.

inc/wc-calypso-bridge-products.php
```diff
@@ -0,0 +1,62 @@
...
+		foreach ( $known_referrers as $known_referrer => $context ) {
+			if ( false !== strpos( $referrer, $known_referrer ) ) {
+				return $context;
+			}
+		}
+
+		return 'ajax-unknown';
+	}
+
+	$request_uri = isset( $_SERVER['REQUEST_URI'] ) ? $_SERVER['REQUEST_URI'] : '';
```
justinshreve: These equal signs should be aligned.

inc/wc-calypso-bridge-products.php
```diff
@@ -0,0 +1,62 @@
...
+	$product_post_types = array( 'product', 'product_variation' );
+	if ( ! in_array( $post->post_type, $product_post_types ) ) {
+		return;
+	}
+
+	$created_via = wc_calypso_bridge_products_get_context();
+	update_post_meta( $post_ID, '_created_via', $created_via );
+}
+
+add_action( 'wp_insert_post', 'wc_calypso_bridge_products_wp_insert_post', 10, 4 );
```
justinshreve: You pass `4` here, but only accept 3 arguments above in `wc_calypso_bridge_products_wp_insert_post`

allendav: Fixed in https://github.com/Automattic/wc-calypso-bridge/pull/32/commits/f705b65953b536c96f15865bfd668c35fc402532

inc/wc-calypso-bridge-products.php
allendav: Aligned in https://github.com/Automattic/wc-calypso-bridge/pull/32/commits/e69a0aecbdc223db09d4f0e04add5162742dc5b4

inc/wc-calypso-bridge-products.php
allendav: Aligned in https://github.com/Automattic/wc-calypso-bridge/pull/32/commits/e69a0aecbdc223db09d4f0e04add5162742dc5b4

PR #37, opened 2018-03-02
Fix Email Footer Text

### LINKED ISSUE #7: Match footer text behaviour with WooCommerce Core
ticket - 771133-zen

Background: Installing WooCommerce on a WordPress.com Business plan should feel identical to WooCommerce on any other host.

Relating to woocommerce/woocommerce/pull/17057 - can we revert or amend woocommerce/wc-api-dev/pull/32 to match? 

The ideal scenario, I can amend the [Footer text](https://docs.woocommerce.com/document/configuring-woocommerce-settings/#section-20) and it will show whatever I want but _not_ a hardcoded `Powered by WooCommerce' appended.

timmyc:
This morning @dougaitken informed me via slack that the issue around custom email footer text was still lingering. While we did remove the email template over-rides, I failed to notice the logic in `inc/wc-calypso-bridge-email-site-title.php` which is filtering the footer text too.

The logic was added in https://github.com/woocommerce/wc-api-dev/pull/32 by @justinshreve so wondering if this is okay to remove, or if there is still a flow where this can be problematic.

Since this code runs against all pressable woo installs, part of me feels like we should just remove it, and look at other ways to fix the initial problem these filters were aiming to solve. But really I'd be interested to hear the back story because the initial PR steps to create the problem seem pretty edge case, and I wasn't sure if it was changing the site title in calypso or in wp-admin.

Confirmed that this branch fixes #7


justinshreve:
I'm trying to remember back. Essentially, if you changed your site title (via wp-admin or Calypso), the footer text would still contain the old site title. When people were creating their stores -- this would literally read "Site Title -- Powered by WooCommerce" and we offer no email setting in Calypso to change this.

Since then, https://github.com/woocommerce/woocommerce/issues/17039 was addressed which adds a replacement variable. I think it would also be trivial to add a footer text setting to `https://wordpress.com/store/settings/email/:site` now that we have a settings screen.

Either way, I think removing this filter is fine.

PR #40, opened 2018-03-07
Store: Add `woocommerce_email_footer_text` to the `settings_email_groups` response

justinshreve:
This PR adds `woocommerce_email_footer_text` to the `/wc/v3/settings_email_groups` endpoint (used by email settings in Calypso).

It also fixes tests by removing the require's for webhook factories (which are no longer in core).

To Test:
* Run `phpunit` and make sure all tests pass.
* Use something like postman to make a request to `/wc/v3/settings_email_groups` and make sure `woocommerce_email_footer_text` is present in the response.

### DIFF
--- a/tests/bootstrap.php
+++ b/tests/bootstrap.php
@@ -112,11 +112,6 @@ class WC_Calypso_Bridge_Unit_Tests_Bootstrap {
 	 * If needed, these can also be shipped with the dev plugin and a local copy can be called.
 	 */
 	public function includes() {
-
-		// factories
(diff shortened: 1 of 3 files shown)


tests/bootstrap.php
```diff
@@ -112,11 +112,6 @@ public function install_wc() {
 	 * If needed, these can also be shipped with the dev plugin and a local copy can be called.
 	 */
 	public function includes() {
-
-		// factories
```
timmyc: Heh I got tangled up by these missing files too

PR #42, opened 2018-03-16
WooCommerce Analytics: bring inline with PR in Jetpack for testing

greenafrican:
Updating this PR inline with code changes in Jetpack PR - https://github.com/Automattic/jetpack/pull/9052


greenafrican:
@timmyc @psealock 

I know we spoke about it but can't remember if we included something to prevent a conflict with Jetpack 5.9? I don't see anything but am perhaps not looking in the right place.

timmyc:
Thanks for adding in the logic here too @greenafrican - though I think with the logic right now in this plugin, if Jetpack is v 5.9 or greater - this code will never load.

I think we should just target pushing updates via Jetpack for now since that is really where this code will live moving forward. We should get a PR up here to remove all the logic from calypso-bridge now that it is in Jetpack.

timmyc:
Sorry @greenafrican - forgot to note that it was #35 which added in the restriction so the logic here will no longer run on JP > 5.9

psealock:
Agree with @timmyc here, lets just have this code in JetPack

PR #43, opened 2018-03-16
Remove WC Analytics now that Jetpack 5.9 has been released.

timmyc:
Now that Jetpack 5.9 is live, and WooCommerce Analytics has shipped as part of it - this branch removes all of the logic from calypso bridge.


timmyc:
> despite relying on the slow Jetpack release cycle

Yeah that is a bummer. If there is a serious bug though we could still leverage this plugin to ship hot fixes.

PR #44, opened 2018-04-05
MailChimp: Truncate tables on deactivation.

timmyc:
This branch is part of the cleanup project for the `mailchimp-for-woocommerce` plugin ( background p90Yrv-CD-p2 ). In lieu of making an endpoint to clear out the MC queue tables when we remove the plugins - I opted to hook into the deactivation process here.

FWIW I spoke with the developers today, and they are going to add similar logic to clean up the mysql tables ( possibly DELETE'ing them ) upon deactivation. I still feel a bit more comfortable with the `TRUNCATE` here, just in case for some odd reason a site has a plugin/logic that depends on these tables. Please tell me if I'm being too paranoid here 😟

__To Test__
- Install the `mailchimp-for-woocommerce` plugin on a test site if it isn't already present
- deactivate the plugin, verify nothing explodes
- View the logs under WooCommerce > Status to verify the log entries for truncating all 3 tables are SUCCESS:

deactivate

### DIFF
--- a/inc/wc-calypso-bridge-mailchimp-deactivate-hook.php
+++ b/inc/wc-calypso-bridge-mailchimp-deactivate-hook.php
@@ -1,0 +1,31 @@
+<?php
+/**
+ * Clear out MailChimp tables on deactivation
+ */
+
+if ( ! defined( 'ABSPATH' ) ) {
+	exit;
+}
+
+register_deactivation_hook( 'mailchimp-for-woocommerce/mailchimp-woocommerce.php', 'wc_calypso_bridge_clear_mailchimp_tables' );
+
+function wc_calypso_bridge_clear_mailchimp_tables() {
+	global $wpdb;
+	$log = new WC_Logger();
+
+	$mailchimp_tables = array(
+		"{$wpdb->prefix}queue",
+		"{$wpdb->prefix}failed_jobs",
+		"{$wpdb->prefix}mailchimp_carts",
+	);
+
+	foreach( $mailchimp_tables as $table ) {
+		$sql = "TRUNCATE TABLE `$table`";
+
+		if ( $wpdb->query( $sql ) ) {
+			$log->add( 'mailchimp-for-woocommerce', "Plugin Deactivated: success clearing `$table` table." );
+		} else {
+			$log->add( 'mailchimp-for-woocommerce', "Plugin Deactivated: FAILURE clearing `$table` table." );
+		}
+	}
+}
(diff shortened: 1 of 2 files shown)


allendav:
> FWIW I spoke with the developers today, and they are going to add similar logic to clean up the mysql tables ( possibly DELETE'ing them ) upon deactivation

Might want to have them DROP their tables on plugin delete if they don't do that already too

inc/wc-calypso-bridge-mailchimp-deactivate-hook.php
```diff
@@ -0,0 +1,31 @@
...
+
+if ( ! defined( 'ABSPATH' ) ) {
+	exit;
+}
+
+register_deactivation_hook( 'mailchimp-for-woocommerce/mailchimp-woocommerce.php', 'wc_calypso_bridge_clear_mailchimp_tables' );
+
+function wc_calypso_bridge_clear_mailchimp_tables() {
+	global $wpdb;
+	$log = new WC_Logger();
```
allendav: Please test for WC_Logger before using it. It is possible WooCommerce is not active when the mailchimp plugin is deactivated

inc/wc-calypso-bridge-mailchimp-deactivate-hook.php
```diff
@@ -0,0 +1,31 @@
...
+
+	$mailchimp_tables = array(
+		"{$wpdb->prefix}queue",
+		"{$wpdb->prefix}failed_jobs",
+		"{$wpdb->prefix}mailchimp_carts",
+	);
+
+	foreach( $mailchimp_tables as $table ) {
```
allendav: Optional. I almost always cast the expected array in a foreach explicitly as an array, e.g.

`foreach( (array) $mailchimp_tables as $table ) {`

inc/wc-calypso-bridge-mailchimp-deactivate-hook.php
```diff
@@ -0,0 +1,31 @@
...
+		$sql = "TRUNCATE TABLE `$table`";
```
allendav: Please use https://developer.wordpress.org/reference/classes/wpdb/prepare/ instead

inc/wc-calypso-bridge-mailchimp-deactivate-hook.php
timmyc: I think nothing in the plugin here is ever loaded if WC is not active [#](https://github.com/Automattic/wc-calypso-bridge/blob/master/wc-calypso-bridge-class.php#L40) - but will add a check just to be extra safe.

timmyc:
@allendav is this looking okay to merge now?

PR #53, opened 2018-06-05
Disable publicize on API Product Create

timmyc:
Fixes https://github.com/Automattic/wp-calypso/issues/19457

When creating a new product via Store on WordPress.com - any active publicize connections will publish the new product creation. Since we have no user interface in Store to select which connections to publicize too, this can come as quite a surprise that products are showing up on all configured publicize services.

To best test this PR, it might be nice to replicate the problem first:

__To Verify Bug__
- Setup a publicize connection on a Store ( i use a fake page on Facebook for testing )
- Create a product via Store on WPCOM
- Check the publicize service and see that your new product was published

__To Test The Fix__
- Install this branch on the same test site
- Also having `Post Meta Inspector` plugin installed is handy
- Take the same steps above, and verify the new product is not publicized
- Also inspect the post meta on the new product, and verify you see an entry like:

meta-skip

__Verify Publicize Still Works on non-API Products__
- Then create a product using wp-admin
- Make sure the publicize service is checked in the sidebar of the editor
- Verify the item is still publicized as expected after the product is published

### DIFF
--- a/inc/wc-calypso-bridge-disable-publicize.php
+++ b/inc/wc-calypso-bridge-disable-publicize.php
@@ -31,8 +31,8 @@ function wc_calypso_bridge_maybe_disable_publicize( $submit_post, $post_id, $ser
 			$should_disable_publicize = true;
 			break;
 		}
-	}
-
+    }
+    
     $post_type = get_post_type( $post_id );
 
 	// If not a product, or not a REST API request, return.
--- a/inc/wc-calypso-bridge-disable-publicize.php
+++ b/inc/wc-calypso-bridge-disable-publicize.php
@@ -23,12 +23,12 @@ add_action( 'publicize_save_meta', 'wc_calypso_bridge_maybe_disable_publicize', 
  */
 function wc_calypso_bridge_maybe_disable_publicize( $submit_post, $post_id, $service_name, $connection ) {
     $trigger_strings = array( '/wp-json/wc/v', '/?rest_route=%2Fwc%2Fv' );
-    $should_disable_publicize = false;
+    $is_rest_api_request = false;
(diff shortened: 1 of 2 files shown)


inc/wc-calypso-bridge-disable-publicize.php
```diff
@@ -0,0 +1,51 @@
...
+ * @since 0.2.3
+ *
+ * @param bool $submit_post Should the post be publicized.
+ * @param int $post->ID Post ID.
+ * @param string $service_name Service name.
+ * @param array $connection Array of connection details.
+ */
+function wc_calypso_bridge_maybe_disable_publicize( $submit_post, $post_id, $service_name, $connection ) {
+    $trigger_strings = array( '/wp-json/wc/v', '/?rest_route=%2Fwc%2Fv' );
+    $should_disable_publicize = false;
```
justinshreve: Minor: Maybe name this `is_rest_api_request` instead, since it's used in combination with the product type to decide if we should publicize or not.

inc/wc-calypso-bridge-disable-publicize.php
```diff
@@ -0,0 +1,51 @@
...
+    $post_type = get_post_type( $post_id );
+
+    // If not a product, or not a REST API request, return.
+    if ( 'product' != $post_type || ! $should_disable_publicize ) {
+        return;
+    }
+
+    // Since this is a product, and we are in an API request, disable publicize.
+    if ( ! empty( $connection->unique_id ) ) {
+	    $unique_id = $connection->unique_id;
```
justinshreve: Minor: Spacing/indent here.

PR #56, opened 2018-10-12
Move legacy store on wp.com code

justinshreve:
This PR moves the current "Store on WP.COM" hot-fixes and compatibility code to a new `store-on-wpcom` folder.

This allows us to separate out the code that is necessary for current stores to run, and any new code we write for the hosted eCommerce / Calypsoify experience. It also splits out the tests so any new ones we write do not require `wc-api-dev` like the old ones. It also fixes a few tests that seem to have broken since the last WC update.

This definitely needs more real-world testing before being bundled for a release and pushed, but I don't think there is a need to do that right now, so nothing is stopping us from iterating on this project here in GH. Is master OK or should we create a master feature branch for all this work?

cc @timmyc 

To Test:
* Run `phpunit` from the main plugin folder and see the new example test pass.
* Run `phpunit` from `store-on-wpcom`.
* Spot check a few things like assets correctly loading in the customizer.


wc-calypso-bridge-class.php
```diff
@@ -30,92 +19,11 @@ class WC_Calypso_Bridge {
...
-			$controller_instance = new $controller();
-			$controller_instance->register_routes();
-		}
-
-		// We include it here because rest_api_init is a proper context for mocked add_settings_error function
-		include_once( dirname( __FILE__ ) . '/inc/wc-calypso-bridge-mailchimp-add-settings-error.php' );
-	}
+    }
+    
+    function init() {
```
timmyc: Guessing this could be removed?

justinshreve:
Thanks @timmyc. I must have had some combination of plugin(s)/wc-api-dev versions that prevented me from seeing that error, but I was able to reproduce and got that fixed up. All tests should be passing now. Also removed the extra debug call.

PR #59, opened 2018-10-17
Add WC Calypsoify support for bundled extensions

justinshreve:
The PR ended up being small, but I did go through and test/confirm the following extensions are Calypsoified and show up in the correct menu sections. A good deal of them don't need extra code because they hook into the payment settings, shipping settings, integration tabs, or the customizer. All the extensions I tested as a part of this PR make up the bundled package.

Extensions Tested:

* Facebook
* MailChimp
* Storefront
* Product Add-Ons
* Storefront PowerPack
* TaxJar
* Stripe
* PayPal Express Checkout
* eWay
* Square
* Klarna
* PayFast
* USPS
* Canada Post
* WooCommerce Services
* Royal Mail
* UPS
* Australia Post

To Test:
* Install some (or all) of these extensions and verify they show up in the right section.


timmyc:
Okay something still seems amiss on my local install that is preventing me from testing things properly. Here is the Calypsoified plugin screen, i'm running the `ecommerce-plan` branch on it:

image

timmyc:
Okay finally got things loaded up! I'm seeing the following error now on

http://calypswoo.test/wp-admin/plugins.php?plugin_status=all&paged=1&s

image

If I add the following change, all is good, the plugins page renders:

```
diff --git a/includes/page-controller.php b/includes/page-controller.php
index 4b187e0..c1d4a3c 100644
--- a/includes/page-controller.php
+++ b/includes/page-controller.php
@@ -148,6 +148,9 @@ class WC_Calypso_Bridge_Page_Controller {
         */
        public function register_menu( $options ) {
                global $menu, $submenu;
+               if ( ! array_key_exists( 'path', $options ) ) {
+                       return;
+               }
                $this->menus[] = $options;
                $this->id_mapping_to_path[ $options['id' ] ] = $options['path'];
        }
(END)
```

But then when opening up WooCommerce, no menu items are showing:

image

justinshreve:
@timmyc Oops - I didn't have the correct debug flags in my new WP setup so I didn't even see those notices 🤦‍♂️. The 'path' mapping thing was a copy and paste fail from my wc-admin branch and isn't used here, so none of the menus will get registered with that extra `if` check.

I just pushed a fix to remove that, and it seems to be working now.

timmyc:
Perfect! working well for me now. So I installed Product Add-ons and Facebook. When I click on the Facebook item, my screen looks like:

image

So it seems the sidebar is showing the plugin view, and not the Woo view. Not sure if that is related to this PR, so I'm good with merging and fixing that in another PR

PR #73, opened 2018-10-22
Add masterbar to setup wizard

### LINKED ISSUE #70: OBW: Try to add Masterbar / Calypsoify Mods
This one _might_ be a bit of a stretch, but the goal is to see if we can pull in the blue masterbar, and its related [calypsoify Jetpack modifications](https://github.com/Automattic/jetpack/blob/master/modules/class.jetpack-calypsoify.php#L114-L141).

One approach that @claudiulodro mentioned might work was to over-ride the `setup_wizard_header()` that is in [woo core](https://github.com/woocommerce/woocommerce/blob/60fafd821d3fff2e7e57df9892c90102650dbe68/includes/admin/class-wc-admin-setup-wizard.php#L305-L322)

joshuatf:
Fixes #70 

This PR overrides the default `WC_Admin_Setup_Wizard` class so that we can add in the masterbar and other template changes.

This PR is branched from https://github.com/Automattic/wc-calypso-bridge/pull/72 which includes the initial setup fxns.

### Screenshots

screen shot 2018-10-22 at 5 08 50 pm

### Testing

1.  Visit `/wp-admin/admin.php?page=wc-setup&calypsoify=1`
2.  Verify that admin bar (masterbar) is shown and calypsoified.


includes/setup-wizard.php
```diff
@@ -0,0 +1,53 @@
+<?php
+/**
+ * Setup Wizard Class
+ *
+ * Extends the original WC setup wizard so we can add in new templates to the view.
+ */
+
+if ( ! defined( 'ABSPATH' ) ) {
```
timmyc: Perhaps I'm being paranoid, since this should _never_ happen since this plugin won't load unless Woo is installed, but should we verify that the `WC_Admin_Setup_Wizard` class exists here too and exit if not?

includes/setup-wizard.php
```diff
@@ -0,0 +1,53 @@
...
+			<meta http-equiv="Content-Type" content="text/html; charset=utf-8" />
+			<title><?php esc_html_e( 'WooCommerce &rsaquo; Setup Wizard', 'woocommerce' ); ?></title>
+			<?php do_action( 'admin_enqueue_scripts' ); ?>
+			<?php wp_print_scripts( 'wc-setup' ); ?>
+			<?php do_action( 'admin_print_styles' ); ?>
+			<?php do_action( 'admin_head' ); ?>
+		</head>
+		<body class="wc-setup wp-core-ui">
+			<?php wp_admin_bar_render(); ?>
+			<h1 id="wc-logo"><a href="https://woocommerce.com/"><img src="<?php echo esc_url( WC()->plugin_url() ); ?>/assets/images/woocommerce_logo.png" alt="WooCommerce" /></a></h1>
```
timmyc: Nice! I'm guessing we _might_ be removing this once we get the designs, so handy that we can do so here.

includes/setup-wizard.php
joshuatf: That seems reasonable.  I was contemplating whether or not we needed this since we'll presumably check for dependencies prior to this (actually looks like we currently don't have any dependency checks yet).

We can err on the side of caution here though.

includes/setup-wizard.php
joshuatf: Designs are in and it looks like you're right- we won't have the WC logo!

joshuatf:
@timmyc Thanks for the comments!  

Hmm, the masterbar not showing up is odd.  Can you check if `id="wpadminbar"` is rendered in your HTML?  I had to do a hard refresh to get the Calypsoify styles showing up properly.

timmyc:
Yep i don't see the DOM element, the only `wpadminbar` ref I see is `<style type="text/css" media="print">#wpadminbar { display:none; }</style>`

Double checked i'm on the correct branch, and the correct test site ( these things trip me up often 😆 ) - pulled from origin again, and hard refreshed. To verify I am on http://calypswoo.test/wp-admin/admin.php?page=wc-setup&calypsoify=1 - this is the correct path right?

timmyc:
Yep confirmed the `WC_Calypso_Bridge_Admin_Setup_Wizard` is being loaded, but still not seeing the masterbar bits. Any chance you forgot to check-in some local changes or something?

justinshreve:
I did get the masterbar, but only after going to `/wp-admin/plugins.php?calypsoify=1` first, making sure that CSS/JS was loaded, and then loading `/wp-admin/index.php?page=wc-setup&calypsoify=1`. Otherwise I just got the normal gray wp-admin bar.

Does `admin_init` load during the onboarding wizard? The `check_param` Calypsoify code might not be running on the setup page: https://github.com/Automattic/jetpack/blob/master/modules/class.jetpack-calypsoify.php#L23

There seems to be some extra spacing on the reader masterbar item:

screen shot 2018-10-23 at 12 35 44 pm

There also seems to be a double notification bell (which doesn't work when I click on it):

screen shot 2018-10-23 at 12 36 20 pm

joshuatf:
@timmyc Can you throw a `wp_die()` anywhere in [here](https://github.com/Automattic/wc-calypso-bridge/blob/587dc80003e3ac4c88abd376901f4d03bb79d21a/includes/setup-wizard.php#L31) to see if it's not loading our setup wizard or if something is blocking your admin bar from loading?

Also a heads up that the masterbar will not be fully loaded (see my screenshot above for an example) if the site is not connected to Jetpack.

@justinshreve Good question.  Looks like the `check_param()` wasn't being fired in time.  Should be fixed by https://github.com/Automattic/wc-calypso-bridge/pull/73/commits/587dc80003e3ac4c88abd376901f4d03bb79d21a so we can use that query param.

As for styling, I spent a lot more time than I'm proud of to figure out that the hooks for this page are a bit buggy (possibly due to output buffering?) since it doesn't use the default WP admin page.  Enqueueing a script on `admin_enqueue_scripts` doesn't guarantee that it gets loaded.  

Reordering and adding new hooks for some reason calls some of the scripts ( https://github.com/Automattic/wc-calypso-bridge/pull/73/commits/66e52b4ee68224b0b3764f5f1c2daaa02d250405 ).  That resolves most of our styling issues, but we may end up needing to add a couple fixes to our own stylesheet.  This feels hacky, but I can't find a way around this right now.

joshuatf:
@justinshreve That spacing issue between "My Sites" and "Reader" is not resolved here since the scripts aren't able to be added in via the loader, but I believe that's the only remaining issue and we can throw that into our own calypsoify styles.

Issue #60, opened 2018-10-17
Make sure we load Calypsoify + WooCommerce in the correct context.


justinshreve:
Right now Calypsoify only activates if you hit a specific URL with a specific query string. Per p90Yrv-OY-p2, we might want to force WooCommerce to always be skinned, so users don't get confused with the skin changing.

This task includes:

- [ ] Work with the Jetpack Team to add a hook for loading Calypsoify.
- [ ] Work with the team implementing the onboarding flow to make sure we set some kind of flag/option on the end site during provisioning.
- [ ] Add logic to `wc-calypso-bridge` for loading the Calypsoify + Calypso Bridge code and menu handling in the correct context.
- [ ] Add a user preference (default on for these new stores) but add a UI toggle to allow users to toggle it off if they please.

justinshreve:
I think we can close this out.

A hook for Calyspoify still probably makes sense in general, but is not a blocker for v1 of the eCommerce plan, since we will continue using the query string: https://github.com/Automattic/wc-calypso-bridge/pull/75#issuecomment-432383114. We also will not need a user preference right now.

We have already handled the logic for loading Calypsoify in the correct context.

Finally, there is an issue open for the option during site creation: https://github.com/Automattic/atomic/issues/588.

PR #75, opened 2018-10-23
Add/calypso styles/69

### LINKED ISSUE #69: OBW: Calypsoify Styles
Again this is fully pending hi-fi designs that @josemarques is currently working on. But until then, perhaps we can explore enqueuing the existing calypsoify styles to see how far that can get us. The enqueuing of the styles can be [seen here in Jetpack](https://github.com/Automattic/jetpack/blob/master/modules/class.jetpack-calypsoify.php#L92-L94)

joshuatf:
Fixes #69 (also maybe a part of #60 )

This PR forces the styles and menu hooks used in Jetpack Calyspofiy without the need for the`calypsoify` param in the URL.

### Screenshots

screen shot 2018-10-23 at 11 59 21 am

### Testing

1.  Visit any URL in the dashboard
2.  Calypsoify styles should be on every page

### Potential Changes

* Should we add an option for users to turn this off?
* Is there any context in this plugin which we don't want to force calypsoify?


justinshreve:
During our 1-on-1 the other day, @timmyc and I talked about sticking with the query string for enabling for v1. It's already in use and I think will work fine for initial launch. I should have updated the GitHub issue but I haven't gotten that. cc @timmyc to make sure that is what we still want to do.

Forcing this to always on for all views/plugins (even visiting /wp-admin/ directly) is probably something we want to think about more. And as you asked, if we did that we would probably want a user option.

joshuatf:
@justinshreve Okay, sounds good; I'll ping @timmyc and go ahead and close this issue if that's the case.

Will the query string work without `JETPACK_DEV_DEBUG` enabled?  I believe the masterbar module (and many tools) are only enabled in dev mode currently.

timmyc:
Yeah lets do stick with the query string approach for v1, but I could see us looking at "forcing" it in a subsequent release just because there are so many ways to access the non-skinned version. But before forcing it, we need some input from other folks/design/product to determine how this would best work, and if we expose a site or user setting to toggle the theme etc...

justinshreve:
> Will the query string work without JETPACK_DEV_DEBUG enabled? I believe the masterbar module (and many tools) are only enabled in dev mode currently.

It does if the site is connected (just confirmed as we were DMing and I disconnected my site).

We could always add our version of `check_param` to make sure it's active -- but if the AT site gets disconnected they will have a host of other problems too.

PR #90, opened 2018-10-25
Add phpcs and format files

joshuatf:
This PR adds in phpcs linting and formats a number of the include files.  WPCS rulesets are automatically pulled in and set for phpcs.

screen shot 2018-10-25 at 1 55 32 pm

### Testing

1.  First run `composer install` from your terminal
2.  Run `composer phpcs .` and note errors/warnings for files
3.  Verify no regressions or broken functionality from class renaming.

### Remaining tasks

Here are some other issues that might be best broken down into other PRs.

* Doc blocks are needed for the classes and all files under `includes/connect`
* `store-on-wpcom/` still needs formatting.
* Adding a pre-commit hook to test for lint errors.


joshuatf:
@justinshreve I believe that's a correct response since there are still errors/warnings.  I can't seem to find a reference to the exact error codes, but it will change depending on errors/warnings if some of them are resolved.

Here's a reference to those in phpcbf which should be similar - https://github.com/squizlabs/PHP_CodeSniffer/issues/1270#issuecomment-272768413

justinshreve:
@joshuatf Ok cool. I wasn't sure if it was expected, but I think you are right based off that GH link. LGTM then!

PR #95, opened 2018-10-26
Fix Calypsoify compatibility with WooCommerce and Gutenberg

### LINKED ISSUE #62: Fix post-new.php Gutenberg Calypsoify Loading Issue
If you have Gutenberg and Jetpack Calypsoify installed, all pages using `post-new.php` will have the sidebar completely removed. We don't want this for order, coupon, or product management.

This task includes working with the Jetpack team on resolving this issue, so that we can have the correct chrome on our pages.

justinshreve:
Fixes #62.

This PR requires you to be running the latest Jetpack `master` (which includes a fix I made for the Calypsoify module: see https://github.com/Automattic/jetpack/pull/10389).

This PR uses the hook added there, to disable Calypsoify from overtaking the screen for WooCommerce post types. All the post types here are from the supported list of extensions. I updated the directions for Calypsoifying extensions to include this step: p90Yrv-Po-p2.

To Test:
* Run latest Jetpack
* Install Gutenberg
* Visit the add new product screen, coupon screen, bookings add resource screen, or another extension like Boxoffice and see that the navigation stays and Gutenberg no longer overtakes the page

### DIFF
--- a/includes/gutenberg.php
+++ b/includes/gutenberg.php
@@ -15,7 +15,7 @@ function wc_calypso_bridge_disable_gutenberg_for_post_type( $current_value, $pos
 		'wc_product_tab',
 		'wishlist',
 		'wc_zapier_feed',
- );
+	);
 	if ( in_array( $post_type, $wc_post_types ) ) {
 		return false;
 	}
--- a/includes/gutenberg.php
+++ b/includes/gutenberg.php
@@ -1,0 +1,24 @@
+<?php
+function wc_calypso_bridge_disable_gutenberg_for_post_type( $current_value, $post_type ) {
+	$wc_post_types = array(
+		'shop_coupon',
+		'shop_order',
+		'product',
+		'bookable_resource',
+		'wc_booking',
+		'event_ticket',
+		'wc_membership_plan',
+		'wc_user_membership',
+		'wc_voucher',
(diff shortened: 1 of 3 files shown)


includes/gutenberg.php
```diff
@@ -0,0 +1,24 @@
...
+		'event_ticket',
+		'wc_membership_plan',
+		'wc_user_membership',
+		'wc_voucher',
+		'wc_pickup_location',
+		'shop_subscription',
+		'wc_product_tab',
+		'wishlist',
+		'wc_zapier_feed',
+ );
```
joshuatf: Indentation needed here.

justinshreve:
> This may be a stupid question, but how many post types will use Gutenberg? Would it be easier to whitelist the post types that can use it then blacklist those that can't?

That was my original attempt for the Jetpack branch, but WordPress 5.0 (where Gutenberg gets merged) does a blacklist approach at the moment, so I updated the Jetpack branch to use the same method/filter instead of introducing our own special Jetpack filter.

I'm thinking this could lead to some issues with other plugins (not just WooCommerce post types) since many of them won't be block ready -- so I think I will leave some feedback to WordPress core next week.

Otherwise, I think we can use this in the meantime.

Issue #64, opened 2018-10-17
Setup Task List


justinshreve:
This task involves creating a setup task list screen to help walk our new eCommerce users through setup of their store. 

On the tech side of things, we have agreed to keep v1 basic and just copy HTML/CSS from Calypso and fit it to our needs.

* [x] Create base task list based on Calypso styles
* [x] Implement static task text and action buttons
* [x] Create a method for tracking if a task has been completed
* [x] Create a method for dismissing the entire setup list

PR #92, opened 2018-10-25
Hide notices in wp-admin

jeffikus:
Hides all plugin activation and wp-admin admin notices in the calypsoified wp-admin.

Closes #65

PR #97, opened 2018-10-26
Adds setup checklist base

justinshreve:
This PR adds a new base page for the setup checklist (see #64). It adds a 'Setup' page to the top of the WooCommerce navigation. It also hides all notices on the page, so we can have a nice clean task list.

There is not much to see in this PR, but this figured like a good cut-off point for review before actually building the task-list code.

screen shot 2018-10-26 at 10 45 12 am

To Test:
* Verify you see the 'Setup' link under WooCommerce, above orders
* Verify the page loads with the correct navigation, and is totally blank (no notifications)
* Set `atomic-ecommerce-setup-checklist-complete` to true, and the page should go away.


includes/class-wc-calypso-bridge-setup-checklist.php
```diff
@@ -0,0 +1,127 @@
...
+		}
+
+		add_action( 'admin_menu', array( $this, 'admin_menu' ) );
+
+		// priority is 20 to run after https://github.com/woocommerce/woocommerce/blob/a55ae325306fc2179149ba9b97e66f32f84fdd9c/includes/admin/class-wc-admin-menus.php#L165.
+		add_action( 'admin_head', array( $this, 'admin_menu_structure' ), 20 );
+
+		// run early and late
+		if ( isset( $_GET['page'] ) && 'wc-setup-checklist' === $_GET['page'] ) {
+			add_action( 'admin_notices', array( $this, 'hide_all_notices_on_setup_page' ), 0 );
```
joshuatf: This is a cool solution.  Could we simplify this to `remove_all_actions( 'admin_notices' );` or is there something else this function does?

includes/class-wc-calypso-bridge-setup-checklist.php
```diff
@@ -0,0 +1,127 @@
...
+if ( ! defined( 'ABSPATH' ) ) {
+	exit;
+}
+
+/**
+ * WC_Calypso_Bridge_Admin_Setup_Checklist class.
+ */
+class WC_Calypso_Bridge_Admin_Setup_Checklist {
+
+	static $instance = false;
```
joshuatf: Probably needs a `protected` scope modifier.

includes/class-wc-calypso-bridge-setup-checklist.php
```diff
@@ -0,0 +1,127 @@
...
+			'wc-setup-checklist',
+			array( $this, 'checklist' )
+		);
+	}
+
+	/**
+	 * Puts the 'Setup' menu item at the very top of the WooCommerce link.
+	 * We have to do some shuffling, because WooCommerce does some overwriting with the 'Orders' link.
+	 */
+	public function admin_menu_structure() {
```
joshuatf: Nice!

includes/class-wc-calypso-bridge-setup-checklist.php
justinshreve: Unfortunately, not all notices are hooked the same way (see https://github.com/Automattic/wc-calypso-bridge/pull/92/ for a few of the different ways that need done to selectively unhook things). I gave this a try to make sure, but I still had some notices come through -- so I think this might be the best way to hide everything.

justinshreve:
@joshuatf I added a `protected` and tried out the remove_all_actions, but it didn't fully work (see comment above). OK to proceed with what I have?

includes/class-wc-calypso-bridge-setup-checklist.php
joshuatf: Can you try this?  Only tested with a couple but seemed to remove all my notices including those from Canada Post.

```php
	/**
	 * Hooks into WordPress to add our new setup checklist.
	 */
		// If setup has been completed, do nothing.
		if ( true === (bool) get_option( 'atomic-ecommerce-setup-checklist-complete', false ) ) {
			return;
		}

		add_action( 'admin_menu', array( $this, 'admin_menu' ) );
		add_action( 'admin_head', array( $this, 'remove_notices' ) );
		// priority is 20 to run after https://github.com/woocommerce/woocommerce/blob/a55ae325306fc2179149ba9b97e66f32f84fdd9c/includes/admin/class-wc-admin-menus.php#L165.
		add_action( 'admin_head', array( $this, 'admin_menu_structure' ), 20 );
	}

	/**
	 * Remove all admin notices
	 */
	public function remove_notices() {
		remove_all_actions( 'admin_notices' );
	}
```

joshuatf:
@justinshreve Pre-approving changes.  If https://github.com/Automattic/wc-calypso-bridge/pull/97#discussion_r229037393 does not work to remove all notices, LGTM!

justinshreve:
Thanks @joshuatf! https://github.com/Automattic/wc-calypso-bridge/pull/97#discussion_r229037393 ended up working.

justinshreve:
Oops, I accidentally left the unused functions in there, but I'll remove them in my next PR.

PR #126, opened 2018-10-31
Forms design updates

josemarques:
screen shot 2018-10-31 at 17 06 24

Product edit:
- Adds card with the title and description fields in the  

screen shot 2018-10-31 at 17 07 33

Settings:
- Adds cards to match Calypso's cards patterns
- Updates typography in titles and description and makes it a header section

### DIFF
--- a/assets/css/calypsoify.css
+++ b/assets/css/calypsoify.css
@@ -326,13 +326,17 @@ table.widefat {
 }
 
 #title-prompt-text {
+	clip: initial !important; /*avoid clipping when field is on focus */
+	clip-path: initial;
 	color: #2e4453 !important;
 	display: block;
 	font-size: 14px !important;
 	font-weight: 600;
-	margin-bottom: 5px !important;
+	height: auto;
+	margin: 0 0 5px 0 !important;
 	padding: 0 !important;
 	position: static !important;
+	width: auto;
 }
 
 #wp-content-editor-tools {
--- a/assets/css/calypsoify.css
+++ b/assets/css/calypsoify.css
@@ -339,6 +339,78 @@ table.widefat {
 	background: #fff;
 }
 
+/* Adds card layout style to forms */
+
+#mainform h2,
+#mainform h3 {
+	background-color: #fff;
+	border: 1px solid rgba(200,215,225,0.5);
+	border-bottom: none;
+	color: #2e4453;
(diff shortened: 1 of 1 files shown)


assets/css/calypsoify.css
```diff
@@ -316,6 +316,105 @@ table.widefat {
 	color: #2E4453;
 }
 
+/* Forms */
+
+#post-body-content {
+	background-color: #fff;
+	box-shadow: 0 0 0 1px rgba(200,215,225,0.5), 0 1px 2px #e9eff3;
```
joshuatf: Can we simplify this to a `border: 1px solid #hex` here?  It appears to be a single solid border around its edge, but I may be missing some details here.

assets/css/calypsoify.css
```diff
@@ -316,6 +316,105 @@ table.widefat {
...
+#wp-content-editor-tools {
+	background: #fff;
+}
+
+/* Adds card layout style to forms */
+
+#mainform h2,
+#mainform h3 {
+	background-color: #fff;
+	border: 1px solid rgba(200,215,225,0.5);
```
joshuatf: Do these need to be rgba or can they be hex?

assets/css/calypsoify.css
```diff
@@ -316,6 +316,105 @@ table.widefat {
...
+
+#post-body-content {
+	background-color: #fff;
+	box-shadow: 0 0 0 1px rgba(200,215,225,0.5), 0 1px 2px #e9eff3;
+	box-sizing: border-box;
+	padding: 24px;
+}
+
+#title-prompt-text {
+	clip: initial !important; /*avoid clipping when field is on focus */
```
joshuatf: Nice work on the clipping property!  If you change the selector here to `#titlediv #title-prompt-text` we can probably remove all of those important tags unless you saw something else overriding these properties.

assets/css/calypsoify.css
josemarques: I copied the value from .nav-tab-wrapper. But we can use #e4eaef.

assets/css/calypsoify.css
josemarques: Will do.

josemarques:
@joshuatf were these changes squashed? I'm not seeing the update on the forms on master.

PR #129, opened 2018-10-31
Add calypso notice styling

### LINKED ISSUE #118: Update notices to match calypso
Notices currently look like this:

screen shot 2018-10-31 at 11 27 09

They should follow the calypso styles:

screen shot 2018-10-31 at 11 28 17

### LINKED ISSUE #122: Nav tabs and sub nav should not be separated by notifications
Admin notifications separate the nav tabs and sub nav currently.  Need to find a way to separate these.  

screen shot 2018-10-31 at 10 34 50 am

If the notifications are positioned absolutely or fixed (Calypso style) by https://github.com/Automattic/wc-calypso-bridge/issues/118 this will probably resolve this issue.

screen shot 2018-10-31 at 10 41 17 am

joshuatf:
Fixes #118 , fixes (maybe) #122 

This PR adds styling to notices to make them more like Calypso notices.  

### Screenshots
screen shot 2018-11-01 at 9 14 11 am
screen shot 2018-11-01 at 9 31 05 am

### Testing
1.  Visit any pages with notices with `calypsoify=1` enabled.
2.  Check that notices appear correctly with respective styles (warning, error, success, notice).
3.  Check that notices fix to bottom of screen on mobile (<480px).

### Notes

* This PR hides all notices that aren't moved into the `wc-calypso-bridge-notices` container to prevent FOUC.  It does this after the document is ready so it should capture nearly all notices, but if a plugin author added a notice via AJAX, it would most likely be hidden.
* With many notices, the screen can easily become unusable (see screenshots).  We can:
    1. Make the notice container not fixed and less Calypso like (default area for notices, but still styled like Calypso
    2. Add a dismissable button that would close the notice to temporarily make the screen more usable, however the notice would persist on next refresh.
    3. Other ideas?

@josemarques What are your thoughts on the last point?  Any other changes you think need to be made here?

### DIFF
--- a/assets/js/calypsoify.js
+++ b/assets/js/calypsoify.js
@@ -42,5 +42,20 @@
             );
         }
     } );
+    
+    /**
+     * Move admin notices
+     */
+    $( document ).ready( function() {
+        $( 'body' ).append( '<div class="wc-calypso-bridge-notices"></div>' );
+        $( 'div.updated, div.error, div.notice' ).not( '.inline, .below-h2' ).appendTo( $( '.wc-calypso-bridge-notices' ) );
+    } );
+
+    /**
+     * Append icons to notices
+     */
+    $( 'div.notice:not(.notice-success), div.updated:not(.notice-success)' ).prepend( '<span class="wc-calypso-bridge-notice-icon-wrapper"><svg class="gridicon gridicons-info" height="24" width="24" xmlns="http://www.w3.org/2000/svg" viewBox="0 0 24 24"><g><path d="M12 2C6.477 2 2 6.477 2 12s4.477 10 10 10 10-4.477 10-10S17.523 2 12 2zm1 15h-2v-6h2v6zm0-8h-2V7h2v2z"/></g></svg></span>' );
+    $( 'div.error' ).prepend( '<span class="wc-calypso-bridge-notice-icon-wrapper"><svg class="gridicon gridicons-notice" height="24" width="24" xmlns="http://www.w3.org/2000/svg" viewBox="0 0 24 24"><g><path d="M12 2C6.477 2 2 6.477 2 12s4.477 10 10 10 10-4.477 10-10S17.523 2 12 2zm1 15h-2v-2h2v2zm0-4h-2l-.5-6h3l-.5 6z"/></g></svg></span>' );
+    $( 'div.notice-success' ).prepend( '<span class="wc-calypso-bridge-notice-icon-wrapper"><svg class="gridicon gridicons-checkmark" height="24" width="24" xmlns="http://www.w3.org/2000/svg" viewBox="0 0 24 24"><g><path d="M9 19.414l-6.707-6.707 1.414-1.414L9 16.586 20.293 5.293l1.414 1.414"/></g></svg></span>' );
 
 } )( jQuery );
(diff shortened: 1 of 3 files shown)


assets/js/calypsoify.js
```diff
@@ -0,0 +1,43 @@
...
+            icon = '<svg class="gridicon gridicons-notice" height="24" width="24" xmlns="http://www.w3.org/2000/svg" viewBox="0 0 24 24"><g><path d="M12 2C6.477 2 2 6.477 2 12s4.477 10 10 10 10-4.477 10-10S17.523 2 12 2zm1 15h-2v-2h2v2zm0-4h-2l-.5-6h3l-.5 6z"/></g></svg>';
+        }
+        $( this ).prepend( '<span class="wc-calypso-bridge-notice-icon-wrapper">' + icon + '</span>' );
+    } );
+
+    /**
+     * Replace dismissal buttons in notices
+     */
+    $( document ).ready( function() {
+        $( '.notice-dismiss' ).html( '<svg class="gridicon gridicons-cross" height="24" width="24" xmlns="http://www.w3.org/2000/svg" viewBox="0 0 24 24"><g><path d="M18.36 19.78L12 13.41l-6.36 6.37-1.42-1.42L10.59 12 4.22 5.64l1.42-1.42L12 10.59l6.36-6.36 1.41 1.41L13.41 12l6.36 6.36z"/></g></svg>' );
```
justinshreve: I don't think we should add dismissal buttons unless we can persist these across page views. I think it would be fine to just remove this for now.

justinshreve: There are some notices that do have persistable dismiss buttons. The placement of the 'X' is off for those though

Here's a screenshot:

screen shot 2018-11-01 at 11 10 05 am

joshuatf: Nice find!  I added some styling to override the WooCommerce dismiss styles. 

This JS doesn't make any assumptions or add dismiss buttons if they don't exist; it's only responsible for replacing the dismiss button with the SVG used in Calypso so I think it's safe to use- do you agree?

joshuatf:
>My vote would be for Make the notice container not fixed and less Calypso like (default area for notices, but still styled like Calypso).

@justinshreve I tend to agree with you on this since I think we might be creating more problems with the fixed/floating style.  If we remove it, it won't fix https://github.com/Automattic/wc-calypso-bridge/issues/122 but we can handle this separately.

@josemarques Can you let me know if you have strong opinions on this one way or another when you get a chance?

assets/js/calypsoify.js
justinshreve: > it's only responsible for replacing the dismiss button with the SVG used in Calypso so I think it's safe to use- do you agree?

Sorry my bad. I think I just confused myself during testing because some didn't persist, but you're right that this just replaces them. Non-persisting notices are the plugins fault, so this line is fine with me :).

joshuatf:
@josemarques Thanks!  Just to confirm- you want to keep them fixed on the screen at the top right while we scroll, right?

@justinshreve Can you ✅ if everything code-wise is up to snuff?

josemarques:
@joshuatf I meant not fixed as Justin suggested. Sorry about that.

joshuatf:
@timmyc Just rebased.  Can you sanity check design after unfixing the notifications in https://github.com/Automattic/wc-calypso-bridge/pull/129/commits/e022a2f4e276ef0039c8e40484c0ec21486ed3ee ?

PR #130, opened 2018-11-01
Setup checklist: Header/footer, hide completed, "I'm done"

justinshreve:
Closes #64.

This PR adds a header to the setup page, based on the mock up. It also adds an "I'm done" button that will dismiss the entire page. The "hide/show" completed task functionality present in the Calypso component also works now.

I've added a bit of code to turn Calypsoify on if the check list is visited directly. The list will be linked to from an onboarding email, and if Calypsoify somehow  gets turned off, it would be good to turn it back on.

screen shot 2018-11-01 at 2 20 08 pm

screen shot 2018-11-01 at 2 20 15 pm

screen shot 2018-11-01 at 2 20 20 pm

To Test:
* Disable Calypsoify.
* Visit `wp-admin/admin.php?page=wc-setup-checklist` directly, verify Calypsoify is turned on and you can see the checklist.
* Note the new header.
* Test the hide/show functionality.
* Test the "I'm done" button in the footer. You should be redirected to the orders page (later we can make this wc-admin).

PR #133, opened 2018-11-01
Track events inside OBW and checklist

### LINKED ISSUE #85: Add Track Events
It appears that we can now use Jetpack to record non-`jetpack` prefixed events! Awesome! There are two helper utils for making recording tracks in our plugin quite easy:

- PHP tracks client [usage example](https://github.com/Automattic/jetpack/blob/f8078c2cd12ac508334da2fb08e37a92cf283c14/_inc/lib/tracks/client.php#L7-L14) enables using `jetpack_tracks_record_event()` from within PHP
- [JS Ajax client](https://github.com/Automattic/jetpack/blob/f8078c2cd12ac508334da2fb08e37a92cf283c14/_inc/lib/tracks/tracks-ajax.js#L6)  provides an ajax helper for recording tracks. I think using the approach we have setup in wc-admin might be a bit nicer from JS, but lets use what we have here via Jetpack.

I've created a spreadsheet for us to add in all the tracks we are recording as part of this project. I have yet to create entries for all the task list steps, but lets just update the spreadsheet as we go along:

[https://docs.google.com/spreadsheets/d/17yZeaWktUSq_lf6fVhe61hyTCXO_09gKmqXdcUzGj18/edit?usp=sharing](https://docs.google.com/spreadsheets/d/17yZeaWktUSq_lf6fVhe61hyTCXO_09gKmqXdcUzGj18/edit?usp=sharing)

I opted to standardize on the `atomic_` prefix for these track names, since it is already whitelisted, and these are indeed all atomic sites. /cc @jhnstn to make sure that is an okay namespace to use for tracking events in this project.

Note: To get Jetpack tracks helpers working in your local install, you might need to manually set the following site option:

`update_option( 'jetpack_tos_agreed', true );`

joshuatf:
Fixes #85 

This PR adds event tracking for the following events:
* OBW step completion
* Tasklist item clicked
* Tasklist finished (clicked "I'm done")
* Edit address in OBW clicked
* Calypsoify manually turned on/off

### Screenshots
screen shot 2018-11-01 at 3 41 10 pm

### Testing
1.  Make sure you have a Jetpack connected site with tracking enabled.
2.  Try the various above events to make sure they come through (may take a few minutes to show up) - https://mc.a8c.com/tracks/live/?eventname=%25atomic_%25&user=&useragent=

### Notes
* The javascript events are prefixed automatically with `jetpack_`
* The task list completion event is dependent on https://github.com/Automattic/wc-calypso-bridge/pull/130

### DIFF
--- a/assets/js/calypsoify.js
+++ b/assets/js/calypsoify.js
@@ -19,4 +19,24 @@
         );
     } );
 
+    /**
+     * Track 'I'm done' completion on task list
+     */
+    $( '.setup-footer a' ).click( function(e) {
+        e.preventDefault();
+        var progressNumber = $( '.checklist__header-progress-number' ).text().split( '/' );
+        var complete = progressNumber[0];
+        var total = progressNumber[1];
+        var percentage = parseFloat( complete / total ).toFixed( 2 ) * 100;
+        window.jpTracksAJAX.record_ajax_event(
+            'atomic_wc_tasklist_finished',
+            'click',
+            { 
+                complete: complete,
+                total: total,
+                percentage: percentage
+            }
+        );
+    } );
+
 } )( jQuery );
--- a/assets/js/calypsoify.js
+++ b/assets/js/calypsoify.js
@@ -7,12 +7,14 @@
     $( '.checklist__task-title a, .checklist__task-secondary a' ).click( function() {
         var $task = $( this ).closest( '.checklist__task' )
         var status = $task.hasClass( 'is-completed' ) ? 'complete' : 'incomplete';
+        var taskId = $task.data('id');
         var taskTitle = $task.data('title');
 
         window.jpTracksAJAX.record_ajax_event(
             'atomic_wc_tasklist_click',
             'click',
-            { 
+            {
+                id: parseInt( taskId ),
                 title: taskTitle, 
                 status: status,
             }
--- a/assets/js/calypsoify.js
+++ b/assets/js/calypsoify.js
@@ -10,7 +10,7 @@
         var taskTitle = $task.data('title');
 
         window.jpTracksAJAX.record_ajax_event(
-            'atomic_wc_tasklist_clicked',
+            'atomic_wc_tasklist_click',
             'click',
             { 
                 title: taskTitle, 
@@ -29,7 +29,7 @@
         var total = progressNumber[1];
         var percentage = parseFloat( complete / total ).toFixed( 2 ) * 100;
         window.jpTracksAJAX.record_ajax_event(
-            'atomic_wc_tasklist_finished',
+            'atomic_wc_tasklist_finish',
             'click',
             { 
                 complete: complete,
--- a/assets/js/calypsoify.js
+++ b/assets/js/calypsoify.js
@@ -14,7 +14,7 @@
             'atomic_wc_tasklist_click',
             'click',
             {
-                id: parseInt( taskId ),
+                id: taskId,
                 title: taskTitle, 
                 status: status,
             }
--- a/assets/js/calypsoify.js
+++ b/assets/js/calypsoify.js
@@ -1,0 +1,22 @@
+( function( $ ) {
+    'use strict';
+
+    /**
+     * Record checklist task click
+     */
+    $( '.checklist__task-title a, .checklist__task-secondary a' ).click( function() {
+        var $task = $( this ).closest( '.checklist__task' )
+        var status = $task.hasClass( 'is-completed' ) ? 'complete' : 'incomplete';
+        var taskTitle = $task.data('title');
+
+        window.jpTracksAJAX.record_ajax_event(
+            'atomic_wc_tasklist_clicked',
+            'click',
+            { 
+                title: taskTitle, 
+                status: status,
+            }
+        );
+    } );
+
+} )( jQuery );
--- a/assets/js/calypsoify.js
+++ b/assets/js/calypsoify.js
@@ -30,15 +30,17 @@
         var complete = progressNumber[0];
         var total = progressNumber[1];
         var percentage = parseFloat( complete / total ).toFixed( 2 ) * 100;
-        window.jpTracksAJAX.record_ajax_event(
-            'atomic_wc_tasklist_finish',
-            'click',
-            { 
-                complete: complete,
-                total: total,
-                percentage: percentage
-            }
-        );
+        if (window.jpTracksAJAX) {
+            window.jpTracksAJAX.record_ajax_event(
+                'atomic_wc_tasklist_finish',
+                'click',
+                { 
+                    complete: complete,
+                    total: total,
+                    percentage: percentage
+                }
+            );
+        }
     } );
 
 } )( jQuery );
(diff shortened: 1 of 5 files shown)


class-wc-calypso-bridge.php
```diff
@@ -91,9 +92,31 @@ public static function instance() {
...
+	 * Track Calypsoify events when turned on or off
+	 */
+	public function track_calypsoify_toggle() {
+		if ( isset( $_GET['calypsoify'] ) ) { // WPCS: CSRF ok.
+			$current_user      = wp_get_current_user();
+			$calypsoify_status = (int) get_user_meta( $current_user->ID, 'calypsoify', true );
+			if ( 1 === $calypsoify_status && 0 === (int) $_GET['calypsoify'] ) { // WPCS: CSRF ok.
+				jetpack_tracks_record_event(
+					$current_user,
+					'atomic_wc_calypsoify_off'
```
justinshreve: While not explicitly addressed in the "Tracks Event Design & Naming Conventions" field guide page, there is a section in the FG that talks about not using variables in the name such as post type, and instead use a single event with a prop. I feel like that can apply here. 

So I'm wondering if maybe we should name this `atomic_wc_calypsoify_toggle` instead with a prop for off/on.

assets/js/calypsoify.js
```diff
@@ -0,0 +1,42 @@
...
+    /**
+     * Record checklist task click
+     */
+    $( '.checklist__task-title a, .checklist__task-secondary a' ).click( function() {
+        var $task = $( this ).closest( '.checklist__task' )
+        var status = $task.hasClass( 'is-completed' ) ? 'complete' : 'incomplete';
+        var taskTitle = $task.data('title');
+
+        window.jpTracksAJAX.record_ajax_event(
+            'atomic_wc_tasklist_clicked',
```
justinshreve: This should be `atomic_wc_tasklist_click`

> use a singular noun or present tense verb

assets/js/calypsoify.js
```diff
@@ -0,0 +1,42 @@
...
+     * Track 'I'm done' completion on task list
+     */
+    $( '.setup-footer a' ).click( function(e) {
+        e.preventDefault();
+        var progressNumber = $( '.checklist__header-progress-number' ).text().split( '/' );
+        var complete = progressNumber[0];
+        var total = progressNumber[1];
+        var percentage = parseFloat( complete / total ).toFixed( 2 ) * 100;
+        window.jpTracksAJAX.record_ajax_event(
+            'atomic_wc_tasklist_finished',
```
justinshreve: This should be `atomic_wc_tasklist_finish`

> use a singular noun or present tense verb

assets/js/calypsoify.js
```diff
@@ -0,0 +1,42 @@
+( function( $ ) {
+    'use strict';
+
+    /**
+     * Record checklist task click
+     */
+    $( '.checklist__task-title a, .checklist__task-secondary a' ).click( function() {
+        var $task = $( this ).closest( '.checklist__task' )
+        var status = $task.hasClass( 'is-completed' ) ? 'complete' : 'incomplete';
+        var taskTitle = $task.data('title');
```
justinshreve: We might want to add a normalized, untranslated id to the array of task fields instead and use that here (or both).

I know for the initial launch none of these tasks are translated, but in the future the task titles could be changed/translated and we'll still want to associate those events together.

assets/js/calypsoify.js
```diff
@@ -0,0 +1,42 @@
...
+        var complete = progressNumber[0];
+        var total = progressNumber[1];
+        var percentage = parseFloat( complete / total ).toFixed( 2 ) * 100;
+        window.jpTracksAJAX.record_ajax_event(
+            'atomic_wc_tasklist_finished',
+            'click',
+            { 
+                complete: complete,
+                total: total,
+                percentage: percentage
```
justinshreve: Great idea passing the percentage here! 👍

class-wc-calypso-bridge.php
```diff
@@ -91,9 +92,31 @@ public static function instance() {
...
+				);
+			} elseif ( 0 === $calypsoify_status && 1 === (int) $_GET['calypsoify'] ) { // WPCS: CSRF ok.
+				jetpack_tracks_record_event(
```
justinshreve: It looks like we need to manually pass the `blog_id` prop for PHP events. You can get the blog id from `Jetpack_Options::get_option( 'id' );`.

The Javascript events already seem to include it.

class-wc-calypso-bridge.php
joshuatf: That seems like a good idea plus we can clean up the logic here a bit.  https://github.com/Automattic/wc-calypso-bridge/pull/133/commits/001ffba4bc7f194019218b4ee1db1322d09ab74b

assets/js/calypsoify.js
joshuatf: Updated!

assets/js/calypsoify.js
joshuatf: Updated!

assets/js/calypsoify.js
joshuatf: Ah good thinking!  Let me know if this looks reasonable - https://github.com/Automattic/wc-calypso-bridge/pull/133/commits/6e076c1a6b2ad3aea6cf13ba9c847b790a193f23

timmyc:
This is looking great @joshuatf! The only thing I'm wondering about is if we should first do a null check on `window.jpTracksAJAX` to ensure it is defined. While _highly unlikely_ this will ever be undefined in the context of these AT sites, if somehow the site gets disconnected from Jetpack, I believe that particular global is not defined.

includes/class-wc-calypso-bridge-admin-setup-wizard.php
```diff
@@ -107,6 +107,15 @@ public function setup_wizard() {
 
 		// @codingStandardsIgnoreStart
 		if ( ! empty( $_POST['save_step'] ) && isset( $this->steps[ $this->step ]['handler'] ) ) {
+			$current_user = wp_get_current_user();
+			jetpack_tracks_record_event(
```
timmyc: I suppose if we are being ultra paranoid around checking to see if the JS helper is defined we should check here too.

assets/js/calypsoify.js
```diff
@@ -0,0 +1,44 @@
+( function( $ ) {
+    'use strict';
+
+    /**
```
timmyc: I suppose we could just wrap these in a `if (window.jpTracksAJAX)` but i'm likely being too paranoid :)

joshuatf: Seems like a safe idea 😄

includes/class-wc-calypso-bridge-admin-setup-wizard.php
joshuatf: Moved this into its own function to dry things up a bit - https://github.com/Automattic/wc-calypso-bridge/pull/133/commits/9e06bcf64eba90d75012341b77a9ffc69d92bc25 . Let me know what you think.

justinshreve:
> Cleaned up a bit and made those changes. Let me know if you think we need to change anything else here.

Looking good @joshuatf! I left one more comment about the id naming here:
https://github.com/Automattic/wc-calypso-bridge/pull/133#discussion_r230413976

joshuatf:
@justinshreve Just pushed a change for the IDs.  Let me know how we're looking now!

class-wc-calypso-bridge.php
```diff
@@ -108,6 +110,42 @@ public function check_setup_param() {
...
+		}
+	}
+
+	/**
+	 * Record event using JetPack if enabled
+	 *
+	 * @param string $event_name Name of the event.
+	 * @param array  $event_params Custom event params to capture.
+	 */
+	public static function record_event( $event_name, $event_params ) {
```
justinshreve: Nice 👍

joshuatf:
Awesome!  Thanks for the all the input @justinshreve and @timmyc 😃

PR #142, opened 2018-11-02
Add an early param check for calypsoify

### LINKED ISSUE #141: Calypsoify param does not work on wc-setup page
Navigating to `/wp-admin/admin.php?page=wc-setup` and appending the `calypsoify=1` param to the URL does not work.

Appending `calypsoify=1` should work on the wc-setup page since this may be the first page a user lands on coming from Calypso.

screen shot 2018-11-02 at 3 22 57 pm

joshuatf:
Fixes #141 

This PR adds the Jetpack Calypsoify param check to the `init` hook so we can check if this is set on certain pages (i.e., the wc-setup page).

It also fixes the issue where loading a page with the `calypsoify=1` param requires a refresh to get Calypsoify scripts running.

### Testing
1.  Visit `/wp-admin/edit.php?post_type=shop_order&calypsoify=1` and verify the calypsoify is turned on.
2.  Visit `/wp-admin/edit.php?post_type=shop_order&calypsoify=0` and verify the calypsoify is turned off.
3.  Visit `/wp-admin/admin.php?page=wc-setup&calypsoify=1` and verify the calypsoify is turned on.
4.  Visit `/wp-admin/admin.php?page=wc-setup&calypsoify=0` and verify the calypsoify is turned off.

### DIFF
--- a/class-wc-calypso-bridge.php
+++ b/class-wc-calypso-bridge.php
@@ -36,8 +36,31 @@ class WC_Calypso_Bridge {
 	 * Constructor.
 	 */
 	public function __construct() {
-		add_action( 'init', array( $this, 'possibly_load_calypsoify' ), 1 );
+		add_action( 'init', array( $this, 'check_calyposify_param' ), 1 );
 		add_action( 'init', array( $this, 'check_setup_param' ) );
+		add_action( 'init', array( $this, 'possibly_load_calypsoify' ), 2 );
+	}
+
+	/**
+	 * Check for calypsoify param in URL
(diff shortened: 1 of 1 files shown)


class-wc-calypso-bridge.php
```diff
@@ -109,5 +132,6 @@ public function check_setup_param() {
 	}
 
 }
-
-WC_Calypso_Bridge::instance();
+if ( is_admin() ) {
```
timmyc: Nice.

PR #150, opened 2018-11-06
Hide Woo OBW admin alert.

### LINKED ISSUE #144: Hide OBW Admin Alert
We should filter off the default OBW wizard admin alert that is shown when a site is first setup:

image

Looks like we can filter that off [here](https://github.com/woocommerce/woocommerce/blob/d0491072e8a7d0a4535db0ec2bae4a6d43ff64fd/includes/admin/class-wc-admin-notices.php#L180)

timmyc:
Fixes #144 

This PR adds in a new class where we can keep any random logic needed to filter off admin alerts from being shown in the ecomm plan. The first of which is the OBW alert that is added by Woo Core to freshly installed Woo sites.

__Before__
notice-before

__After__
notice-after

__To Test__
You can force the alert to be shown on any site by adding the following code somewhere:

`WC_Admin_Notices::add_notice( 'install' );`

And refreshing a calypsoified woo page to verify you see the alert as pictured in the "Before" grab above. Then checkout this branch and verify it is no longer shown.

### DIFF
--- a/includes/class-wc-calypso-bridge-hide-alerts.php
+++ b/includes/class-wc-calypso-bridge-hide-alerts.php
@@ -14,38 +14,38 @@ defined( 'ABSPATH' ) || exit;
  */
 class WC_Calypso_Bridge_Hide_Alerts {
 
-    /**
-     * Class instance.
-     *
-     * @var WC_Calypso_Bridge_Hide_Alerts instance
-     */
-    protected static $instance = false;
+	/**
+	 * Class instance.
+	 *
+	 * @var WC_Calypso_Bridge_Hide_Alerts instance
+	 */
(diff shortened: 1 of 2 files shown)


includes/class-wc-calypso-bridge-hide-alerts.php
```diff
@@ -0,0 +1,51 @@
...
+ */
+
+defined( 'ABSPATH' ) || exit;
+
+/**
+ * WC Calypso Bridge Hide Alerts
+ */
+class WC_Calypso_Bridge_Hide_Alerts {
+
+    /**
```
joshuatf: WPCS nitpicking but looks like we need tabs here instead of spaces 🔔

PR #161, opened 2018-11-12
Add calypso styling to form inputs

### LINKED ISSUE #148: Restyle inputs to be more Calypso-like
Form styling is still very much the wp-admin default.  Restyle all default inputs to be more like Calypso.

### Before
screen shot 2018-11-06 at 4 30 09 pm

### After
screen shot 2018-11-06 at 4 30 37 pm

joshuatf:
This PR adds Calypso styling to all form inputs.

Fixes #148 

#### Screenshots
screen shot 2018-11-12 at 9 23 25 am
screen shot 2018-11-12 at 9 23 15 am

#### Testing
1. Visit pages with form inputs (e.g. Product edit, and Categories pages).
2. Check that input styling is correct and doesn’t break existing form fields.

@josemarques Since the inputs are 40px tall and this PR increases the height of all select boxes to 40px, I also adjusted the buttons in bulk actions to 40px in height (see the bulk action select options screenshot above).  Can you let me know if this looks okay or if we should make both the buttons and inputs back to the previous height of 28px?

### DIFF
--- a/assets/css/calypsoify.css
+++ b/assets/css/calypsoify.css
@@ -170,6 +170,18 @@ textarea:focus {
 	-moz-outline: none;
 	-moz-user-focus: ignore;
 }
+input[type="text"]::placeholder,
+input[type="search"]::placeholder,
+input[type="email"]::placeholder,
+input[type="number"]::placeholder,
+input[type="password"]::placeholder,
+input[type=checkbox]::placeholder,
+input[type=radio]::placeholder,
+input[type="tel"]::placeholder,
+input[type="url"]::placeholder,
+textarea::placeholder {
+    color: #87a6bc;
+}
 .form-wrap {
 	display: block;
 	margin: 0 auto 10px auto;
--- a/assets/css/calypsoify.css
+++ b/assets/css/calypsoify.css
@@ -116,6 +116,52 @@ textarea {
 	box-shadow: none;
 	height: auto;
 }
+input[type="text"]:focus,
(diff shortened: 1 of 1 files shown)


josemarques:
It's looking ok in the filters but in need a bit padding between the filters and the table:

screen shot 2018-11-13 at 10 08 34

I've noticed some funkyness in a couple of places:

screen shot 2018-11-13 at 10 06 49
screen shot 2018-11-13 at 10 06 43

joshuatf:
@justinshreve @josemarques Added styling for select2 dropdowns as well and matched button heights next to all select dropdowns.  That should resolve all those issues, but let me know if you see any others.

timmyc:
I imagine we are going to find other visual gotchas with this set of style changes. Should we get this merged in as-is and open up follow-up bugs as we see things pop up? I think we have to either merge this in, and anticipate we will find more glitches like this, or hold off on doing this for v1.  @josemarques thoughts?

justinshreve:
The check-boxes feels like a blocker, but otherwise I think we could merge this and follow-up with anything else.

joshuatf:
@timmyc @justinshreve I'll push some changes for the checkboxes shortly so we can get this merged and we can address other page specific issues as we find them.

joshuatf:
@justinshreve Pushed those changes for the checkboxes and also radio buttons.  Can you take a look and let me know if things are looking okay now?

PR #164, opened 2018-11-12
Calypso style search and subnav search

### LINKED ISSUE #151: Try Moving Search Products to Sub Nav
From the conversation with @josemarques we discussed exploring the option of trying to move the search bar on the products page ( and guessing other post listing pages ) into the sub-nav bar:

image

We were unsure if there was an available hook to do so, but figured I would still document it as an issue.

joshuatf:
This PR styles the search box to look like Calypso's and also moves it into the subnav if one exists on the page.

Fixes #151 

#### Screenshots
screen shot 2018-11-12 at 2 31 37 pm
screen shot 2018-11-12 at 2 31 44 pm
screen shot 2018-11-12 at 2 31 47 pm

#### Testing
1.  Visit pages with search boxes both with a subnav and without.  E.g., `/wp-admin/edit-tags.php?taxonomy=product_cat&post_type=product` and `/wp-admin/edit.php?post_type=product`
2.  Check that styling and functionality work for both types and match Calypso's - https://wordpress.com/store/products/

##### Some things to test for
* Search icon (magnifying glass) should open the input on products and focus the input.
* Close icon (x) should only be visible in subnav search bar and should close menu.
* Search should still work on both

#### Design
The Calypso style search runs on type, while the WP forms only run on submit.  I've kept the submit buttons inside the form for the time being, but should we change this?  /cc @josemarques 

We could also:
* Auto-submit when the user finishes typing
* Use a different icon for search?

### DIFF
--- a/assets/css/calypsoify.css
+++ b/assets/css/calypsoify.css
@@ -272,6 +272,15 @@ div.woocommerce-message, .wc-helper .start-container {
 }
 
 /* Navigation tabs */
+.subsubsub {
+	position: relative;
+	border: 0;
+	-webkit-box-shadow: 0 0 0 1px rgba(200,215,225,0.5), 0 1px 2px #e9eff3;
+    box-shadow: 0 0 0 1px rgba(200,215,225,0.5), 0 1px 2px #e9eff3;
+}
+.subsubsub a, .filter-links li>a {
+    padding: 11px 15px;
+}
 .woocommerce nav.woo-nav-tab-wrapper {
 	width: 100%;
 }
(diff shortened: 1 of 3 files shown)


josemarques:
> The Calypso style search runs on type, while the WP forms only run on submit. I've kept the submit buttons inside the form for the time being, but should we change this? /cc @josemarques

Is it possible to submit on enter and hide the button?

joshuatf:
@josemarques Sure!  Hid buttons and added a function to allow search on 'enter.'

@justinshreve Good idea on the auto-expand if a query is set.  Added that in.  Also disabled the autocomplete since it looks like this is turned off in Calypso.

PR #188, opened 2018-11-15
Hides all plugin activation and wp-admin admin notices in the calypso…

### LINKED ISSUE #65: Hide notices that are a part of the setup task list
See #64.

A number of the extensions we are bundling show notices right after installation. We can suppress the ones we are including on the setup task list.

jeffikus:
…ified wp-admin.

Closes #65


jeffikus:
Refactored version of #92

jeffikus:
Nice, I'm glad it's working well :)

PR #194, opened 2018-11-15
Fix Calypsoify on Woo Post Types

timmyc:
I _think_ this might have been a regression introduced in #183 - but with `master` of calypso-bridge installed, and Gutenberg 4.4 activated, visiting an edit/create page for a woo post type ( orders, coupons, products etc ) resulted in a partially calypsoified state:

image

__Steps to Reproduce the Bug__
- Install `master` of calypso-bridge
- Install/activate gutenberg 4.4
- View an order edit page and verify the screen above

__Steps to test this fix__
- Checkout the branch
- Visit a coupon/order/product page and verify all is calypsoified as expected


timmyc:
> but I'm wondering if this would be resolved by setting: define( 'JETPACK_DEV_DEBUG', true );

I experienced the issue locally with JEPACK_DEV_DEBUG true, and on an Atomic test site which is connected to wpcom ( and where we can't really lean on setting the DEBUG constant ) - so I still _think_ this is a valid issue on production that we need to fix.

PR #203, opened 2018-11-16
Add Jetpack dependencies early on to prevent breaking disconnected sites

### LINKED ISSUE #198: Disconnected Jetpack sites lose connected menu items
If a site is for some reason disconnected from Jetpack, all connected menu items get removed and other odd styling behavior occurs on the site.  

It seems like a good idea to check if Jetpack modules are loading early on and return if not as a safety measure.

screen shot 2018-11-16 at 8 45 23 am
styling

joshuatf:
Adds in early dependency checks to avoid fatal errors and removal of admin menu items if Jetpack is disconnected or somehow disabled.

Fixes #198 

#### Testing
1.  Check that plugin still loads correctly with a connected Jetpack site.
2.  Disabling Jetpack should not throw any fatal errors on WooCommerce routes (e.g. Orders, Products, etc).
3.  Enabling Jetpack should still not load this plugin unless `JETPACK_DEV_DEBUG` is set to true or Jetpack is connected.

### DIFF
--- a/class-wc-calypso-bridge.php
+++ b/class-wc-calypso-bridge.php
@@ -68,7 +68,7 @@ class WC_Calypso_Bridge {
 	 */
 	public function possibly_load_calypsoify() {
 		add_action( 'admin_init', array( $this, 'track_calypsoify_toggle' ) );
-		if ( 1 === (int) get_user_meta( get_current_user_id(), 'calypsoify', true ) ) {
+		if ( $this->dependencies_satisfied() ) {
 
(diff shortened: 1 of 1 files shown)


class-wc-calypso-bridge.php
```diff
@@ -83,6 +83,32 @@ public function possibly_load_calypsoify() {
 		}
 	}
 
+	/**
+	 * Check if dependencies are met to load Calypsoify
+	 *
+	 * @return bool
+	 */
+	public function dependencies_satisfied() {
```
justinshreve: Nice 👍

PR #219, opened 2018-11-19
Expand submenu in sidebar on tablets

joshuatf:
This should keep the sidebar submenu items from collapsing when viewing on mobile screen sizes.

#### Before
screen shot 2018-11-19 at 9 11 16 am

#### After
screen shot 2018-11-19 at 11 31 43 am

#### Testing
1.  Visit any dashboard page with a submenu (Orders, Products, etc.)
2.  Narrow viewport between 661px and 782px.
3.  Check that submenu items are shown and menu background/border is also now showing.

### DIFF
--- a/assets/css/calypsoify.css
+++ b/assets/css/calypsoify.css
@@ -1439,8 +1439,20 @@ table.widefat {
 		width: 190px;
 		position: relative !important;
 	}
+	#adminmenu .wp-has-current-submenu .wp-submenu {
+		display: block;
+		top: 0 !important;
+		left: 0;
+		width: 100%;
+	}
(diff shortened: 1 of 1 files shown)


joshuatf:
@justinshreve Could you give me the viewport size?  And was this after resize or after refresh?  I seem to recall something similar on resize; maybe we need a defer in here.

justinshreve:
I think it was around 730px or something -- however I just tested again and now it's working. I thought I tried clearing cache but maybe I was on the wrong branch or something by mistake? Either way -- looks like this is testing well for me now, so I'll get it merged.

PR #220, opened 2018-11-19
Append notices to first subnav only

joshuatf:
The WC Product Vendors subnav includes malformatted HTML with `<ul class="subsubsub"><ul class="subsubsub">...`.

This PR appends the notices to the first subnav only which is probably a safe practice anyway.

#### Screenshots
screen shot 2018-11-19 at 11 42 44 am

#### Testing
1.  Visit `/wp-admin/admin.php?page=wcpv-commissions`
2.  Make sure that notices are only appended underneath first subnav.

### DIFF
--- a/assets/js/calypsoify.js
+++ b/assets/js/calypsoify.js
@@ -98,7 +98,7 @@
     $( document ).ready(function() {
         var $subNavigation = $( '.wrap .subsubsub' );
         if ( $subNavigation.length ) {
-            $( 'div.notice, div.error, div.updated, div.warning' ).insertAfter( $subNavigation );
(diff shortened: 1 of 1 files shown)


timmyc:
Should we create a [new issue](https://github.com/woocommerce/woocommerce-product-vendors/issues) upstream for this bug too?

timmyc:
I'm not seeing the duplicate from WCS fwiw:

image

joshuatf:
@timmyc Good call.  Opened an issue and a PR here - https://github.com/woocommerce/woocommerce-product-vendors/pull/430

@justinshreve I'm having a hard time replicating the issue.  Could you try checking out this branch again and clearing cache?

justinshreve:
After toggle the Apple pay setting and clearing cache again, I only see one notice.

PR #228, opened 2018-11-19
Fix modal styles

### LINKED ISSUE #207: Improve shipping label discount modal display
screen shot 2018-11-16 at 1 38 31 pm

screen shot 2018-11-16 at 1 38 26 pm

The modal/overlay doesn't look great with the Calypsoify styles. Can we lighten it a bit? Do something else? @josemarques

justinshreve:
Fixes #207.

This PR fixes the modal styles by applying a light dim that matches Calypsos. It also hides the JITM notices on add/edit orders, because the WC JITM message is really already handled by the inline notice and NUX walkthrough. This notice is also visible on other pages. When present, the pointer gets pushed down and no longer points at the correct metabox.

Calypso Style:

screen shot 2018-11-19 at 12 09 50 pm

Before:

screen shot 2018-11-16 at 1 38 31 pm

screen shot 2018-11-16 at 1 38 26 pm

After:

screen shot 2018-11-19 at 12 11 00 pm

To Test:

* In `woocommerce-services/classes/class-wc-connect-nux.php`, force `is_new_labels_user()` to true, and update `$dismissed_pointers = $this->get_dismissed_pointers();` on line 90 to `$dismissed_pointers = array();`. This should force the NUX to show if previously dismissed.
* Go to edit order or add order page and test the NUX modal.

### DIFF
--- a/assets/css/calypsoify.css
+++ b/assets/css/calypsoify.css
@@ -1895,15 +1895,12 @@ p.search-box {
 /** WooCommerce Services / Modal */
 
 .wcs-pointer-page-dimmer {
-	z-index: 100200 !important; /* cover header too, like Calypso */
-	background-color: rgba(233,239,243,0.8) !important;
-}
 
-.wp-pointer {
-	z-index: 100201 !important;
+	background-color: rgba(233,239,243,0.8) !important;
 }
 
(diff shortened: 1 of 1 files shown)


josemarques:
Looks good to me.

justinshreve:
Good call. Done.

Issue #231, opened 2018-11-19
Calypsoify param set not working on setup walkthrough


justinshreve:
Per p1542653179364700-slack-wpcom-ecommerce-plan, a hard refresh is needed to see the skin.

You can reproduce by creating a store through Calypso.

* You need to be part of the tabbed plans A/B test. See p1542656406380400-slack-wpcom-ecommerce-plan.
* Activate the store sandbox (see field guide) and add a test card.
* Purchase an eCommerce plan. When you click "Create store" after purchasing, you are dropped in the setup flow. ?calypsoify=1 is passed as a URL -- but the theme is not actually set.

PR #236, opened 2018-11-20
Force cache invalidation on calypsoify param

joshuatf:
This PR forces cache invalidation and changes the redirect status when checking the calypsoify param.  This should force calypsoify to activate when visiting from the store setup.

Fixes (possibly) #231 

#### Testing
1.  Visit `wp-admin/admin.php?page=wc-setup&calypsoify=0` 2 times to "cache" the route.
2.  Visit `wp-admin/admin.php?page=wc-setup&calypsoify=1` and calypsoify should be active.

This needs to be tested through an ecommerce plan purchase on wp.com to verify that it's working.

#### Other possible solutions
When testing, it seems like the check param also was fired every time when a `sleep(2)` was added to the `check_calyposify_param()` function.

Unclear as to whether or not this managed to invalidate the cache somehow or if there are other issues with the `update_user_meta` not completing in time.


timmyc:
Going to close this one out since we merged in #239

PR #259, opened 2018-11-21
Automatically activate shipping based on geolocation

### LINKED ISSUE #253: No way of selecting shipping plugins to activate
We bundle a suite of shipping plugins such as WooCommerce Australia Post Shipping, WooCommerce Canada Post Shipping, WooCommerce Royal Mail, etc. See p58i-7iq-p2 for the full list.

These are all deactivated by default and should be activated depending on store country per the master thread.

We need to add the logic to have this happen automatically during the setup wizard.

joshuatf:
Enables shipping plugins for locations based on user's store location.
Fixes #253 

* Everywhere: UPS
* UK: Royal Mail
* US: USPS
* AU: Australia Post
* CA: Canada Post

#### Screenshots
screen shot 2018-11-21 at 10 31 16 am
screen shot 2018-11-21 at 10 31 25 am

#### Testing
1.  Go to `wp-admin/admin.php?page=wc-setup`.
2.  Select different countries from the dropdown for your store (test with above and one not in the list to verify UPS).
3.  Click "Continue" and check that plugins have been activated.

### DIFF
--- a/includes/class-wc-calypso-bridge-admin-setup-wizard.php
+++ b/includes/class-wc-calypso-bridge-admin-setup-wizard.php
@@ -105,12 +105,7 @@ class WC_Calypso_Bridge_Admin_Setup_Wizard extends WC_Admin_Setup_Wizard {
 		$this->steps = apply_filters( 'wc_calypso_bridge_setup_wizard_steps', $default_steps );
 		$this->step  = isset( $_GET['step'] ) ? sanitize_key( $_GET['step'] ) : current( array_keys( $this->steps ) ); // WPCS: CSRF ok, input var ok.
 
-		// @codingStandardsIgnoreStart
-		if ( ! empty( $_POST['save_step'] ) && isset( $this->steps[ $this->step ]['handler'] ) ) {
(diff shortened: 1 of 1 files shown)


includes/class-wc-calypso-bridge-admin-setup-wizard.php
```diff
@@ -339,6 +334,47 @@ public function get_next_step_link( $step = '' ) {
...
 	}
 
+	/**
+	 * Save step after continuing to next step
+	 */
+	public function save_step() {
+		// @codingStandardsIgnoreStart
+		if ( ! empty( $_POST['save_step'] ) && isset( $this->steps[ $this->step ]['handler'] ) ) {
+			WC_Calypso_Bridge::record_event( 'atomic_wc_obw_step_complete', array( 'name' => $this->step ) );
+			if ( method_exists( $this, 'pre_' . $this->steps[ $this->step ]['handler'][1] ) ) {
```
timmyc: Some more string duplication here and line 346. Just makes it a bit harder to read at first glance.

PR #261, opened 2018-11-21
Wait for link track completion and continue to href

### LINKED ISSUE #252: "I’m done" setting up button does not work
See https://trello.com/c/a0OGEH6U/74-im-done-setting-up-button-does-not-work

This was working previously, but there might have been a regression moving things around. The "I'm done" button should dismiss the setup page and redirect you to the orders page.

joshuatf:
Waits for the link track AJAX completion event before directing the window to the clicked link's href.

Fixes #252 

#### Screenshots
screen shot 2018-11-21 at 11 53 21 am
screen shot 2018-11-21 at 11 53 24 am

#### Testing
1.  Go to `/wp-admin/admin.php?page=wc-setup-checklist`.
2.  Click on both the task action buttons and the "I'm Done" button.
3.  Verify that clicks are still tracked and that each goes to the link after tracking.

Note: You can update the `atomic-ecommerce-setup-checklist-complete` option to `0` if you need to "uncomplete" the setup.

### DIFF
--- a/assets/js/calypsoify.js
+++ b/assets/js/calypsoify.js
@@ -15,20 +15,28 @@
      * Record checklist task click
      */
     $( '.checklist__task-title a, .checklist__task-secondary a' ).click( function() {
-        var $task = $( this ).closest( '.checklist__task' )
-        var status = $task.hasClass( 'is-completed' ) ? 'complete' : 'incomplete';
-        var taskId = $task.data('id');
-        var taskTitle = $task.data('title');
+        const $task = $( this ).closest( '.checklist__task' )
+        const status = $task.hasClass( 'is-completed' ) ? 'complete' : 'incomplete';
+        const taskId = $task.data('id');
+        const taskTitle = $task.data('title');
+        const href = $( this ).attr('href');
 
-        window.jpTracksAJAX.record_ajax_event(
-            'atomic_wc_tasklist_click',
-            'click',
-            {
-                id: taskId,
-                title: taskTitle, 
-                status: status,
-            }
-        );
+        if ( window.jpTracksAJAX ) {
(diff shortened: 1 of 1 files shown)


assets/js/calypsoify.js
```diff
@@ -15,33 +15,42 @@
...
      */
     $( '.setup-footer a' ).click( function(e) {
         e.preventDefault();
-        var progressNumber = $( '.checklist__header-progress-number' ).text().split( '/' );
-        var complete = progressNumber[0];
-        var total = progressNumber[1];
-        var percentage = parseFloat( complete / total ).toFixed( 2 ) * 100;
-        if (window.jpTracksAJAX) {
-            window.jpTracksAJAX.record_ajax_event(
+        const progressNumber = $( '.checklist__header-progress-number' ).text().split( '/' );
```
timmyc: There is a slight delay now while the request completes. Should we consider adding a `disabled` class to the element to have it verify the button has indeed been clicked and the "form" is submitting?

image

timmyc:
@josemarques can I get a quick design check from you on this? I added a `disabled` class when the actions are clicked to inform the users that the action is in progress while the track is being recorded:

disabled-done

Does this look okay to you? Otherwise the screen just kind of lingers while the callback from recording the track completes.

timmyc:
I'm going to merge this in as-is since it is broken right now. If the disabled addition is not good or we want to do something else then we can fix in a follow-up PR.

assets/js/calypsoify.js
josemarques: I think that's fine.

josemarques:
> Does this look okay to you? Otherwise the screen just kind of lingers while the callback from recording the track completes.

I think that's fine.

PR #263, opened 2018-11-21
Fig taxonomy form inconsistencies and tag successful message

### LINKED ISSUE #243: Tags screen inconsistencies
I noticed a few things on the tags screen:

tags
- The "action header" bar has different a height when it includes a button, which translates into a slight jump when entering/exiting the edit mode. 
- When we click the "Add New" button and the form is revealed the search box should be hidden along with the listing table.  
- When the user adds a new tag by pressing the "Add new tag" button the form should go away, the list should be revealed with the new tag, and a notification should inform the user that a new tag was added. 
- The search bar should have the same width as the table below

joshuatf:
Adds in extra styling to fix some alignment issues and watches the dom for the added category via AJAX to add a success message.

Fixes #243 

#### Screenshots
screen shot 2018-11-21 at 3 35 35 pm

#### Testing
1.  Visit `/wp-admin/edit-tags.php?taxonomy=product_tag&post_type=product`
2.  Check that styling issues noted in #243 are fixed.
3.  Add a new category.
4.  Check that form is toggled back to category list and that success message is added.

### DIFF
--- a/assets/js/calypsoify.js
+++ b/assets/js/calypsoify.js
@@ -126,12 +126,19 @@
     /**
      * Toggle taxonomy form
      */
-    $( '.taxonomy-form-toggle' ).click( function(e) {
-        e.preventDefault();
+    function toggleTaxonomyForm() {
         $( '#col-container > #col-left' ).toggle();
         $( '#col-container > #col-right' ).toggle();
         $( '.taxonomy-form-toggle' ).toggle();
         $( '.wrap .search-form' ).toggle();
+    }
+
+    /**
+     * Click handler for taxonomy add new/cancel buttons
+     */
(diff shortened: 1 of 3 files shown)


assets/js/calypsoify.js
```diff
@@ -295,5 +303,43 @@
         $( '.wp-list-table-wrapper__inner' ).scroll();
     } );
 
+    /**
+     * Detect changes to tag/category table
+     */
+    var addedTags = [];
+    $( window ).load( function() {
+        $( 'body' ).on( 'DOMSubtreeModified', '#the-list[data-wp-lists="list:tag"]', function() {
```
timmyc: 😮 magic.

PR #285, opened 2018-11-23
Add styling for select2 clear icon

### LINKED ISSUE #281: Clear page icon in Page setup needs more padding
I noticed the Page Setup in Advanced Settings displays a cross icon to clear the selected page:

screen shot 2018-11-22 at 18 17 59

This is how it looks in core:

screen shot 2018-11-22 at 18 20 32

Because our select boxes are narrower the icon is too close to the page title and almost seems like it's a bug (that's what I thought at first).

I suggest for this page we make the select boxes width consistent between them.

joshuatf:
Adds in styling for the clear icons in select2 dropdowns and increases the width of the settings dropdowns.

@josemarques I don't think there's precedent for clear icons in Calypso, but I used a similar close icon that seems to be commonly used.  Can you let me know if this looks okay or if we should revert?

Fixes #281 

#### Screenshots
screen shot 2018-11-23 at 12 27 00 pm

#### Testing
1.  Visit `/wp-admin/admin.php?page=wc-settings&tab=advanced`
2.  Check that styling for clear icons looks okay and is not overlapping text.
3.  Check that clear icon still works as expected.

### DIFF
--- a/assets/css/calypsoify.css
+++ b/assets/css/calypsoify.css
@@ -320,6 +320,15 @@ input[type=radio]:checked:before {
 .select2-container--default .select2-selection--single .select2-selection__placeholder {
 	color: #2e4453;
 }
+.select2-container--default .select2-selection--single .select2-selection__clear {
+	font-size: 0;
+}
+.select2-container--default .select2-selection--single .select2-selection__clear:before {
+	content: '\2715';
+	color: #c3d7e2;
+	font-size: 14px;
+	line-height: 24px;
+}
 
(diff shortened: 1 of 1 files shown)


josemarques:
@joshuatf you're right. I guess my initial reaction was to establish parity with Core and did not think about calypso. 

Would it be to late to just hide those? It's a weird pattern anyway.

justinshreve:
Per Slack:

>  for the sake of this release we can keep it as it is on that PR

PR #313, opened 2018-11-26
Change WC menu item to 'Store'

### LINKED ISSUE #296: Store and WooCommerce menu item incosistentcy
As @apeatling rightly pointed out, the 2 menu items on wp.com and wp-admin are different.

#### WP.com
screen-shot-2018-11-20-at-3-12-56-pm 1

#### wp-admin
screen-shot-2018-11-20-at-3-13-02-pm 1

We should try to change the wp-admin "WooCommerce" menu item to "Store."

joshuatf:
Changes the menu name to match that of WP.com to keep things consistent across the two.

Fixes #296 

#### Before
screen shot 2018-11-26 at 3 20 49 pm

#### After
screen shot 2018-11-26 at 3 20 38 pm

#### Testing
Check that the menu item name has changed and that the icon color matches the correct styles when active and not active.


sararosso:
Would the original request on the card - "Store, powered by WooCommerce" not fit? Just wondering where we can make sure WooCommerce is present but not the only thing they see. If it won't fit we'll need to look at bringing in the Woo branding in other places.

justinshreve:
I think that text is a bit long for a menu/nav item that is always present -- but I'm curious to hear @josemarques opinion.

We have this on the footer of all pages (which matches the branding we have for Store on WP.com/Calypso):

screen shot 2018-11-27 at 3 34 45 pm

Maybe this could work for v1 and we can see where else to pull in branding?

josemarques:
The sidebar is not a great place to add additional branding (except by using it in icons). 

I liked the idea of keeping the product name in the sidebar as the footer might not have enough visibility.

I think to keep consistency between Calypso and the WooCommerce navigation we could change the icon and Store title in Calypso to WooCommerce:

screen shot 2018-11-28 at 13 54 43

sararosso:
I don’t think WooCommerce in the Calypso sidebar is going to be recognizable enough. I think Store is fine in Calypso, but some additional Woo branding in the store “Home” is needed.

Open to ideas on this - not a blocker for launch but would like to get more Woo branding in the store itself.

sararosso:
Agreed @justinshreve thanks.

PR #330, opened 2018-11-27
Prevent WooCommerce deactivation

joshuatf:
This PR prevents WC deactivation by several means:
* Removes the deactivation links
* Prevents deactivating directly via URL
* Prevents the plugin from being removed from the active plugins array altogether

#### Screenshots
screen shot 2018-11-27 at 1 44 47 pm
screen shot 2018-11-27 at 1 49 20 pm

#### Testing
1.  Visit `/wp-admin/plugins.php` and note the lack of a "Deactivate" link next to WooCommerce.
2.  Visit the deactivation link directly `/plugins.php?action=deactivate&plugin=woocommerce%2Fwoocommerce.php&plugin_status=all&paged=1&s&_wpnonce=844b58b1fb` - the plugin should not be deactivated and you should recieve a notice stating it cannot be deactivated.
3.  If you really want to test the limits, try and run `deactivate_plugins( 'woocommerce/woocommerce.php' );` inside a plugin and note that the plugin is still active.

### DIFF
--- a/includes/class-wc-calypso-bridge-plugins.php
+++ b/includes/class-wc-calypso-bridge-plugins.php
@@ -36,6 +36,11 @@ class WC_Calypso_Bridge_Plugins {
 	 */
 	private function __construct() {
 		add_action( 'admin_init', array( $this, 'remove_mailchimp_redirect' ), 5 );
+		add_action( 'admin_init', array( $this, 'remove_mailchimp_redirect' ), 5 );
+		add_filter( 'plugin_action_links', array( $this, 'remove_woocommerce_deactivation_link' ), 10, 2 );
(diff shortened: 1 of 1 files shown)


includes/class-wc-calypso-bridge-plugins.php
```diff
@@ -36,6 +36,11 @@ public static function get_instance() {
 	 */
 	private function __construct() {
 		add_action( 'admin_init', array( $this, 'remove_mailchimp_redirect' ), 5 );
+		add_action( 'admin_init', array( $this, 'remove_mailchimp_redirect' ), 5 );
```
justinshreve: There is a duplicate `remove_mailchimp_redirect` call here.

PR #334, opened 2018-11-27
Add setup checklist return notice

### LINKED ISSUE #299: No notice to go back to setup checklist after action
When users click on an action in the checklist "Do it," they are sent to the action, but may not be able to find their way back as @MarinaWoo noted.

> After creating first product it was unclear how to get back to checklist. I clicked the X top right and went back to dashboard and then saw a notice # – be nice if that showed up on the product creation page, too.

It would be nice if we gave them an easy way to navigate back to the checklist.  Would a dismissible notice be appropriate here?  Something like this:
screen shot 2018-11-26 at 9 14 40 am

/cc @timmyc @josemarques

joshuatf:
Adds a notice to return to the setup checklist after navigating away. Fixes #299.

#### The notice is shown when
* User has clicked on a task from the checklist
* AND the task condition is met (complete)

#### The notice is hidden when
* The user clicks the dismiss "X" icon in the notice (persisted)
* The user returns to the setup checklist page
* The user 'uncompletes' a task

#### Screenshots
screen shot 2018-11-27 at 5 49 43 pm

#### Testing
1.  Visit the setup checklist `wp-admin/admin.php?page=wc-setup-checklist`
2.  Click on any task item.
3.  Complete that task's requirements if incomplete and you should see the notice to return to checklist.  If already completed, you should see the notice immediately.
4.  The notice will be hidden by either the dismiss button or navigating back to the setup checklist page.

#### Feedback needed
* Do we want to show these notices even after clicking on an already completed item (as it is currently)?
* Do we want to add in completion strings for each task item?  E.g., `You've finished setting up Stripe!` instead of `You've completed the task "Setup payments with Stripe!"`

### DIFF
--- a/assets/js/setup-checklist.js
+++ b/assets/js/setup-checklist.js
@@ -18,4 +18,115 @@
 		$( '.checklist__header-complete-label' ).text( i18nstrings.hide );
 	} );
 
+	/**
+     * Track 'I'm done' completion on task list
+     */
+    $( '.setup-footer a' ).click( function(e) {
+        e.preventDefault();
+        const progressNumber = $( '.checklist__header-progress-number' ).text().split( '/' );
+        const complete = progressNumber[0];
+        const total = progressNumber[1];
+        const percentage = parseFloat( complete / total ).toFixed( 2 ) * 100;
+        const href = $( this ).attr('href');
+        $( this ).addClass( 'disabled' );
+
+        if ( window.jpTracksAJAX ) {
+            const trackedEvent = window.jpTracksAJAX.record_ajax_event(
+                'atomic_wc_tasklist_finish',
+                'click',
+                { 
+                    complete: complete,
+                    total: total,
+                    percentage: percentage
+                }
+            );
+            trackedEvent.complete( function() {
+                window.location = href;
+            } );
+        } else {
+            window.location = href;
+        }
+	} );
+	
+	/**
+     * Checklist task click event
+     */
+    $( '.checklist__task-title a, .checklist__task-secondary a' ).click( function( e ) {
+        e.preventDefault();
+        const $task = $( this ).closest( '.checklist__task' )
+        const status = $task.hasClass( 'is-completed' ) ? 'complete' : 'incomplete';
+        const taskId = $task.data('id');
+        const taskTitle = $task.data('title');
+        const href = $( this ).attr('href');
+        $( this ).addClass( 'disabled' );
+
+        $.when(
+            trackTaskClick( taskId, taskTitle, status ),
+            setActiveTask( taskId )
+        ).done( function() {
+            window.location = href;
+        } );
+
+    } );
+
+    /**
+     * Track task clicks in Jetpack if enabled
+     *
+     * @param {string} taskId Task ID
+     * @param {string} taskTitle Task title
+     * @param {string} status Status - complete/incomplete
+     */
+    function trackTaskClick( taskId, taskTitle, status ) {
+        if ( window.jpTracksAJAX ) {
+            return window.jpTracksAJAX.record_ajax_event(
+                'atomic_wc_tasklist_click',
+                'click',
+                {
+                    id: taskId,
+                    title: taskTitle, 
+                    status: status,
+                }
+            );
+        } else {
+            return true;
+        }
+    }
+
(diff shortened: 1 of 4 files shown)


includes/class-wc-calypso-bridge-admin-setup-checklist.php
```diff
@@ -131,6 +135,14 @@ public function load_checklist_styles() {
 				'show' => esc_html__( 'Show completed', 'wc-calypso-bridge' ),
 			)
 		);
+		wp_localize_script(
+			'wc-calypso-bridge-setup-checklist',
+			'wccb',
+			array(
+				'ajaxUrl' => admin_url( 'admin-ajax.php' ),
+				'nonce' => wp_create_nonce( 'woocommerce_setup_active_task' ),
```
justinshreve: Minor indentation here, these lines should be aligned.

assets/js/setup-checklist.js
```diff
@@ -18,4 +18,115 @@
...
+	 *
+	 * @param {string} taskId Task ID
+	 */
+    function setActiveTask( taskId ) {
+        return $.ajax(
+            {
+                url: wccb.ajaxUrl,
+                data: (
+                    {
+						action: 'set_woocommerce_setup_active_task',
```
justinshreve: These three lines are not aligned correctly.

includes/class-wc-calypso-bridge-admin-setup-checklist.php
```diff
@@ -178,6 +190,61 @@ public function admin_menu_structure() {
...
+					$active_task = $task;
+				}
+			}
+			if ( $active_task && $active_task['condition'] ) {
+				?>
+				<div class="notice notice-success is-dismissible" id="woocommerce-setup-return-notice">
+					<p>
+						<?php
+						echo sprintf(
+							wp_kses( 'You\'ve completed the task "%s!"  Ready to continue with the setup checklist?  <a href="%s">Back to setup</a>.', 'wc-calypso-brigde' ),
```
justinshreve: Typo here. `brigde` -> `bridge`

joshuatf: Nice catch!

includes/class-wc-calypso-bridge-admin-setup-checklist.php
joshuatf: Looks like all the other lines were indented with spaces and these were the only ones using tabs.  Updated everything to tabs!

assets/js/setup-checklist.js
joshuatf: Same here!

josemarques:
> This behavior is a bit odd and I'm not really sure what to do about it. My original idea was to not add "You've completed the task X" and just write "Ready to continue with the setup checklist?" to keep things a bit more generic. @josemarques Do you have any thoughts on this one?

Since this adds a notice to the top of the page that call for attention I think it's a bit off to have a task suggesting to continue was nothing has been done yet.

justinshreve:
> Are you able to fix it with Justin's suggestion?

I didn't have a good solution for that issue 😆.  

My 2 cents is to go with a generic message like @joshuatf mentioned. At least for now, if we want this for next week's release. "Ready to continue with the setup checklist?" or ""Done reviewing? Go back to the setup checklist."

josemarques:
I'm a bit lost here. Will the message you are suggesting "Ready to continue with the setup checklist?" show up as soon as the user opens **any** screen? 

If that's the case I think we should look for exceptions (adding a product or checking out the stripe task) and only apply that pattern (displaying the generic message when entering the screen) only where we can't detect if the task is completed.

joshuatf:
The generic message would show up on any screen if the task is complete. We
can check for completion of any task, but for some tasks, the condition for
completion is just clicking on the task.

This means when they show up on a screen with a task that only requires a
click, they are instantly shown this message.

@josemarques does that make sense?  Should these be more generic or more
specific messages?

josemarques:
@joshuatf got it. Let's go with the generic message then.

joshuatf:
@justinshreve Reverted to the generic message.  Can you take a look and let me know if everything looks solid now?

PR #340, opened 2018-11-27
Only show Apple Pay notice on settings page

### LINKED ISSUE #321: Adjust Stripe Apple Pay Notice
See https://trello.com/c/enIthoPk/86-apple-pay-domain-failure-message-suppress-fix-keep

> "two notices are a bit scary, can they be hidden"
> Sara's note: is that error something we need to fix on our end? On the customer's end? suppress?

I'm wondering if this is something we need to show on all pages. Maybe we could limit it to settings / stripe settings? Also, is this something turned on by default? Maybe we should set an initial value during the OBW.

justinshreve:
Closes #321.

Per https://trello.com/c/enIthoPk/86-apple-pay-domain-failure-message-suppress-fix-keep, this might be handled automatically with the webhook solution Hydra has mentioned.

In the meantime, for the initial launch, I think it makes sense to limit this notice to settings pages. That way, it won't be in your face during the other screens, but can still be surfaced enough to take action on. It will also be present when they review shipping settings / Stripe.

This also follows the pattern with some other errors that only show in WC settings:

screen shot 2018-11-27 at 11 29 16 am

To Test:
* Add a stripe key/secret. These can be bogus. Verify you can get the notice show.
* Apply branch.
* View a settings page and see the notice.
* View another page, like add a product, and see the notice does not display.

### DIFF
--- a/includes/class-wc-calypso-bridge-hide-alerts.php
+++ b/includes/class-wc-calypso-bridge-hide-alerts.php
@@ -40,7 +40,7 @@ class WC_Calypso_Bridge_Hide_Alerts {
 		add_filter( 'woocommerce_helper_suppress_connect_notice', '__return_true' );
 		add_filter( 'woocommerce_show_admin_notice', '__return_false' );
 
-		add_action( 'admin_head', array( $this, 'hide_apple_pay_alert_on_non_settings_pages' ) );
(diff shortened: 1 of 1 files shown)


justinshreve:
I made an additional change here after the demo and talking to Timmy. We decided to move the Akismet notice to settings pages as well (and also any other none Store route).

PR #351, opened 2018-11-28
Fix setup checklist redirect when logged out

### LINKED ISSUE #322: wc-setup-checklist infinite redirect when logged out
The `wc-setup-checklist` page does an infinite redirect when the user is logged out. The setup page is linked from some onboarding emails, so we should make sure the login page accurately displays.

See p1543264772694800-slack-wpcom-ecommerce-plan.

joshuatf:
Run the check param on `admin_init` to avoid the `wp_safe_redirect` when activating Calypsoify.

Fixes #322 

#### Testing
1.  Visit `wp-admin/admin.php?page=wc-setup-checklist` and make sure things are still working as expected.
2.  Visit `wp-admin/admin.php?page=wc-setup-checklist` when logged out and check that login screen is shown.

### DIFF
--- a/class-wc-calypso-bridge.php
+++ b/class-wc-calypso-bridge.php
@@ -296,7 +296,10 @@ class WC_Calypso_Bridge {
 	 * Activates Calypsoify if the setup page is visited directly and it's not previously active.
 	 */
 	public function check_setup_param() {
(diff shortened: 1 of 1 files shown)


joshuatf:
@justinshreve Good catch on that.  Moved this back to init since the cap check in https://github.com/Automattic/wc-calypso-bridge/pull/351/commits/1d38ac08c00183cf8221b7b1789e191d894b82ee already resolves the issue.

PR #401, opened 2018-11-30
Hide installed extensions from search results on the "Extensions" screen

mattyza:
When searching for extensions via the "Extensions" screen, we currently show installed extensions as part of the search results. This can lead to confusion and potentially trying to purchase an extension twice.

While we'll see about getting this into WooCommerce core as well, this pull request replaces the "Extensions" screen with an extended class, containing additional logic to hide installed extensions from the search results.

### Testing Instructions
- Ensure you have at least one WooCommerce extension installed and activated on your test store (I've been using WooCommerce Brands as a test case).
- Visit "WooCommerce > Extensions" in WP Admin.
- Search for a keyword contained in the name of the extension above.
- The extension installed above should not show in the search results.

### Screenshots

image

cc @timmyc @sararosso


includes/class-wc-calypso-bridge-addons-screen.php
```diff
@@ -0,0 +1,79 @@
...
+		return $addons;
+	}
+
+	/**
+	 * Receives a plugin slug as plugin/plugin.php, and strips after the forward slash.
+	 * @param  string $plugin_slug full plugin slug
+	 * @return string              plugin directory name, with no attached file name
+	 */
+	public static function prepare_plugin_directory_slug( $plugin_slug ) {
+		$plugin_directory_name = explode( '/', $plugin_slug )[0];
```
justinshreve: I believe this is specific to PHP 5.4. Atomic v2 is running PHP7+ so it should be OK, but I wanted to note it if you are going to be doing something similar in core.

(https://docs.woocommerce.com/document/server-requirements/ makes it sound like the PHP version is 7+ as well, but I think we are still targeting 5.2+ like WP core?)

includes/class-wc-calypso-bridge-addons-screen.php
```diff
@@ -0,0 +1,79 @@
+<?php
+class WC_Calypso_Bridge_Addons_Screen extends WC_Admin_Addons {
```
justinshreve: There are some linter/coding style issues with both of these new files.

From the root directory, you can run `composer install` (see https://github.com/Automattic/wc-calypso-bridge#getting-started) and it should create pre-commit hooks for you to catch these. You can also lint the file manually:

`$ ./vendor/bin/phpcs includes/class-wc-calypso-bridge-addons.php`
`$ ./vendor/bin/phpcs includes/class-wc-calypso-bridge-addons-screen.php`

mattyza:
Thank you, @justinshreve !

PR #428, opened 2019-01-19
Remove store guided tour files

### LINKED ISSUE #426: Remove Store Guided Tour
I believe there are some bugs in the guided tour that was created for Store over a year ago - I think we should remove the code since the focus has shifted to supporting/promoting the e-commerce plans:

https://github.com/Automattic/wc-calypso-bridge/blob/master/store-on-wpcom/wc-calypso-bridge-class.php#L55

and

https://github.com/Automattic/wc-calypso-bridge/blob/master/store-on-wpcom/inc/class-customizer-guided-tour.php

/cc @Aljullu

Aljullu:
Fixes #426.

Removes store guided tour class and fixes PHPCS errors in the modified files.

It looks like GitHub doesn't recognize a file that was renamed, so it might be easier for you to review each commit for itself instead of reviewing the entire diff of this PR.

**Steps to test**

- Verify the plugin still loads without errors.
- Run tests and verify they still pass.
- Verify there are no regressions.


jeffstieler:
Behavior seems to be the same as `master`, though the styles seem to be conflicting:

screen shot 2019-01-21 at 9 56 32 am

If that's a non-issue, this LGTM.

jeffstieler:
> What are the steps to reproduce that? I'm unable to see this issue.

The PHP notice happens on `master` as well, so we can ignore that for this PR.

I enabled the plugin after setting my `at_options` and then appended `&calypsoify=1` to the url.

Aljullu:
Thanks @jeffstieler!

PR #432, opened 2019-02-08
Update load order for Store on WP.com REST API

justinshreve:
Re https://github.com/Automattic/wpcomsh/pull/230 @timmyc 

> when attempting to test out reviews - I was getting a rest_no_route response:

I swear that this was working for me before, but I can confirm that `calypso-reviews` (and `data/counts`) are not loading. These are loaded by `wc-calypso-bridge` in Store on WP.com mode.

I got this to work after updating the load order.

To test: Load up a Store on WP.com with this branch and test the reviews tab.
Note that I plan on rebuilding the .zip/release and regenerating the wpcomsh PR, so testing can happen there after that is well.

### DIFF
--- a/store-on-wpcom/class-wc-calypso-bridge.php
+++ b/store-on-wpcom/class-wc-calypso-bridge.php
@@ -39,17 +39,6 @@ class WC_Calypso_Bridge {
 	}
 
 	/**
-	 * Loads API includes and registers routes.
-	 */
-	private function init() {
-		if ( $this->is_woocommerce_valid() ) {
-			$this->includes();
-			// Ensure wc-api-dev has already registered routes.
-			add_action( 'rest_api_init', array( $this, 'register_routes' ), 20 );
-		}
-	}
-
-	/**
 	 * Makes sure WooCommerce is installed and up to date.
 	 */
 	public function is_woocommerce_valid() {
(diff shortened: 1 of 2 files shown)


store-on-wpcom/class-wc-calypso-bridge.php
```diff
@@ -32,7 +32,10 @@ class WC_Calypso_Bridge {
 	 * Constructor.
 	 */
 	public function __construct() {
-		add_action( 'woocommerce_init', array( $this, 'init' ), 20 );
```
timmyc: Can we remove `private function init() {` from this file with this change too?

justinshreve: 🤦‍♂️ Thanks for catching that. Removed.

timmyc:
So my test setup that I'm trying to validate this with is my wpsandbox site running the latest woo. I've disabled the wpcomsh loader, so that is not interfering. But I'm still not seeing the calypso-reviews endpoint output at https://timmydcrawford.wpsandbox.me/wp-json

Digging deeper it appears `class_exists( 'woocommerce' ) &&` is returning false in my setup when that code is evaluated. It seems like woo has yet to load. I changed the action back to `woocommerce_init` and it worked though :(

justinshreve:
Hmm - thanks for looking further. I'm not sure what the issues are here, but I'm going to set this back to `In Progress`, pause for now, and dig a bit deeper later into the load order. `wc-api-dev` can sit around a bit longer.

timmyc:
Yeah it really seems odd to me too. I tinkered around with it for half an hour this morning and couldn't get to a solution.

PR #433, opened 2019-02-14
Ecom Setup Checklist: Look at the proper Paypal settings.

matthusby:
This was checking for paypal api keys, but the setup item is for express checkout
that doesn't have api keys so it was not posible to complete the item.

Testing:
I am not sure how to test this, any pointers?

### DIFF
--- a/includes/class-wc-calypso-bridge-admin-setup-checklist.php
+++ b/includes/class-wc-calypso-bridge-admin-setup-checklist.php
@@ -434,9 +434,6 @@ class WC_Calypso_Bridge_Admin_Setup_Checklist {
 				'learn_more'      => 'https://woocommerce.com/products/woocommerce-gateway-paypal-checkout/',
 				'condition'       => isset( $click_settings['paypal'] ) &&
 								true === (bool) $click_settings['paypal'] &&
-								! empty( $paypal_settings['api_username'] ) &&
-								! empty( $paypal_settings['api_password'] ) &&
-								( ! empty( $paypal_settings['api_signature'] ) || ! empty( $paypal_settings['api_certificate'] ) ) &&
 								'yes' === $paypal_settings['enabled'],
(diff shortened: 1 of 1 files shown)


donpark:
Not all but most changes to wpcomsh can be tested using a wpsandbox.me site after commenting out host checking part of the wpcomsh loader. As for reviewing, I think more details are needed on what needs to be checked in WooCommerce to verify the change is working.

matthusby:
I have a site setup wpcomsh, but I have not been able to figure out to trick it into thinking it has an ecommerce plan so I can test this checklist.

And to expand on what this is fixing: In the checklist we are looking for the api credentials for paypal credit card processing but the action item is to setup paypal express (just the buy with paypal button).  So the checklist doesn't think you have completed the step even when you have.

justinshreve:
> I have a site setup wpcomsh, but I have not been able to figure out to trick it into thinking it has an ecommerce plan so I can test this checklist.

Checkout this section, specifically option values, and the Calypsoify param. You also need Jetpack connected or in debug mode. https://github.com/Automattic/wc-calypso-bridge#activating-calypsoify

> This was checking for paypal api keys, but the setup item is for express checkout
that doesn't have api keys so it was not posible to complete the item.

That's not true. The extension that gets installed is `WooCommerce PayPal Checkout Gateway` with the slug `woocommerce-gateway-paypal-express-checkout/woocommerce-gateway-paypal-express-checkout.php`. Clicking the settings link takes me to `/wp-admin/admin.php?page=wc-settings&tab=checkout&section=ppec_paypal` which has an `API Credentials` section.

matthusby:
@justinshreve Thanks for pointing me in the right direction!  I will have to look into this more, as when I tested it on my JP site it did fill the the api details, but on my ecommerce site it doesn't fill them in when I use the connect to pp button

matthusby:
After playing with this again I am going to abandon this as it is working how it should be.  Not sure where this disconnect happened, but its resolved :)

PR #436, opened 2019-03-14
Add localization

dzver:
This change creates a languages dir and adds some .mo files. As a result, the plugin will be translated into a number of languages.

To test:
- Switch your language settings to any of the languages
http://calypso-bridge.localhost/wp-admin/options-general.php

- Fake that you have the ecommerce plan, follow the option values part of https://github.com/Automattic/wc-calypso-bridge

- Make sure the page gets translated
http://calypso-bridge.localhost/wp-admin/admin.php?page=wc-setup-checklist


akirk:
It looks good to me but I don't have the environment to test this.

Issue #462, opened 2019-07-12
wc-admin support


timmyc:
Blocked by #461 

One goal of working on the ecommerce plan is to pave the way for bringing wc-admin to all e-comm plan customers. While the plugin currently isn't working as noted in the issue above, it seems there are a few things that need to be addressed once that blocker is out of the way:

- [x] Ensure Analytics Menu is displayed/accessible. Currently it is not part of the "supported" woo screens configured in wc-calypso-bridge
- [ ] What to do about the navigation bar and activity panels. Do we disable them on the ecomm plan as there is redundant breadcrumbs/screen titles. Perhaps we keep the activity panels still?
- [ ] Do we disable the legacy report menus?
- [ ] Identify issues with using the new OBW / setup tasklist from wc-admin

psealock:
>What to do about the navigation bar and activity panels. Do we disable them on the ecomm plan as there is redundant breadcrumbs/screen titles. Perhaps we keep the activity panels still?

For reference, here are both nav bars revealed with some inline CSS. Since the eComm breadcrumbs are incorrect, we should use wc-admin's. It would be a shame to remove functionality for those paying for the eCommerce plan.

![Screen Shot 2019-07-22 at 1 14 07 PM](https://user-images.githubusercontent.com/1922453/61600171-6a8f9a00-ac83-11e9-89bb-8c42698aefbc.png)

timmyc:
@jameskoster do you have any input / thoughts on how we should handle the wc-admin nav bar in the ecomm plan?

rrennick:
This script works to copy the breadcrumbs to Calypso for most instances

```
jQuery(document).ready(function($){
	$('#wpwrap').click(function(){
		wc_breadcrumbs = $('.woocommerce-layout__header-breadcrumbs').children();
		cb_crumb_wrap  = $( '.action-header__breadcrumbs' );

		if ( ! wc_breadcrumbs ) {
			return;
		}

		cb_crumb_wrap.children().remove();
		wc_breadcrumbs.clone().appendTo( cb_crumb_wrap );
	});
});
```

Opinions on pursuing a jQuery solution for the breadcrumbs and making the activity panel header visible?

rrennick:
> For reference, here are both nav bars revealed with some inline CSS.

@psealock Do you want to do a PR to make the activity panel visible but keep the breadcrumbs hidden?

psealock:
[...]

PR #477, opened 2019-07-22
disable WC Admin dev docs menu item by default

rrennick:
See #462 

This PR adds 

- a class for WooCommerce Admin customizations
- removes the `Dev Docs` menu item by default since people using Calypso are unlikely to be doing dev work in that environment
- Dev Docs can be restored with the `wc_calypso_bridge_development_mode` filter

### Testing

- Switch WC Admin to https://github.com/woocommerce/woocommerce-admin/tree/add/feature-config-filter
- Add the following filter
```
add_filter( 'wc_calypso_bridge_development_mode', '__return_false', 1001 );
```
- The Dev Docs menu item should be removed.


includes/class-wc-calypso-bridge-woocommerce-admin.php
```diff
@@ -0,0 +1,64 @@
...
+
+	/**
+	 * Remove the Dev Docs menu item unless allowed by the `wc_admin_get_feature_config` filter.
+	 */
+	public function remove_devdocs_menu_item() {
+		if ( ! function_exists( 'wc_admin_get_feature_config' ) ) {
+			/**
+			 * Override the default local configuration file generated by WC Admin.
+			 */
+			function wc_admin_get_feature_config() {
```
joshuatf: It seems like this will override our feature config.  Might be dangerous since certain items are not yet ready for the wild (like onboarding).

includes/class-wc-calypso-bridge-woocommerce-admin.php
```diff
@@ -0,0 +1,64 @@
...
+			self::$instance = new self();
+		}
+		return self::$instance;
+	}
+
+	/**
+	 * Initialize the class
+	 */
+	public function init() {
+		if ( ! apply_filters( 'wc_calypso_bridge_allow_wca_devdocs', false ) ) {
```
joshuatf: Optional: In https://github.com/Automattic/wc-calypso-bridge/pull/479 I added Jetpack-similar style pattern for development flag/filter using `wc_calypso_bridge_development_mode`.

We could optionally use the same filter here as well so we don't have to add a bunch of different flags to get a dev environment set up.

includes/class-wc-calypso-bridge-woocommerce-admin.php
rrennick: I think a better way to do this would be add a filter to WC Admin so that the filter can be used here to remove the feature.

rrennick: Added in https://github.com/woocommerce/woocommerce-admin/pull/2689

includes/class-wc-calypso-bridge-woocommerce-admin.php
rrennick: filter updated in https://github.com/Automattic/wc-calypso-bridge/pull/477/commits/eed62b5737e29a26e3992eca78655a3961258183

rrennick:
@joshuatf Thanks for the feedback. I've updated the PR description and applied the discussed changes.

timmyc:
> Pre-approving based on your discretion (or anyone else that wants to weigh in), but I'm wondering if we really need to filter this here. The devdocs menu item won't show up under wc-admin if not in dev mode, right? Are we worried that someone might be running in dev mode while working in wc-calypso-bridge and get confused?

Indeed I think it might be odd that someone runs in dev mode on an ecomm plan, but I suppose all things are possible.

I think having this filter here though will still allow for customizing/toggling features on/off as we seek to roll out wc-admin features to the ecomm plan. Theoretically we could just install from .org, then use this to control features.... though i'm guessing we will likely pull it in via wpcomsh, but this still allows for some control specific to the ecomm plan env.

PR #478, opened 2019-07-22
Setup: Ensure MailChimp does not redirect

timmyc:
This branch seeks to fix https://github.com/Automattic/wp-calypso/issues/32884

The current approach that was performed in that PR was to delete an option used by the MailChimp plugin to determine if they should redirect to their settings page upon activation. In multiple tests lately, the bug is still present.

Since the plugin itself doesn't offer a filter to determine if a redirect should happen, and apparently some code loading order shenanigans are happening with the current approach, I have opted to hook into the redirect logic itself. While this approach is working in my testing, I still might open a PR with MailChimp to add in a filter that we could use instead. But for now, here it goes!

__To Test__
To reproduce the redirect locally, all one must do is install and activate the MailChimp for WooCommerce plugin. It might be good to do this once before you test this branch. Then deactivate/delete the plugin.

1. Check out this branch
2. Install MC plugin again, activate
3. Note that you end up on the Setup dashboard instead of the mailchimp settings page.

For bonus points, you can mark your setup checklist as complete, and repeat steps above ( uninstall MC, reinstall ) and note that you are redirected to the MC settings page as this logic only loads when the setup checklist has yet to be completed.

### DIFF
--- a/includes/class-wc-calypso-bridge-setup.php
+++ b/includes/class-wc-calypso-bridge-setup.php
@@ -42,7 +42,7 @@ class WC_Calypso_Bridge_Setup {
 		add_filter( 'admin_body_class', array( $this, 'add_calypsoify_class' ) );
 
 		// If setup has yet to complete, make sure MailChimp doesn't redirect the flow.
-		$has_finshed_setup = get_option( 'atomic-ecommerce-setup-checklist-complete', false );
+		$has_finshed_setup = (bool) get_option( 'atomic-ecommerce-setup-checklist-complete' );
 		if ( ! $has_finshed_setup ) {
 			add_filter( 'wp_redirect', array( $this, 'prevent_mailchimp_redirect' ), 10, 2 );
 		}
--- a/includes/class-wc-calypso-bridge-setup.php
+++ b/includes/class-wc-calypso-bridge-setup.php
@@ -41,6 +41,12 @@ class WC_Calypso_Bridge_Setup {
 
 		add_filter( 'admin_body_class', array( $this, 'add_calypsoify_class' ) );
 
+		// If setup has yet to complete, make sure MailChimp doesn't redirect the flow.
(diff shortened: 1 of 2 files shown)


includes/class-wc-calypso-bridge-setup.php
```diff
@@ -69,6 +75,23 @@ public function remove_unused_steps( $default_steps ) {
...
+	 *
+	 * @param string $location Redirect location.
+	 * @param string $location Status code.
+	 * @return string
+	 */
+	public function prevent_mailchimp_redirect( $location, $status ) {
+		if ( 'admin.php?page=mailchimp-woocommerce' === $location ) {
+			// Delete the redirect option so we don't end up here anymore
+			delete_option( 'mailchimp_woocommerce_plugin_do_activation_redirect' );
+			$location = '/wp-admin/admin.php?page=wc-setup-checklist&calypsoify=1';
```
joshuatf: Should we use `get_admin_url()` to generate the URL here?

includes/class-wc-calypso-bridge-setup.php
```diff
@@ -35,6 +35,12 @@ public static function get_instance() {
 	 * Constructor.
 	 */
 	private function __construct() {
+		// If setup has yet to complete, make sure MailChimp doesn't redirect the flow.
+		$has_finshed_setup = (bool) get_option( 'atomic-ecommerce-setup-checklist-complete' );
```
joshuatf: Can we use `is_checklist_done()` from the class `WC_Calypso_Bridge_Admin_Setup_Checklist`?  I've introduced some new logic in https://github.com/Automattic/wc-calypso-bridge/pull/479 that would change what qualifies as complete.

We can probably change it to a static method.

timmyc:
Thanks for the review @joshuatf - implemented the updates, and rebased.

PR #479, opened 2019-07-23
Mark checklist complete if orders exist

### LINKED ISSUE #457: Setup Checklist on Upgraded Business Plan
On an existing business plan store, with an established Woo install ( products, orders, settings, customized etc ) I'm still being shown the setup checklist, and it appears to ignore the fact that I already have products setup:

image

#### Steps to reproduce
1. Establish a woo install on a business plan, setup products
2. Upgrade to the ecommerce plan
3. Note the setup list still shows steps like products needing to be setup

#### What I expected

Honestly I kind of feel like if a site already has at least a product and an order, that we shouldn't display the setup checklist.

joshuatf:
Fixes #457 

If a store has any orders, this PR will mark the setup as complete.

I agree with @timmyc's reasoning in #457 that orders probably mean a store has already been setup, but open to other input here.

Since this would impede development, I've also added a `wc_calypso_bridge_development_mode` filter that prevents the setup from ever being marked complete.

```php
add_filter( 'wc_calypso_bridge_development_mode', '__return_true' );
```

### Testing

1.  Enable calypsoify.
2. If you have no orders and the checklist has not been marked complete with the "I'm done" button, you should still see the checklist.
3. If you have any orders, the setup menu item should be gone and visiting the checklist page manually should redirect to the orders page. `wp-admin/admin.php?page=wc-setup-checklist`
4.  Adding the filter for development mode should once again allow you to access the setup screens. `add_filter( 'wc_calypso_bridge_development_mode', '__return_true' );`

### DIFF
--- a/README.md
+++ b/README.md
@@ -13,6 +13,10 @@ To check WPCS and lint errors via CLI, run the following from the root directory
 To automatically fix errors and beautify files, run the following from the root directory.
 `./vendor/bin/phpcbf [filename]`
 
+To turn on development mode for this plugin and prevent the setup checklist from being completed, the following filter can be added:
+
+`add_filter( 'wc_calypso_bridge_development_mode', '__return_true' );`
+
 ### Activating Calypsoify
 
 To Calypsoify the dashboard and test various functionality in this plugin, there are a number of conditions that must be met.
--- a/includes/class-wc-calypso-bridge-admin-setup-checklist.php
+++ b/includes/class-wc-calypso-bridge-admin-setup-checklist.php
@@ -86,12 +86,25 @@ class WC_Calypso_Bridge_Admin_Setup_Checklist {
 	}
 
 	/**
-	 * Returns if the checklist is "done" via the user clicking the "I'm done" button.
+	 * Returns if the checklist is "done" via the user clicking the "I'm done" button
+	 * or if the store already contains any orders.
 	 *
-	 * @return bool True if the done button has been clicked.
+	 * @return bool
 	 */
 	private function is_checklist_done() {
-		return (bool) get_option( 'atomic-ecommerce-setup-checklist-complete', false );
+		if ( apply_filters( 'wc_calypso_bridge_development_mode', false ) ) {
+			return false;
+		}
+
+		global $wpdb;
+		$order_count = (int) $wpdb->get_var(
+			"SELECT COUNT( DISTINCT posts.ID )
+				FROM {$wpdb->posts} as posts
+				WHERE posts.post_type = 'shop_order'
+				LIMIT 1"
+		);
+
+		return $order_count > 0 || (bool) get_option( 'atomic-ecommerce-setup-checklist-complete', false );
 	}
 
 	/**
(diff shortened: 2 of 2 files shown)


README.md
```diff
@@ -13,6 +13,10 @@ To check WPCS and lint errors via CLI, run the following from the root directory
 To automatically fix errors and beautify files, run the following from the root directory.
 `./vendor/bin/phpcbf [filename]`
 
+To turn on development mode for this plugin and prevent the setup checklist from being completed, the following filter can be added:
+
+`add_filter( 'wc_calypso_bridge_development_mode', '__return_true' );`
```
timmyc: Thanks for updating the docs with this 🌠

includes/class-wc-calypso-bridge-admin-setup-checklist.php
```diff
@@ -86,12 +86,25 @@ private function handle_redirects() {
...
-	 * Returns if the checklist is "done" via the user clicking the "I'm done" button.
+	 * Returns if the checklist is "done" via the user clicking the "I'm done" button
+	 * or if the store already contains any orders.
 	 *
-	 * @return bool True if the done button has been clicked.
+	 * @return bool
 	 */
 	private function is_checklist_done() {
-		return (bool) get_option( 'atomic-ecommerce-setup-checklist-complete', false );
+		if ( apply_filters( 'wc_calypso_bridge_development_mode', false ) ) {
```
timmyc: As you [noted in your review](https://github.com/Automattic/wc-calypso-bridge/pull/478#discussion_r306224823), should we change this method to public so we can utilize elsewhere?

includes/class-wc-calypso-bridge-admin-setup-checklist.php
```diff
@@ -86,12 +86,25 @@ private function handle_redirects() {
...
+			return false;
+		}
+
+		global $wpdb;
+		$order_count = (int) $wpdb->get_var(
```
timmyc: In lieu of a custom query, did you consider using `WC_Order_Query` or `wp_count_posts()`

includes/class-wc-calypso-bridge-admin-setup-checklist.php
joshuatf: Updated to a static method in https://github.com/Automattic/wc-calypso-bridge/pull/479/commits/01368a411b63751fde6ec8c6920e3f5188ece472

includes/class-wc-calypso-bridge-admin-setup-checklist.php
joshuatf: Because this is run so early, `WC_Order_Query` is not yet available and `wp_count_posts` does not seem to return the correct amounts (and because it's grouped by status, we'd have to loop through all default/custom statuses to check if any existed).

joshuatf:
Thanks for the review @timmyc!  Updated to a static method and left a note regarding the post count query.

timmyc:
Thanks Josh, i'll merge this in and rebase my PR

PR #481, opened 2019-07-23
use jQuery to copy WC Admin breadcrumbs to Calypso

rrennick:
See #462 

This PR uses jQuery to copy the WooCommerce Admin breadcrumbs to the Calypso breadcrumbs when the current page is a WooCommerce Admin page.

### DIFF
--- a/assets/js/calypsoify.js
+++ b/assets/js/calypsoify.js
@@ -377,5 +377,35 @@
 		}
 	}
 
+	/**
+	 * Copy WooCommerce Admin breadcrumbs to Calypso header.
+	 */
+	var wca_crumb_retry = 0;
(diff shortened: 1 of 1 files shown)


psealock:
This is working well for me. Nice job @rrennick. I left a [comment on parent issue](https://github.com/Automattic/wc-calypso-bridge/issues/462#issuecomment-514427068).

>I haven't had the chance to tinker around, but would it be possible to do something similar with Activity Panels? At that point, maybe it would be easier to absolutely position the entire wc-admin header and hide Calypso's?

Any thoughts on this approach just for wc-admin pages?

rrennick:
> Any thoughts on this approach just for wc-admin pages?

I saw that one first and responded there.

PR #489, opened 2019-07-24
move the activity panel header down and hide breadcrumbs

rrennick:
See #462 

This PR moves the WC Admin `.woocommerce-layout__header` down and hides the breadcrumbs (which are being copied to the Calypso activity bar).

### Testing

1. Activity header should be visible directly below the Calypso header
1. WC Admin breadcrumbs should not be visible
1. Activity panels should function

### DIFF
--- a/assets/css/calypsoify.css
+++ b/assets/css/calypsoify.css
@@ -2520,3 +2520,14 @@ h3#woocommerce_stripe_connection_status {
 .woocommerce-help-text {
 	text-align: center;
 }
+
+/**
+ * Styles for WooCommerce Admin
+ */
+.calypsoify-active .woocommerce-layout__header {
+	top: 93px;
+}
+
+.calypsoify-active .woocommerce-layout__header-breadcrumbs {
+	display: none;
+}


timmyc:
Screen grab:

image

There are still some rough edges to work out if we want to go with this approach. I'm seeing a left margin on the nav panel in the above screenshot. Just visually having the Action buttons above the activity panels just feels a bit off to me.

Also when viewed at different breakpoints, things go a bit sideways:

breakpoints2

Still kind of thinking the same thing as I [commented here](https://github.com/Automattic/wc-calypso-bridge/issues/462#issuecomment-514761005). Seems like we should spend a bit more time on this from a design standpoint.

rrennick:
After working on #491 I realized that the activity header is the only way to access notices on the WC Admin screens. Do we want to leave those hidden or make them accessible in some way?

timmyc:
I think if we disable the Activity Panel feature, the admin alerts will show as expected.

rrennick:
> I think if we disable the Activity Panel feature, the admin alerts will show as expected.

Disabling it doesn't show the notices. I'll look into having them display as normal.

PR #494, opened 2019-07-26
Add support for wc-admin layout

joshuatf:
Fixes (part of) #462 

Removes padding on layout elements not needed when Calypsoify is turned on.  Also makes wc-admin pages full width since the analytics need room to breathe.

### Before

Screen Shot 2019-07-26 at 3 32 08 PM
Screen Shot 2019-07-26 at 3 31 46 PM

### After
Screen Shot 2019-07-26 at 3 49 21 PM
Screen Shot 2019-07-26 at 3 52 10 PM

### Testing
1.  Enable calypsoify.
2.  Visit various pages and make sure layout looks the same as when wc-admin is deactivated.
3.  Visit wc-admin pages with calypsoify turned on.
4.  Make sure wc-admin pages match other calypsoify page padding and are full width.

### DIFF
--- a/assets/css/calypsoify.css
+++ b/assets/css/calypsoify.css
@@ -2554,3 +2554,20 @@ h3#woocommerce_stripe_connection_status {
 .woocommerce-help-text {
 	text-align: center;
 }
+
+/* WooCommerce Admin */
+.woocommerce-embed-page .woocommerce-layout__primary {
+	padding: 0;
+}
+.woocommerce-embed-page .wrap {
+	padding: 0;
+}
+.woocommerce-layout__primary {
+	margin: 0;
+}
+.woocommerce_page_wc-admin .wrap {
+	max-width: 100%;
+}
+.woocommerce-layout .woocommerce-layout__main {
+	padding-right: 0;
+}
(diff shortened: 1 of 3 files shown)


rrennick:
@joshuatf This worked great in my testing. I did find one related issue:

Without Calypsoify active, the `calypsoify-active` class is still being added producing some visual issues

Screen Shot 2019-07-30 at 2 09 06 PM

While that doesn't directly apply to this PR, possibly we can make the body class conditional in this PR.

psealock:
@rrennick in https://github.com/Automattic/wc-calypso-bridge/pull/494/commits/fd5c4f0a78a65532d2679d4cc3694aa5663dce52 I added a check to only add the `calypsoify-active` class if the page is a calypso bridge page.

### Test

1. Go to a Calypsoifyed page, see no change.
2. Go to `/wp-admin` (not a Calypsoifyed page) and run `document.querySelectorAll('.calypsoify-active')` and see the class has not been applied.

joshuatf:
We don't want to limit this class just to calypso bridge pages since there are other pages that could benefit from this class, but I put the `calypsoify-active` behind the dependency check so this should only load when calypsoify is active.

Can you give this another check? /cc @rrennick @psealock?

PR #499, opened 2019-09-01
1.1.0 version bump and notes

psealock:
Version bump to 1.1.0 and relevant changelog entries.

PR #501, opened 2019-09-27
Bail early if Woo is not active or present

nerrad:
See p9F6qB-477-p2 for original report.

The previous check just looked for if woocommerce.php existed but not if WooCommerce itself is active (included!).


nerrad:
Hmm I just realized that `wc-calypso-bridge` is a mu-plugin so the fix I added is definitely NOT correct (since woo won't be loaded yet for all sites).

psealock:
I pushed a version bump and release notes to this branch in preparation for a release. These are the equivalent changes made for the previous release in #499 and #500.

PR #510, opened 2019-12-16
Update Onboarding Colors

justinshreve:
Per the conversation in Slack ( p1576076602446600-slack-team-isotope ), this PR updates the purples used in Onboarding to use WordPress blue equivalents.

It also implements the pink buttons on other Calypsoify screens.

The CSS variables are hardcoded, because we don't currently have a build system, and there is no way to pull in the package recursively with composer (since wc-calypso-bridge is itself installed by wpcomsh via composer). I think this is fine, because variables should rarely change.

Screen Shot 2019-12-16 at 10 58 11 AM

Screen Shot 2019-12-16 at 10 59 03 AM

Screen Shot 2019-12-16 at 11 14 05 AM

To Test:
* Spot check various onboarding screens in both the profile wizard and task list
* Check some other Calypsoify pages for the pink buttons

### DIFF
--- a/assets/css/calypsoify.css
+++ b/assets/css/calypsoify.css
@@ -1,3 +1,7 @@
+:root {
+	--studio-wordpress-blue-60: #004e6e;
+}
+
 /**
 * Styles to override Jetpack Calypsoify styles for OBW and some WooCommerce screens
 */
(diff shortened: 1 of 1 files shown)


justinshreve:
@timmyc This does not support the color themes. Calypsoify itself does not handle the color schemes yet: https://github.com/Automattic/jetpack/issues/11478. The base work should be handled there, but further updates would most likely need to be done in this repo as well.

Issue #520, opened 2020-02-25
Two Setup Task Lists


timmyc:
With WooCommerce Beta 1 installed on a local copy of the ecomm plan, I'm seeing both the ecomm plan setup checklist, and the new onboarding checklist present:

<img width="1680" alt="image" src="https://user-images.githubusercontent.com/22080/75206273-e4f92e80-572a-11ea-923f-131cec48e495.png">

and

<img width="1680" alt="image" src="https://user-images.githubusercontent.com/22080/75206292-efb3c380-572a-11ea-82e4-c07a87f25e17.png">

@pmcpinto which of these two should be shown on the ecomm plan?

pmcpinto:
@timmyc I think that as for now we should simplify and keep the legacy task list. We can reevaluate if it makes sense to change it for the new cycle (depending on if we can remove calypsoify).

cc @jameskoster for any additional feedback.

timmyc:
Okay for 4.0 I'm going to disable the new wc-admin task list on the ecomm plan.

timmyc:
shipped in #521

PR #521, opened 2020-02-26
Ecomm: Never show the new setup checklist shipping in 4.0

timmyc:
For #520 

This branch seeks to disable the new setup checklist that is in wc-admin when the Calypsoify view is active. Since the calyspoified version of the ecommerce plan already has its own bespoke setup checklist, we have decided to disable the new one for now.

__Steps to Test__
First it is helpful to create the "issue" of both setup checklists being shown. To do this:

- Visit the orders Listing page `/wp-admin/edit.php?post_type=shop_order` And then go to Help | Setup Wizard and click enable on the setup checklist if shown:

enable-task-list(1)

- Visit `/wp-admin/admin.php?page=wc-admin&calypsoify=1` to enable calypsoify and view the wc-admin dashboard. Verify the screen looks like so:

image

- Next up, checkout this branch, and refresh the page, and verify the screen now looks like:

image

__FYI__
Since the file used to filter the options that force the new setup checklist to be hidden is only loaded when Calypsoify is active, if you "break out" of calypsoify by visiting `/wp-admin` and subsequently navigate to the wc-admin dashboard, the setup checklist will be shown there:

image

Which I feel is actually a *good* thing since the calypsoify checklist is not shown in the standard wp-admin view.

/cc @pmcpinto @jameskoster

### DIFF
--- a/includes/class-wc-calypso-bridge-woocommerce-admin.php
+++ b/includes/class-wc-calypso-bridge-woocommerce-admin.php
@@ -35,6 +35,8 @@ class WC_Calypso_Bridge_WooCommerce_Admin {
 	 */
 	public function init() {
 		add_filter( 'wc_admin_get_feature_config', array( $this, 'maybe_remove_devdocs_menu_item' ) );
+		add_filter( 'pre_option_woocommerce_task_list_hidden', array( $this, 'disable_new_task_list' ) );
+		add_filter( 'pre_option_woocommerce_onboarding_opt_in', array( $this, 'disable_onboarding_opt_in' ) );
 	}
 
 	/**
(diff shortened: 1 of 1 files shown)


jameskoster:
> Which I feel is actually a good thing since the calypsoify checklist is not shown in the standard wp-admin view.

Agreed.

timmyc:
>  but wondering if we should disable the onboarding in its entirety by forcing the woocommerce_onboarding_opt_in option to no

Hmmm I could have sworn I tried that approach and it didn't work, but I'll try it again as that would be a much easier solution.

timmyc:
@joshuatf the approach of only setting `woocommerce_onboarding_opt_in` does not work in my testing. Since this code is going to have a very short shelf-life with calypsoify on the outs, would it be okay if we ship as-is since it is functional?

joshuatf:
Sorry for tacking this on after, but we can probably also remove these lines now that we disable the task list:

https://github.com/Automattic/wc-calypso-bridge/blob/88cf879042ad89ef5b78b6e95f0551caade87cfe/includes/class-wc-calypso-bridge-admin-setup-checklist.php#L38-L47

timmyc:
Thanks @joshuatf - I added a follow up issue here https://github.com/Automattic/wc-calypso-bridge/issues/522

PR #536, opened 2020-04-29
Add tracks filters and host prop.

### LINKED ISSUE #527: Tracks: Add New Prop For E-Comm Plan
With https://github.com/woocommerce/woocommerce/pull/25851 set to land in 4.1, we can now implement these new filters here in wc-calypso-bridge to add props to each track to denote them as "sourced" from ecommerce plan.

Since we only want to track events specific to the ecommerce plan sites, I'm thinking creating a new class and including it, akin to most that live in https://github.com/Automattic/wc-calypso-bridge/tree/master/includes would be a good way to go here.

**Questions**
@pmcpinto we need to determine the name of the `prop` and the `value` specifically to use for the ecomm plan sites.

Should we use `host` for the prop name and `ecommplan` as the value?

**Prior Art**
I created the following [gist](https://gist.github.com/timmyc/edf820bded9a805a0746751a421bfb8a) with some examples of how to implement the filter in JS and PHP.

timmyc:
Fixes #527 

This branch implements usage of two new filters that are shipping in WooCommerce 4.1 that allow us to add in "global" event properties to tracks. This branch adds in a `host` prop for both PHP and JS triggered events with a value of `ecommplan'

__To Test__
- **Install the latest version of WooCommerce 4.1** ( RC2 at the time of PR creation )
- Checkout this branch.
- Ensure your test site has opted in to usage tracking `wp-admin/admin.php?page=wc-settings&tab=advanced&section=woocommerce_com`
- Visit the WooCommerce Admin dashboard page with the network tab of your browser console open. Filter the network requests for `t.gif` to see the track events. Verify the `host` prop is included with a value of `ecommplan`:

test-js-tracks

Testing the PHP tracks are a bit trickier - I just added some error_log'ing to the `add_tracks_php_filter` in this PR, and viewed the php error log to verify it was being called properly:

php-tracks

### DIFF
--- a/class-wc-calypso-bridge.php
+++ b/class-wc-calypso-bridge.php
@@ -140,6 +140,7 @@ class WC_Calypso_Bridge {
 			include_once dirname( __FILE__ ) . '/includes/class-wc-calypso-bridge-addons-screen.php';
 			include_once dirname( __FILE__ ) . '/includes/gutenberg.php';
 			require_once dirname( __FILE__ ) . '/includes/class-wc-calypso-bridge-woocommerce-admin.php';
+			require_once dirname( __FILE__ ) . '/includes/class-wc-calypso-bridge-tracks.php';
 
 			// Shared with store-on-wpcom.
 			include_once dirname( __FILE__ ) . '/store-on-wpcom/inc/wc-calypso-bridge-mailchimp-no-redirect.php';
--- a/includes/class-wc-calypso-bridge-tracks.php
+++ b/includes/class-wc-calypso-bridge-tracks.php
@@ -1,0 +1,71 @@
+<?php
+/**
+ * Tracks modifications for the ecommerce plan.
+ *
+ * @package WC_Calypso_Bridge/Classes
+ * @since   1.1.6
+ * @version 1.0.0
+ */
+
+defined( 'ABSPATH' ) || exit;
+
+/**
+ * WC Calypso Bridge Tracks
+ */
+class WC_Calypso_Bridge_Tracks {
+
+	/**
+	 * Class instance.
+	 *
+	 * @var WC_Calypso_Tracks instance
+	 */
+	protected static $instance = false;
+
+	/**
+	 * Get class instance
+	 */
+	public static function get_instance() {
+		if ( ! self::$instance ) {
+			self::$instance = new self();
+		}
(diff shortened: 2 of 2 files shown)


rrennick:
@timmyc This works fine inside the calypso interface but isn't added if the user is in the WP Admin interface. Do we want it limited to within Calypso?

timmyc:
@rrennick that is a really good question. My feeling is we might want to differentiate between those two use-cases. So track specifically the actions take in the "ecomm view" that has calypsoify, and maybe do a follow-up PR that adds in a different host tracking for people that are on the ecomm plan, but "breaking out" of the calypsoified view. Thoughts @pmcpinto ?

Either way I think it would be good to land this if it is testing okay for you.

pmcpinto:
@rrennick @timmyc good question. It would be nice to also have that segmentation, but in the meantime I think it will be possible to have a workaround for that through Tracks.

> and maybe do a follow-up PR that adds in a different host tracking for people that are on the ecomm plan, but "breaking out" of the calypsoified view.

What would be the best option to include this? A new value on the `host` prop?

rrennick:
> What would be the best option to include this? A new value on the host prop?

Is there a reason to not add another prop for this? It would only be in use for the remainder of the time that Calypso is active on ecomm plans.

timmyc:
If we want to track events on the ecomm plan, that fire outside of the calypsoified views, I think we will need to introduce a third code path in the plugin here for that. Currently we have the Store on WPCOM path - which effectively runs on any Atomic/biz plan site that has Woo installed, and calypsoify not active, and then the calypsoify path for ecom plans only.

So if we want to segment to ecom non-calypsoify only, seems like that would require a little bit more work.

rrennick:
> I think we will need to introduce a third code path in the plugin here for that.

We can move the new tracks class initialization to the main `wc-calypso-bridge` class then add logic to add the second prop if the calypso dependencies are satisfied.

pmcpinto:
> Is there a reason to not add another prop for this?

Yep an event prop it's probably a better option

> So if we want to segment to ecom non-calypsoify only, seems like that would require a little bit more work.

I can open a new issue for this additional work (not urgent). Sounds good?

timmyc:
Sounds good @pmcpinto

PR #549, opened 2020-05-21
add WC Admin feature filters

### LINKED ISSUE #532: WooCommerce Admin Installation on E-Comm Sites
With the vision to roll out the new navigation experience to e-Comm plan users, we will need to decide upon a mechanism to deliver the `woocommerce-admin` plugin to the install base. Currently the Atomic environment supports a few different options for inclusion of plugins like woocommerce-admin:

### wpcomsh
this is how `wc-calypso-bridge` itself is installed on ecom plan sites. `wpcomsh` is a `mu-plugin` and it and its dependencies that are pulled in via composer are always loaded, and not visible in the plugin listing inside wp-admin.

### managed, required
The next level of integration that can be seen for a plugin is one that is "managed" and users are not able to deactivate the plugin, Jetpack is an example of this:

Plugins ‹ Site Title — WordPress - (Private Browsing) 2020-04-22 15-16-56

### managed, not required
A step down from that level is a plugin that is managed/included with the subscription, but it **can be** deactivated by the user in wp-admin:

Plugins ‹ Site Title — WordPress - (Private Browsing) 2020-04-22 15-17-13

### bundled, but no visual indication
The last approach that can be seen on the ecommerce plan is how the extensions from the woo marketplace are included. They are installed, but do not have a banner denoting they provide a feature included with the plan. Users **can deactivate** and uninstall if they choose. @pmcpinto we might want to look into changing this since much of the value we deliver in the commerce plan is based upon the price of these extensions:

Plugins ‹ Site Title — WordPress - (Private Browsing) 2020-04-22 15-17-41

### Or should we consider packagist here?
One last approach that I think we should also consider for `woocommerce-admin` inclusion is exploring the route of packaging a version of it inside `wc-calypso-bridge` itself via composer. This would allow for us to pin to specific versions/builds of woocommerce-admin that have ecomm-plan specific features included, and would also prevent it from being disabled by customers.

rrennick:
Closes #532 

This is a companion PR to https://github.com/woocommerce/woocommerce-admin/pull/4418 which will allow enabling features hidden behind feature flags in the ecomm plan. This PR explicitly turns off the home page. Once this is merged, features will be able to be flipped using the two filters in the PR.

I added this as a new class to start working toward the eventual removal of Calypsoify. Code going in this class would be continue to be used on the ecomm plan.

### Testing

- Checkout https://github.com/woocommerce/woocommerce-admin/pull/4418 if not merged
- Enable WP_DEBUG (which normally turns on the home page)
- Store page should show the dashboard
- Edit this PR to make the same changes as outlined in 4418
- You should see the same results as outlined in that PR

### DIFF
--- a/composer.json
+++ b/composer.json
@@ -1,58 +1,53 @@
 {
-	"name": "automattic/wc-calypso-bridge",
-	"version": "v1.1.6",
-	"autoload": {
-		"files": [
-			"wc-calypso-bridge.php"
-		]
-	},
-	"repositories": [
-		{
-			"type": "package",
-			"package": {
-				"name": "automattic/gridicons",
-				"version": "3.1.1",
-				"source": {
-					"url": "https://github.com/Automattic/gridicons",
-					"type": "git",
-					"reference": "master"
-				}
-			}
-		}
-	],
-	"require": {
-		"composer/installers": "~1.2"
-	},
-	"require-dev": {
-		"phpunit/phpunit": "7.5.20",
-		"squizlabs/php_codesniffer": "*",
-		"wp-coding-standards/wpcs": "1.1.0",
-		"woocommerce/woocommerce-sniffs": "*",
-		"wimg/php-compatibility": "9.0.0",
-		"dealerdirect/phpcodesniffer-composer-installer": "0.4.4",
-		"bjornjohansen/wp-pre-commit-hook": "^0.2.2"
-	},
-	"config": {
-		"platform": {
-			"php": "7.1"
-		}
-	},
-	"scripts": {
-		"test": [
-			"phpunit"
-		],
-		"phpcs": [
-			"phpcs -s -p"
-		],
-		"phpcbf": [
-			"phpcbf -p"
-		]
-	},
-	"extra": {
-		"scripts-description": {
-			"test": "Run unit tests",
-			"phpcs": "Analyze code against the WordPress coding standards with PHP_CodeSniffer",
-			"phpcbf": "Fix coding standards warnings/errors automatically with PHP Code Beautifier"
-		}
-	}
+  "name": "automattic/wc-calypso-bridge",
+  "version": "v1.2.0",
+  "autoload": {
+    "files": [
+      "wc-calypso-bridge.php"
+    ]
+  },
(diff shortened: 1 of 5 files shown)


timmyc:
Looking nice Ron, I like this approach. Just confirming, this logic would run globally - i.e. would impact wp-admin and claypsoify views of WooCommerce correct?

Also looks like CI is unhappy here about some coding standards.

rrennick:
> Just confirming, this logic would run globally - i.e. would impact wp-admin and claypsoify views of WooCommerce correct?

Yes. I thought it would be good to start that with the separate file so future work that needs to work globally can go in that file.

rrennick:
> Also looks like CI is unhappy here about some coding standards.

@timmyc All fixed up.

rrennick:
@timmyc I removed the `woocommerce_admin_features` filter for consistency with https://github.com/woocommerce/woocommerce-admin/pull/4418

timmyc:
Thanks Ron!

composer.json
```diff
@@ -32,6 +32,11 @@
 		"dealerdirect/phpcodesniffer-composer-installer": "0.4.4",
 		"bjornjohansen/wp-pre-commit-hook": "^0.2.2"
 	},
+	"config": {
+		"platform": {
+			"php": "7.1"
+		}
+	},
```
jeffstieler, line 39: Are these composer file changes actually necessary?

includes/class-wc-ecomm-bridge.php
```diff
@@ -0,0 +1,65 @@
...
+
+	/**
+	 * Add hooks and filters if WooCommerce is active.
+	 */
+	public function initialize() {
+		if ( ! function_exists( 'WC' ) ) {
+			return;
+		}
+
+		add_filter( 'woocommerce_admin_features_to_enable_disable', array( $this, 'filter_wc_admin_enabled_features' ) );
```
jeffstieler, line 39: I think this needs to be update to `woocommerce_admin_features`, since https://github.com/woocommerce/woocommerce-admin/pull/4523 was the landed PR.

timmyc:
Closing in favor of #590

PR #556, opened 2020-06-17
Make host prop conditional in tracks.

### LINKED ISSUE #542: Add new value in the Host prop to check eComm plan usage outside of the calypsoified views
Add an event property to track when the eCommerce plan users are interacting with WooCommerce outside of the calypsoified views.

This is a follow up of the discussion that started here: #536

timmyc:
Fixes #542

Follow-up to #536

Currently, we only apply the `host` global tracks prop on calypsoified views on the ecommerce plan. This PR allows that `host` attribute to be added to *all* tracks on the ecommerce plan, **and** business plan woo sites regardless of wether the UI is calypsoified or not.

## Tracks Host Prop
The value of the `host` prop is set as following:

| Plan | `wp-admin` | `calypsoify` |
| ----- | ----------- | ------------- |
Business | `bizplan-wp-admin` | NA |
eComm | `ecommplan-wp-admin` | `ecommplan` | 

## To Test eComm Plan
These steps assume you already have the site [activated for calypsoify].

Just like in #536, testing out the PHP tracks events is made much easier by adding some `error_log` statements to the [add_tracks_filter()](https://github.com/Automattic/wc-calypso-bridge/compare/add/host-to-non-calypsoify?expand=1#diff-d7cab93a89457a8bba491f0285021c99L81) method - so something like:

```
public function add_tracks_php_filter( $properties, $event_name ) {
    $properties['host'] = self::$tracks_host_value;
    error_log( 'recording event: ' . $event_name );
    error_log( print_r( $properties, true ) );
    return $properties;
}
```

With that in place, then test the following:

- I find the coupons page is easy since a track is recorded in both JS and PHP, so open up coupons in a *non calypsoified* view: `/wp-admin/edit.php?post_type=shop_coupon`
-  Inspect the network tab on that page load, and filter by `t.gif` - view the request query string and verify the `host` prop is `ecommplan-wp-admin`:

wp-admin-js-track

- Also inspect your php error log, and if you used the example code above, you should see something like:

coupon-view-wp-admin

- Now adjust the url of the page to force the calypsoify ui: `/wp-admin/edit.php?post_type=shop_coupon&calypsoify=1`
- Inspect the same t.gif request params like above, and note the different `host` value:

image

- Again return to your php error log, and you should see the following:

coupon-view-calypsoify

## To Test Biz Plan
For this test case, you need to either undo the [activate calypsoify steps](https://github.com/Automattic/wc-calypso-bridge#activating-calypsoify) - or more simply, modify `wc_calypso_bridge_is_ecommerce_plan` [here](https://github.com/Automattic/wc-calypso-bridge/blob/master/wc-calypso-bridge.php#L43-L51) to `return false;` right away. This effectively mimics how the `wc-calypso-bridge` plugin operates on sites running WooCommerce on the WordPress.com business plan.

The repeat the steps above, loading the Coupons page, and looking at the error log and validating that the `host` property in those tracks events are using the `bizplan-wp-admin` value.

Worth noting that this change also implements the hiding of the tracks opt-out on biz plan sites, and opts them in my by default too.

### DIFF
--- a/includes/class-wc-calypso-bridge-tracks.php
+++ b/includes/class-wc-calypso-bridge-tracks.php
@@ -122,5 +122,3 @@ class WC_Calypso_Bridge_Tracks {
 		return $settings;
 	}
 }
-
-$wc_calypso_bridge_setup = WC_Calypso_Bridge_Tracks::get_instance();
--- a/includes/class-wc-calypso-bridge-tracks.php
+++ b/includes/class-wc-calypso-bridge-tracks.php
@@ -22,6 +22,14 @@ class WC_Calypso_Bridge_Tracks {
 	protected static $instance = false;
 
 	/**
+	 * Plugin host attribute for tracks.
+	 *
+	 * @var string
+	 */
+	public static $tracks_host_value = '';
+
+
+	/**
 	 * Get class instance
 	 */
 	public static function get_instance() {
@@ -35,6 +43,7 @@ class WC_Calypso_Bridge_Tracks {
 	 * Constructor.
 	 */
 	private function __construct() {
+		$this->set_tracks_host_value();
 		add_filter( 'admin_footer', array( $this, 'add_tracks_js_filter' ) );
 		add_filter( 'woocommerce_tracks_event_properties', array( $this, 'add_tracks_php_filter' ), 10, 2 );
 		add_filter( 'woocommerce_get_sections_advanced', array( $this, 'hide_woocommerce_com_settings' ), 10, 1 );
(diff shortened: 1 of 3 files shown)


timmyc:
Going to move this off needs review and add support for adding a `host` prop to biz plan users who have woo installed also.

timmyc:
Okay this is ready for a review. Testing is a little bit of a bear on this one, with having to hack some files locally - sorry, but for me that is the easiest way to verify this under all three environments this plugin provides.

@pmcpinto just wanting to make sure you are good with the track host names i'm proposing here.

pmcpinto:
> @pmcpinto just wanting to make sure you are good with the track host names i'm proposing here.

Yep, all good!

includes/class-wc-calypso-bridge-tracks.php
```diff
@@ -45,6 +54,26 @@ private function __construct() {
...
+	/**
+	 * Set's the value for the tracks host property.
+	 */
+	public function set_tracks_host_value() {
+		// Default value assumes business plan, inside wp-admin.
+		$host_value = 'bizplan-wp-admin';
+
+		// If an ecomm plan site, update host value.
+		if ( wc_calypso_bridge_is_ecommerce_plan() ) {
+			if ( 1 === (int) get_user_meta( get_current_user_id(), 'calypsoify', true ) ) {
```
mattsherman, line 66: When I run this branch, `get_current_user_id()` here returns `0`, when the current user is `1`, which causes the wrong user meta to be looked up. I think this is because this code is executing before the current user is set.

I think the solution is to wait to call `get_instance` until the `init` action fires. So, instead of...

```
$wc_calypso_bridge_setup = WC_Calypso_Bridge_Tracks::get_instance();
```

do...

```
add_action( 'init', array( 'WC_Calypso_Bridge_Tracks', 'get_instance' ) );
```

I tested this, and it seems to do the trick.

timmyc: Thanks @mattsherman - sorry this wasn't working... and wondering why it was working for me locally. I applied a change per your suggestion and it is testing out well for me, thanks!

mattsherman:
Will re-review/test shortly

PR #561, opened 2020-06-24
Remove Legacy OBW

timmyc:
While I'm still not entirely certain _how_, the old OBW that was built for the ecommerce plan was still launching for new stores. This PR removes all the code around the legacy OBW + Setup Checklist, and instead favors the new default experience that ships in WooCommerce core.

The second commit in this PR adds in some `function_exists` checks to allow for the plugin to be installed "normally" alongside the bundled version in wpcomsh on a biz plan site, which allows for testing of this branch.

__To Test__
This one is a bit involved, but really the only way to truly test this out as I couldn't recreate the buggy state locally myself.

0. Give yourself at least 600 wpcom credits on an account you want to use to...
1. Add a new ecommerce plan test site. This can be done by using the "switch site" tool in the calypso sidebar, and selecting Add new site at the bottom.
2. Go through the checkout flow, use a .blog domain
3. Note in the final step when you click the setup button to launch your store, you are redirected to the new atomic site, and the old OBW is shown ( you can see this all in @pmcpinto's [video here](https://www.loom.com/share/f2ecfdbf95a14f999e101f046b8cc2ac)
4. manually change the URL to `/wp-admin` on your new site
5. Zip up this branch of wc-calypso-bridge and install + activate it on the site. Make sure the unzipped folder for the plugin is named `wc-calypso-bridge`.
6. navigate to `/wp-admin/admin.php?page=wc-setup&calypsoify=1`, note the woo core onboarding flow is shown.


timmyc:
Doh I have some updates I need to do here.

timmyc:
This is indeed ready for 👀

mattsherman:
I'll review this.

mattsherman:
>     6. Zip up this branch of wc-calypso-bridge and install + activate it on the site.

When I do this, I get the following error when activating...

```
Notice: Constant WC_CALYPSO_BRIDGE_CURRENT_VERSION already defined in /srv/htdocs/wp-content/plugins/wc-calypso-bridge-remove-legacy-obw/wc-calypso-bridge.php on line 33
```

What am I doing wrong?

timmyc:
@mattsherman interesting, so you are getting that on an atomic site? I hit a similar bug which led to 42ae981 - but I didn't get that error. It almost seems like [this logic](https://github.com/Automattic/wc-calypso-bridge/blob/master/wc-calypso-bridge.php#L25-L31) is no longer operating properly and the version of `wc-calypso-bridge` that is bundled in `mu-plugins` via wpcomsh is still loading.

mattsherman:
> @mattsherman interesting, so you are getting that on an atomic site? I hit a similar bug which led to [42ae981](https://github.com/Automattic/wc-calypso-bridge/commit/42ae9816db67bb6b2773be869e6d63c2cae9b55c) - but I didn't get that error. It almost seems like [this logic](https://github.com/Automattic/wc-calypso-bridge/blob/master/wc-calypso-bridge.php#L25-L31) is no longer operating properly and the version of `wc-calypso-bridge` that is bundled in `mu-plugins` via wpcomsh is still loading.

Ah, bingo... that helped me figure it out... I was uploading the plugin with a folder name other than `wc-calypso-bridge`. Once I renamed it to that, re-zipped, and uploaded, it activated properly! Now on to verifying the last step...

mattsherman:
>     navigate to `/wp-admin?page=wc-setup&calypsoify=1`, note the woo core onboarding flow is shown.

I assume this is supposed to be `/wp-admin/admin.php?page=wc-setup&calypsoify=1`? When I use that, the Woo Core onboading flow is properly shown!

PR #563, opened 2020-06-24
Ignore *.swp files

mattsherman:

The Vim editor creates `.swp` files by default when editing files. These files are temporary and shouldn't be mistakenly added to git. This PR updates `.gitignore` so that git will ignore these files.

1. Run branch
2. Add a `foo.swp` file somewhere in your `wc-calypso-bridge` directory tree
3. Run `git status` and verify that the `foo.swp` file isn't listed

### DIFF
--- a/.gitignore
+++ b/.gitignore
@@ -6,6 +6,7 @@ project.properties
 .project
 .settings*
 .idea
+*.swp
 
 # OS X metadata
 .DS_Store


mattsherman:
> I'm jealous of your vim usage. But alas, I will return to my vscode, tail between my legs, ignoring my vim usage of days past. 🤣

You can always return... vim will never die.

PR #570, opened 2020-07-02
Correct the menu path for analytics so that it shows in the main menu.

### LINKED ISSUE #569: [WC 4.3 RC.2] Analytics Menu Missing
While testing out the 1.1.9 release today on a wpsandbox.me site running WC 4.3 RC2 - @psealock noticed that the Analytics menu is not being shown in the Calypsoify view. We confirmed the bug does not exist in 4.2.2 so this has somehow regressed in 4.3 - possibly related to the new homescreen work maybe?

We need to investigate the cause this week, and ship a fix ahead of the core release on July 7.

samueljseay:
Fixes #569 

### Testing

1. Ensure you have RC4.3 of WooCommerce plugin installed/activated
2. Ensure you have the wc-calypso bridge activated
3. Confirm that the Analytics menu appears and can be navigated to in the side bar
4. opt out of the new home screen via `WooCommerce > Settings > Advanced > features`
5. Confirm that the Analytics menu appears and can be navigated to in the sidebar once more.

### DIFF
--- a/includes/connect/woocommerce-admin.php
+++ b/includes/connect/woocommerce-admin.php
@@ -18,8 +18,8 @@ wc_calypso_bridge_connect_page(
 wc_calypso_bridge_connect_page(
 	array(
 		'screen_id' => 'woocommerce-revenue',
-		'menu'      => 'wc-admin&path=/analytics/revenue',
-		'submenu'   => 'wc-admin&path=/analytics/revenue',
+		'menu'      => 'wc-admin&path=/analytics/overview',
+		'submenu'   => 'wc-admin&path=/analytics/overview',
 	)
 );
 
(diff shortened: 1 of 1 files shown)


includes/connect/woocommerce-admin.php
```diff
@@ -18,8 +18,8 @@
 wc_calypso_bridge_connect_page(
 	array(
 		'screen_id' => 'woocommerce-revenue',
-		'menu'      => 'wc-admin&path=/analytics/revenue',
-		'submenu'   => 'wc-admin&path=/analytics/revenue',
+		'menu'      => 'wc-admin&path=/analytics/overview',
```
timmyc, line 21: @samueljseay I think we might need to add in an entire new call to `wc_calypso_bridge_connect_page()` with this config, and keep the legacy one around too to cover both the use cases of home screen active and not active. I just tested this branch out locally and confirmed that the analytics menu is now shown 🎉

When I opted out of the new home screen via WooCommerce > Settings > Advanced > features - The analytics menu item is no longer shown.

samueljseay: @timmyc I've implemented the suggested change and tested it, it looks like it works. Would you be ok to take another look?

PR #579, opened 2020-07-15
[4.3] Fix Marketing Menu Display

### LINKED ISSUE #577: Marketing tab disappeared from the Calypso menu
**Describe the bug**
I was testing the new home screen in the eCommerce plan and meanwhile, I noticed that the Marketing menu item disappeared from the Calypso menu

**To Reproduce**
1. Login with an account with the eCommerce plan active
2. Click on Store
3. See that the Marketing item is not present in the menu

**Expected behavior**
See the Marketing menu item in the Store menu

**Screenshots**
https://d.pr/i/8is0OQ

timmyc:
Fixes #577 

In https://github.com/woocommerce/woocommerce-admin/pull/4526 the logic that registers the menu item for the Marketing tab was changed slightly, which resulted in the Marketing tab not appearing in WooCommerce 4.3 - this branch fixes that.

__To Test__
- Install WooCommerce 4.3
- Activate Calypsoify, definitely recommend using p90Yrv-1O5-p2
- Visit `wp-admin/admin.php?page=wc-admin&calypsoify=1` and note that the Marketing menu is not shown
- Checkout this branch
- Refresh the page and see that the Marketing tab is now there again

marketing-is-back

### DIFF
--- a/includes/connect/woocommerce-admin.php
+++ b/includes/connect/woocommerce-admin.php
@@ -34,7 +34,7 @@ wc_calypso_bridge_connect_page(
 wc_calypso_bridge_connect_page(
 	array(
 		'screen_id' => 'woocommerce-marketing',
-		'menu'      => 'wc-admin&path=/marketing',
+		'menu'      => 'woocommerce-marketing',
 		'submenu'   => 'wc-admin&path=/marketing',
 	)
 );


timmyc:
Aside - I went to write some tests around [`woocommerce_menu_handler()`](https://github.com/Automattic/wc-calypso-bridge/blob/master/includes/class-wc-calypso-bridge-menus.php#L124) to prevent regressions like this in the future, but that method interacts with the `$menu` global, which is not present in the standard runtime of the phpunit test suite. 

I started to look for solutions to "force" the admin environment for the tests, but things started to get a little funky, and at that point it started to feel like a bit bigger than a simple unit test :\

PR #599, opened 2020-09-09
Disable Crowdsignal Redirect

timmyc:
Currently the [Crowdsignal Forms](https://wordpress.org/plugins/crowdsignal-forms/) plugin adds in a redirect that results in a newly created ecommerce plan site in being redirected to their setup flow instead of Woo. Discussion and details about this can be seen in 716-gh-greenhouse

The proposed solution here is to hook into the `add_option` for the `crowdsignal_forms_do_activation_redirect` value which is used to determine if this redirect should happen. When the option is added, this branch simply deletes the option right away to prevent the redirect from happening.

__Before__
Getting Started ‹ ecomm woo test — WordPress 2020-09-09 11-15-58

__After__
Plugins ‹ ecomm woo test — WordPress 2020-09-09 11-16-54

__To Test__

- Using `master` of `wc-calypso-bridge` add the Crowdsignal Forms plugin to your site, and activate.
- Note you are redirected to their setup page ( before screenshot above ).
- Deactivate the crowdsignal plugin.
- Apply this branch, and activate the Crowdsignal Forms plugin.
- Note that you are no longer redirected, but instead see the plugins page ( after screenshot above )

### DIFF
--- a/includes/class-wc-calypso-bridge-crowdsignal-redirect.php
+++ b/includes/class-wc-calypso-bridge-crowdsignal-redirect.php
@@ -8,7 +8,7 @@
 defined( 'ABSPATH' ) || exit;
 
 /**
- * WC Calypso Bridge Hide Alerts
+ * WC Calypso Bridge Crowdsignal Redirect
  */
 class WC_Calypso_Bridge_Crowdsignal_Redirect {
 
--- a/includes/class-wc-calypso-bridge-crowdsignal-redirect.php
+++ b/includes/class-wc-calypso-bridge-crowdsignal-redirect.php
@@ -44,10 +44,10 @@ class WC_Calypso_Bridge_Crowdsignal_Redirect {
 	}
 
 	/**
-	 * When the option to redirect is added, update to false.
+	 * When the option to redirect is added, delete the option.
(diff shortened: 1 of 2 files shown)


includes/class-wc-calypso-bridge-crowdsignal-redirect.php
```diff
@@ -0,0 +1,55 @@
...
+/**
+ * Prevents Crowdsignal Forms plugin from doing a redirect.
+ *
+ * @package WC_Calypso_Bridge/Classes
+ */
+
+defined( 'ABSPATH' ) || exit;
+
+/**
+ * WC Calypso Bridge Hide Alerts
```
jeffstieler, line 11: ©️ 🍝

jeffstieler (approved):
Code is simple and tests well.

Note - I tried to get this to work without mucking with options at all, by way of removing [the filter used](https://github.com/Automattic/crowdsignal-forms/blob/6ca0dcba2cbb552e5503e5be475f972d1a9e7387/includes/class-crowdsignal-forms.php#L139) by CrowdSignal to redirect, and failed. Heh.

PR #614, opened 2020-10-27
Remove calypsoify styles and UI elements

joshuatf:
In order to use the new WooCommerce navigation experience, calypsoify needs to be removed.  This PR removes any calypsoify-specific styles and UI elements from the bridge.

**Questions:**
1. Do we need a way to "break out" of the ecommerce plan experience?  Or should changes to alerts, theme setup, etc always remain in effect for ecomm plans?

**Follow-ups:**
1. I think there is a great opportunity to remove the connected pages here and instead rely on the `Screen` class defined in the navigation to check if we're on a WC page.
1. `Support` and `Manage Site` (back button) will both be missing from the navigation and will need to be added in.

### Testing

1. Make sure you follow the readme instructions to set up the ecomm plan view.
1. Make sure no regressions have occurred with non-calypsoify elements.
1. Walk through the new OBW to make sure bridge changes related to the ecomm plan still exist there.


timmyc:
@joshuatf 🥳 

So we need to probably hold off on merging this until 4.8 lands correct? Based off this, should we have a feature branch for nav to roll up all the changes needed for ease of testing everything in one branch?

joshuatf:
>So we need to probably hold off on merging this until 4.8 lands correct?

I think so.  There's also some nav organizational changes that may affect this as well.

>Based off this, should we have a feature branch for nav to roll up all the changes needed for ease of testing everything in one branch?

https://github.com/Automattic/wc-calypso-bridge/pull/615 contains the changes here and turns on the nav by default.  Will have one more commit to handle the necessary addition of support links and back.

psealock:
I think the feature branch is a good idea. I spotted a follow up required and I imagine there will be more. It will be nice to isolate those into separate PRs.

Jetpack's addition of the "Back" button for return to Calypso hides Navigation's "back" button.

Screen Shot 2020-10-29 at 3 55 44 PM

And when using @joelclimbsthings's changes from https://github.com/woocommerce/woocommerce-admin/pull/5497, collapsing the Navigation will require some handling of that button too.

Screen Shot 2020-10-29 at 3 54 31 PM

README.md
```diff
@@ -32,40 +32,14 @@ You will need either a connected Jetpack site or use the following filter to for
 
 `add_filter( 'jetpack_development_mode', '__return_true' );`
```
psealock, line 33: Latest version of Jetpack throws tons of deprecation warnings here, saying to use `jetpack_offline_mode` instead.

assets/js/calypsoify.js
```diff
@@ -1,443 +0,0 @@
...
-	 * Append subnav dropdown for mobile.
-	 */
-	$( document ).on( 'click', '.nav-tab-toggle', function() {
-		$( this ).parent().toggleClass( 'is-open' );
-	} );
-
-	/**
-	 * Toggle taxonomy form.
-	 */
-	function toggleTaxonomyForm() {
```
psealock, line 123: I'm not too familiar with this, but could it be function we'd like to keep?

psealock:
>So we need to probably hold off on merging this until 4.8 lands correct?

Whats happening in WC 4.8 that affects this?

assets/js/calypsoify.js
joshuatf: This is purely a way to emulate switching pages for taxonomy creation to mimic Calypso's behavior.  I don't think we'll need it going forward.

README.md
joshuatf: Updated readme in https://github.com/Automattic/wc-calypso-bridge/pull/614/commits/2011d46aa9d8e07e61999f3c078efbe0cece6df0

joshuatf:
>Jetpack's addition of the "Back" button for return to Calypso hides Navigation's "back" button.

Good catch.  I think since we're generally moving towards removing calypso elements, we should also remove this.  Added a  method to decalypsoify in https://github.com/Automattic/wc-calypso-bridge/pull/614/commits/371094bc6b49de6b4f1f9f3fcea1c1a95dd55019.  Note that it's possible to still see calypso styling from Jetpack by appending `&calypsoify=1` to the URL, but if it's already set in a user's meta profile, it will no longer show.

psealock:
>Note that it's possible to still see calypso styling from Jetpack by appending &calypsoify=1 to the URL, but if it's already set in a user's meta profile, it will no longer show.

I'm not entirely sure how to test this. Separately, should we mimic the behaviour of the link by making the url this Jetpack one:

Screen Shot 2020-11-03 at 9 41 14 AM

Or is that responsibilty on Jetpack and we can create an issue in that repo?

joshuatf:
>I'm not entirely sure how to test this.

If your install was already set to calypsoify, the above Jetpack calypsoify button should now be removed.  But if you need to force it on to test this, then you can add `&calypsoify=1` to the URL and then remove it to make sure calyspoify doesn't persist or add `update_user_meta( get_current_user_id(), 'calypsoify', 1 );` and note that calypsoify does not get turned on.

>Or is that responsibilty on Jetpack and we can create an issue in that repo?

I think we should handle this in the Jetpack repo if we want to add links to the nav or possibly filter the back button behavior.

psealock (approved):
Thanks for the clarifications @joshuatf, this is all looking good to me :shipit:

timmyc:
I don't know why it took me so long to set aside time and just enjoy what is happening code-wise in this PR, but dang I feel pretty good right now after reading this code. I can only imagine that it was a joy to delete things too. _The Joy of Deleting_ - seems like a good Bob Ross / coding riff.

I think all is looking great here, and will be happy to test it alongside the latest 4.8 release ( beta.2 it is sounding like ) that is set to come out tomorrow.

joshuatf:
@psealock Pushed a change in https://github.com/Automattic/wc-calypso-bridge/pull/614/commits/e1532db2d09ad8413f3247ce13e37237781f6d01 which fixes the OBW issue mentioned in https://github.com/Automattic/wc-calypso-bridge/issues/623

psealock:
>@psealock Pushed a change in e1532db which fixes the OBW issue mentioned in #623

@joshuatf Testing well! Thanks for the fix

joshuatf:
@timmyc @psealock Are we ready to merge this one?

timmyc:
@joshuatf I think we are safe to merge. If we get in a position where we have to for some reason do a bug fix on the existing calypsoify setup, we can branch off an old tag and cut a "release" if needed, but I don't imagine us having to do that. Knocks on wood. But yes, let's get the party rolling.

psealock:
There is an issue with Travis at the `composer install --no-dev` step for Woo Core

Screen Shot 2021-01-07 at 1 39 06 PM

I tried updating our packages in https://github.com/Automattic/wc-calypso-bridge/pull/614/commits/3419a30dbc7b1b21b01ec9bcef82dd436ce44df1, but that didn't help. I'll revert those changes

includes/class-wc-calypso-bridge-woocommerce-admin-features.php
```diff
@@ -50,6 +50,10 @@ public function filter_wc_admin_enabled_features( $features ) {
 			$features[] = 'remote-inbox-notifications';
 		}
 
+		if ( ! array_key_exists( 'navigation', $features ) && 'yes' === get_option( 'woocommerce_navigation_enabled', 'yes' ) ) {
```
psealock, line 53: This logic seems to be off. Although the option `woocommerce_navigation_enabled` will be on by default (which we need to do), it appears the feature will be turned off when the option is set to "no".

We should
* Initialize the option to "yes" if it doesn't exist, but respect it if its "no" or already set to "yes"
* Force the feature flag on regardless of the state of the option

psealock: ok, it further appears that I have confused myself! This is working well and I don't think requires a change

joshuatf:
@psealock Seems like the tests failing may be related to https://github.com/woocommerce/woocommerce/blob/9ebf70bf8dfcd13bdf1f3278ec5b841b2ef80da5/.travis.yml#L76

joshuatf:
Got tests passing.  This is related to WooCommerce requiring composer v2, while this project is still on v1.  This was a fairly trivial fix, but did require removing the `bjornjohansen/wp-pre-commit-hook` dependency as this was not compatible.

I'll open a separate issue to address pre commit hooks and getting that fixed up.

psealock:
Thanks Josh 🚢

PR #615, opened 2020-10-28
 Enable navigation feature by default

joshuatf:
Turns on the navigation by default when the ecomm plan is on, but allows it to be toggled on/off.

### Screenshot
Screen Shot 2020-10-28 at 10 45 17 AM

### Testing

1. Make sure you're running a version of WCA that has trhe nav feature.
1. Delete the `woocommerce_navigation_enabled` option from your database.
1. Deactivate the bridge or opt out of the ecomm plan option.
1. Make sure the new nav is not shown.
1. Opt in to the new ecomm plan and make sure the new navigation is shown on WC pages by default.
1. Navigate to Settings->Advanced->Features ( `wp-admin/admin.php?page=wc-settings&tab=advanced&section=features` ) and make sure you can toggle the nav feature on/off.
1. Note that an extra refresh may be required to see your settings (this is an old bug that has regressed and will need to be addressed in a follow-up).

### DIFF
--- a/includes/class-wc-calypso-bridge-setup.php
+++ b/includes/class-wc-calypso-bridge-setup.php
@@ -35,9 +35,10 @@ class WC_Calypso_Bridge_Setup {
 	 * Constructor.
 	 */
 	private function __construct() {
-		add_filter( 'default_option_woocommerce_onboarding_profile', array( $this, 'set_business_extensions_empty' ), 10, 1 );
-		add_filter( 'option_woocommerce_onboarding_profile', array( $this, 'set_business_extensions_empty' ), 10, 1 );
-		add_filter( 'woocommerce_admin_onboarding_themes', array( $this, 'remove_non_installed_themes' ), 10, 1 );
+		add_filter( 'default_option_woocommerce_onboarding_profile', array( $this, 'set_business_extensions_empty' ) );
+		add_filter( 'default_option_woocommerce_navigation_enabled', array( $this, 'enable_navigation_by_default' ) );
+		add_filter( 'option_woocommerce_onboarding_profile', array( $this, 'enable_navigation_by_default' ) );
+		add_filter( 'woocommerce_admin_onboarding_themes', array( $this, 'remove_non_installed_themes' ) );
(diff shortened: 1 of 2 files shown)


psealock (commented):
This is looking good @joshuatf. One more item that should be included in this PR is turning on of the feature flag. The [feature flags in wc-admin](https://github.com/woocommerce/woocommerce-admin/tree/main/config) will remain turned off for all but development. I'm pretty sure this filter can force them on:

https://github.com/Automattic/wc-calypso-bridge/blob/371094bc6b49de6b4f1f9f3fcea1c1a95dd55019/includes/class-wc-calypso-bridge-woocommerce-admin-features.php#L48

joshuatf:
Good call, @psealock.  Updated this to include the nav feature as long as the option flag is set to `yes`.

To test this as well:

1. Set `navigation` to `false` in `woocommerce-admin/includes/feature-config.php`.
1. Test again using the above testing instructions.

psealock (commented):
This is testing well for me. It even respects the option value of "no" which is great.

Something odd is going on because it consistently redirects me to the OBW by adding `&path=%2Fsetup-wizard`. I rebased from the branch in https://github.com/Automattic/wc-calypso-bridge/pull/614 to get this going. However https://github.com/Automattic/wc-calypso-bridge/pull/614 doesn't have the redirect. Are you experiencing this by chance?

psealock (approved):
The oddness I was seeing also happened on `master` so I'm not sure if something is up in my local or not, but it seems unrelated to these changes.

joshuatf:
>The oddness I was seeing also happened on master so I'm not sure if something is up in my local or not, but it seems unrelated to these changes.

I just started seeing the same thing on `master`.  Will check in WCA to see if an issue has been opened around this.

timmyc:
@joshuatf is it okay to merge this one to `remove/calypsoify` so we can have one branch with all the things to test out the nav with 4.8 beta 2?

joshuatf:
@timmyc Absolutely.  Go for it.

PR #618, opened 2020-11-17
Calypsoify: Remove the user meta mechanism

pablinos:
This is a start at removing the code which relies on, and sets, the
Calypsoify user meta value. This means that the Calypsoify styling won't
be applied unless the `calypsoify` query parameter is set.

On more investigation, this probably breaks some of the flows, so it
might be best to wait until Calypsoify isn't required anymore

The PR is very much WIP, and could well be obsolete after the changes to the navigation go live

#### WP PHPCS

To get past the pre-commit hooks I needed to upgrade the PHP standards
package and the XML config. One of the rules was throwing an error.

### DIFF
--- a/class-wc-calypso-bridge.php
+++ b/class-wc-calypso-bridge.php
@@ -102,18 +102,8 @@ class WC_Calypso_Bridge {
 	 * only hooks on admin_init which won't be run by wc-setup
 	 */
 	public function check_calypsoify_param() {
-		if ( isset( $_GET['calypsoify'] ) ) { // WPCS: CSRF ok.
-			if ( 1 === (int) $_GET['calypsoify'] ) { // WPCS: CSRF ok.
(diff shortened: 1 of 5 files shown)


psealock:
Thanks for the pull request @pablinos !

We have similar changes in progress in two PRs you can see at https://github.com/Automattic/wc-calypso-bridge/pulls. Those are pretty much ready to be merged, but any additional input is welcome.

pablinos:
Nice! I should have known you'd already be working on it. I left this as a draft as I figured it would probably get outdated by the navigation work. It was partly so I could link to it in this pbAPfg-15U-p2. I'll subscribe to that other PR, as I think once the need for Calypsoify is removed from here then we should be able to turn it off for everyone.

pablinos:
Closing in favour of #614

PR #627, opened 2021-01-06
Add feature toggles if they don't exist

### LINKED ISSUE #625: Navigation: Add Feature toggle
In https://github.com/woocommerce/woocommerce-admin/pull/5940, the feature toggle was removed from the 1.8.2 release of wc-admin which is the version eComm users will be using on January 12th. This feature toggle will need to be added back in from this repo so it is only visible to eComm users.

One potential gotcha here is that the toggle is still on the master branch because future releases will have it from wc-admin. I'm not sure how this filter handles duplicate entries, but perhaps there is a way to check if this feature toggle has been added already. Alternatively, we can make a note of it and handle the scenario with each Core release.

### Deliverables

* Add toggle removed in https://github.com/woocommerce/woocommerce-admin/pull/5940
* Investigate duplicate toggles potential solution

joshuatf:
Fixes #625

Adds in feature toggles if they don't exist in WCA.

### Screenshot
Screen Shot 2021-01-06 at 2 38 27 PM

### Testing

1. Make sure this branch works without error in the latest `main` of WCA.
2. Check out a branch of WCA that does not contain the feature toggles or comment them out in: https://github.com/woocommerce/woocommerce-admin/blob/c79542c6c6de3755abf5bceca9cefe22674ad9e9/src/Loader.php#L64-L65
3. Check that the features tab still exists under ecomm plans and that navigation can be toggled on/off.

### DIFF
--- a/README.md
+++ b/README.md
@@ -45,6 +45,7 @@ update_option( 'at_options', array( 'plan_slug' => 'ecommerce' ) );
 ```
 >>>>>>> 198c79f... Remove remaining references to calypsoify
 
+<<<<<<< HEAD
 `add_filter( 'jetpack_offline_mode', '__return_true' );`
 
 To make bridge work, the site must have the eCommerce plan.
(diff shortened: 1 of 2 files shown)


psealock:
I rebased the base branch so gave this a rebase as well

psealock (approved):
Nice Josh, this is working well with `main` and prod builds of 1.8.3 and code looks good.

Issue #626, opened 2021-01-04
Navigation: Add Learn More Note


psealock:
https://github.com/woocommerce/woocommerce-admin/pull/5837 Adds a note to "learn more" with a link to a document, https://docs.woocommerce.com/document/woocommerce-navigation.

Since that code will be missing from the 1.8.2 release included in the initial roll out, it will need to be added here.

Similar to https://github.com/Automattic/wc-calypso-bridge/issues/625, this creates a potential for duplicate notes in the future, so a check to see if it exists will be useful.

### Deliverables

* Add the note from https://github.com/woocommerce/woocommerce-admin/pull/5837
* Make sure to avoid future duplication

PR #628, opened 2021-01-06
Add bridge notes and cron events

joshuatf:
Fixes https://github.com/Automattic/wc-calypso-bridge/issues/626

* Adds cron events under the action `wc_calypso_bridge_daily` event name.
* Adds the navigation learn more note.

Up for discussion on whether this belongs here or not.  The events and notes give us a finer level of control inside the bridge, but obviously come with some level of overhead.

TODO: The CTA link still needs to be replaced.

### Testing

1. Make sure you're under dev mode for the ecomm plan.
2. Toggle the navigation on/off and make sure the note is added when it's turned on.
3. Delete the note while the nav is already turned on.
4. Run the `wc_calypso_bridge_daily` event.
5. Make sure the note is added.


psealock (approved):
This is all looking good and testing well, thanks Josh.

>Up for discussion on whether this belongs here or not.

I agree that it would belong in wc-admin, but now that 1.8.3 is already finalized it will have to be here. Plus, this provides the plumbing for additional eComm notes.

psealock:
@joshuatf The assets and link detailed in https://teamisotope.wordpress.com/2021/01/04/navigation-ecomm-release-prep-list/#comment-4782

timmyc:
Thanks for getting these notes ready Josh. Did we consider trying to use the remote inbox system for these at all? The one thing I'm still wondering about with adding notes like this is how it will look to remove/deprecate a note from within the bridge.

joshuatf:
@timmyc I think the remote inbox system may work for this.  The biggest drawback is that we would not be able to use the `wc_calypso_bridge_is_ecommerce_plan()` function to check if the note should be added, but I think we could use the option rule processor to manually check the option.

Not the most elegant in this case, but cc @becdetat in case there's a better way to use a local method that that I missed in the remote inbox system.

joshuatf:
Thanks @psealock and @timmyc for the doc!  Updated in https://github.com/Automattic/wc-calypso-bridge/pull/628/commits/c887f88240af90588a510d4e536836437021000e using the `page_id`.

psealock (approved):
>how it will look to remove/deprecate a note from within the bridge.

Would we ever need to do do this? Maybe if a similar note gets added to wc-admin we can provide a check avoid doing so if this note is already in the database.

At this point, lets roll with this solution as its working well and the resource linked looks fantastic. Thanks @joshuatf and @timmyc

joshuatf:
>Would we ever need to do do this? Maybe if a similar note gets added to wc-admin we can provide a check avoid doing so if this note is already in the database.

I think it's possible, in which case we could copy the plumbing used in WooCommerce Admin that removes notes into the bridge or we could simply add the note name into WooCommerce Admin since it removes by name.

I think it's safe to go ahead and merge this one for now and we can follow up on that if in the future if we need to remove notes.

PR #634, opened 2021-01-15
Filter the nav back URL for ecomm plan users

### LINKED ISSUE #633: Redirect users back to Calypso Home when clicking on the WordPress Dashboard menu item
Currently we're redirecting eCommerce plan users to the wp-admin dashboard when they click on the WordPress Dashboard menu item. However, it will be better to redirect users back to the Calypso Home, as we don't know if they manage the other parts of their website on Calypso or wp-admin. And they can always access wp-admin through the nav option in the Calypso menu.

joshuatf:
Fixes #633 

Filters the back URL used in the new navigation experience to redirect back to the Calypso dashboard.

@pmcpinto Do we also want to change the text from "WordPress Dashboard" on this back button or leave it as is?

### Screenshots
Screen Shot 2021-01-15 at 3 10 13 PM

### Testing

1. Create an ecomm plan site.
2. Load a WC page.
3. Navigate to the root of the navigation and click the back button "WordPress Dashboard"
4. Make sure this takes you to the site's Calypso dashboard. `http://wordpress.com/home/yoursite.com`

### DIFF
--- a/includes/class-wc-calypso-bridge-woocommerce-admin-navigation.php
+++ b/includes/class-wc-calypso-bridge-woocommerce-admin-navigation.php
@@ -1,0 +1,52 @@
+<?php
+/**
+ * Modify the navigation to fit the bridge needs.
+ *
+ * @package WC_Calypso_Bridge/Classes
+ */
+
+defined( 'ABSPATH' ) || exit;
+
+/**
+ * WC_Calypso_Bridge_WooCommerce_Admin_Navigation
+ */
+class WC_Calypso_Bridge_WooCommerce_Admin_Navigation {
+
+	/**
+	 * Class Instance.
+	 *
+	 * @var WC_Calypso_Bridge_WooCommerce_Admin_Navigation instance
+	 */
+	protected static $instance = null;
+
+	/**
+	 * Get class instance
+	 */
+	public static function get_instance() {
+		if ( ! static::$instance ) {
+			static::$instance = new static();
+		}
+		return static::$instance;
+	}
+
+	/**
+	 * Constructor.
+	 */
+	public function __construct() {
+		add_filter( 'woocommerce_navigation_root_back_url', array( $this, 'add_calypso_url' ) );
+	}
+
+	/**
+	 * Add calypso URL.
+	 */
+	public static function add_calypso_url() {
+		$strip_http = '/.*?:\/\//i';
+		$site_slug  = preg_replace( $strip_http, '', get_home_url() );
+		$site_slug  = str_replace( '/', '::', $site_slug );
+
+		$store_url = 'https://wordpress.com/home/' . $site_slug;
+		return $store_url;
+	}
+}
+
+WC_Calypso_Bridge_WooCommerce_Admin_Navigation::get_instance();
--- a/class-wc-calypso-bridge.php
+++ b/class-wc-calypso-bridge.php
@@ -112,6 +112,7 @@ class WC_Calypso_Bridge {
 		include_once dirname( __FILE__ ) . '/includes/class-wc-calypso-bridge-addons-screen.php';
 		include_once dirname( __FILE__ ) . '/includes/gutenberg.php';
 		require_once dirname( __FILE__ ) . '/includes/class-wc-calypso-bridge-woocommerce-admin.php';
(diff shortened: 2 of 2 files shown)


pmcpinto:
> @pmcpinto Do we also want to change the text from "WordPress Dashboard" on this back button or leave it as is?

You can leave it as is. Thanks!

includes/class-wc-calypso-bridge-woocommerce-admin-navigation.php
```diff
@@ -0,0 +1,52 @@
...
+		add_filter( 'woocommerce_navigation_root_back_url', array( $this, 'add_calypso_url' ) );
+	}
+
+	/**
+	 * Add calypso URL.
+	 */
+	public static function add_calypso_url() {
+		$strip_http = '/.*?:\/\//i';
+		$site_slug  = preg_replace( $strip_http, '', get_home_url() );
+		$site_slug  = str_replace( '/', '::', $site_slug );
```
timmyc, line 45: Is this line to catch the cases where WP might be installed in a subdirectory? Just looks vaguely familiar for calypso. If so, thinking we might not need that check since all wpcom sites have WP installed in the root. Also was wondering if there might be a way to get the site "slug" from an atomic option possibly.

timmyc:
Thanks for getting this PR up, this will address one of the feedback items shared this week from happiness 🌟

mattsherman:
Could we look at the referrer or something (another idea set a query param when coming from Calypso?) when we first get into WooCommerce section so that we could take the user back to wp-admin if they came from wp-admin and back to Calypso if they came from Calypso?

joshuatf:
I think that's a solid idea @mattsherman.  @pmcpinto Does that sound okay to you?

pmcpinto:
> I think that's a solid idea @mattsherman. @pmcpinto Does that sound okay to you?

👍

includes/class-wc-calypso-bridge-woocommerce-admin-navigation.php
joshuatf: >Also was wondering if there might be a way to get the site "slug" from an atomic option possibly.

This would be great.  Does not look like there are any shorthand methods for grabbing it under https://github.com/Automattic/wpcomsh/blob/edc01bb249ebc0ca1aa6945319503997346ef764/class.atomic-plan-manager.php

Any ideas on how we might go about finding an option like this if it exists?

>Is this line to catch the cases where WP might be installed in a subdirectory?

I think it's to catch home URLs that may not be at the root and prevent an extra slash from being added to the URL.  E.g.,

`mysite.com/home` -> `mysite.com::home`

joshuatf:
@mattsherman On digging into this a bit more, I'm not quite sure how easily it would work in practice.

The original thought I had was if they were referred from Calypso, then use the Calypso URL as the back URL.  However, after a single page visit, Calypso will no longer be the referrer.

If we cache this value (have you ever been referred from Calypso) this answer should be a consistent "yes" across the board for Atomic sites.

Did you have a different idea on how this should work?

psealock (approved):
Nice, this is working well and the wp.com url is appearing :shipit:

psealock:
>If we cache this value (have you ever been referred from Calypso) this answer should be a consistent "yes" across the board for Atomic sites.

We can cache the value (in localStorage?) and update that value each time the user lands on a WooCommerce page. If the referrer is wp-admin but not a Woo page, then stay with wp-admin. If its Calypso, then use the Calypso value. This would need to be a JS solution though, and a good follow up so we can get this part merged.

mattsherman:
> > If we cache this value (have you ever been referred from Calypso) this answer should be a consistent "yes" across the board for Atomic sites.
> 
> We can cache the value (in localStorage?) and update that value each time the user lands on a WooCommerce page. If the referrer is wp-admin but not a Woo page, then stay with wp-admin. If its Calypso, then use the Calypso value. This would need to be a JS solution though, and a good follow up so we can get this part merged.

That's the general idea I had in my head.

psealock:
Great idea @mattsherman to return users from whence they came. I created https://github.com/Automattic/wc-calypso-bridge/issues/640 to track that so we can go ahead and merge this one as a first iteration.

PR #638, opened 2021-02-03
Set survey source param for ecomm and business plans

### LINKED ISSUE #636: Update the survey tags for ecomm and business plans
Following up on https://github.com/woocommerce/woocommerce-admin/pull/6066 we should use the `woocommerce_admin_survey_query` filter to add in arguments to determine if a user is coming from a business or ecomm plan.

```php
add_filter( 'woocommerce_admin_survey_query', function( $query ) {
	$query['source'] = 'ecomm-plan';
	return $query;
} );

### PR DESCRIPTION
**Pull request by joshuatf**
Fixes #636 

### Screenshots

Screen Shot 2021-02-03 at 5 14 24 PM
Screen Shot 2021-02-03 at 5 14 15 PM

### Testing

1. Activate a fake atomic plan manager plugin and test by changing the return of `current_plan_slug()`:
```php
<?php
/**
 * Plugin Name: Atomic Plan Manager
 */

class Atomic_Plan_Manager {
	/**
	 * Business plan slug
	 * @var string
	 */
	const BUSINESS_PLAN_SLUG   = 'business';

	/**
	 * Ecommerce plan slug
	 * @var string
	 */
    const ECOMMERCE_PLAN_SLUG  = 'ecommerce';
    
    /**
     * Get current plan slug
     */
    public static function current_plan_slug() {
        return self::BUSINESS_PLAN_SLUG;
    }
}
```
2. Trigger a note with a survey (e.g., `wc-admin-navigation-feedback` can be triggered on the daily `wc_admin_daily` cron event if the nav feature is enabled)
3. Check that the survey URL ("Share Feedback") contains the `source` param with the respective plan value.

### DIFF
--- a/includes/class-wc-calypso-bridge-tracks.php
+++ b/includes/class-wc-calypso-bridge-tracks.php
@@ -47,6 +47,7 @@ class WC_Calypso_Bridge_Tracks {
 		add_filter( 'admin_footer', array( $this, 'add_tracks_js_filter' ) );
 		add_filter( 'woocommerce_tracks_event_properties', array( $this, 'add_tracks_php_filter' ), 10, 2 );
 		add_filter( 'woocommerce_get_sections_advanced', array( $this, 'hide_woocommerce_com_settings' ), 10, 1 );
+		add_filter( 'woocommerce_admin_survey_query', array( $this, 'set_survey_source' ) );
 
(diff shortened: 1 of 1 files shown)


psealock (commented):
code LGTM :shipit:

timmyc (approved):
LGTM too.

Issue #640, opened 2021-02-09
Navigation: Back URL filter history aware


psealock:
https://github.com/Automattic/wc-calypso-bridge/pull/634 makes use of a Navigation feature to direct eComm plan users back to Calypso instead of wp-admin. However, sometime eComm plan users will have arrived at WooCommerce via wp-admin.

### Deliverables

* Create a mechanism such that clicking "back to Dashboard" returns eComm user from where they came: wp-admin OR Calypso

### Previous Thoughts

See https://github.com/Automattic/wc-calypso-bridge/pull/634#issuecomment-772982513

>We can cache the value (in localStorage?) and update that value each time the user lands on a WooCommerce page. If the referrer is wp-admin but not a Woo page, then stay with wp-admin. If its Calypso, then use the Calypso value. This would need to be a JS solution though, and a good follow up so we can get this part merged.

PR #655, opened 2021-02-24
Add simple front-end tooling

joelclimbsthings:
We've had [a recent issue](https://github.com/Automattic/wc-calypso-bridge/issues/640) that requires the use of front-end tooling to solve properly, and it seems that this need will only grow in the future. This PR adds simple dev tooling using the architecture provided by Gutenberg and `npm run create-wc-extension`

I've also amended the README to include instructions to utilize this tooling.

**Test Instructions**
- Checkout branch
- Run `npm install` and `npm start`
- Uncomment `console.info` statement in `index.js` file
- Confirm that script is loaded and log statement is printed
- Check live reloading by changing log statement and reloading


samueljseay (approved):
lgtm 🚀

psealock:
Nice Joel! What do you think about changing the file structure such that `src` is PHP files and the js you introduced becomes `client` since that is the way things are in wc-admin? I could make a similar change to the create extension script.

joelclimbsthings:
I'm up for that, @psealock , to keep things consistent.

PR #656, opened 2021-02-25
Notes: Hide db update notices from being shown.

timmyc:
This is just a `try` branch for now, see p9o2xV-1fS-p2 for context and background.

Some feedback shared in that p2 post is that it would be nice to automate the execution of WooCommerce database updates on behalf of Atomic customers. In 5.0, the atomic team was kind enough to do this, but it was noted on the post that even though the update was done via wp-cli, this notice was still shown in the UI... which could be quite confusing.

store-alert-update-db-thanksbro

This branch implements a way for us to entirely suppress all notifications with the type of `update` from being shown in the Store Alerts area in wc-admin. I think a more elegant solution might be to add a filter(s) to wc-admin that would allow for adjusting the note `type`s that are used in the queries for Store Alerts [#](https://github.com/woocommerce/woocommerce-admin/blob/main/src/Features/ActivityPanels.php#L65) [#](https://github.com/woocommerce/woocommerce-admin/blob/main/client/layout/store-alerts/index.js#L283-L288).

But given we have some lag-time between wc-admin releases and core releases, we may need to ship this interim solution if there are db updates we hope to automate between here and there.

__To Test__
- First you will need to get the database update alert to show. The easiest way to do so is to locate the note named `wc-update-db-reminder` and set the status to `unactioned`
- After tweaking the note db entry, visit the home screen and verify the store alert for db updates is shown
- Download this branch, refresh the page, and verify it is no longer shown *and* the loading indicator for Store Alerts isn't shown either.

__After Screenshot__
db-update-note-gone

### DIFF
--- a/includes/class-wc-calypso-bridge-filters.php
+++ b/includes/class-wc-calypso-bridge-filters.php
@@ -8,6 +8,7 @@
  */
 
 defined( 'ABSPATH' ) || exit;
+use Automattic\WooCommerce\Admin\Notes\Notes;
 
 /**
  * WC Calypso Bridge Filters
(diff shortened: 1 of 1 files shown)


includes/class-wc-calypso-bridge-filters.php
```diff
@@ -40,8 +41,43 @@ private function __construct() {
...
+		add_filter( 'woocommerce_shared_settings', array( $this, 'component_settings' ), 30 );
 	}
 
+	/**
+	 * Adjust alert count to the component settings, to prevent loading state from being shown on db updates.
+	 *
+	 * @param array $settings Component settings.
+	 */
+	public function component_settings( $settings ) {
+		$settings['alertCount'] = Notes::get_notes_count( array( 'error' ), array( 'unactioned' ) );
```
timmyc, line 58: This state hydration value is used to determine if the loading indicator for Store Alerts should be shown. As such, we need to remove `update` from the array of types passed in here.

includes/class-wc-calypso-bridge-filters.php
```diff
@@ -40,8 +41,43 @@ private function __construct() {
...
+		return $settings;
+	}
+
+	/**
+	 * Remove the "update" note type from REST API requests to surpress db notices from being displayed.
+	 *
+	 * @param array  $args REST request args
+	 * @param object $request WP_REST_Request
+	 */
+	public function remove_update_note_type_from_args( $args, $request ) {
```
timmyc, line 68: I tried a few different approaches here, but this one seemed the most targeted to the use case of store alert notices being requested from the REST API.

psealock (commented):
I think you're right the more elegant solution is to filter the [`'error', 'update'`](https://github.com/woocommerce/woocommerce-admin/blob/ec8471dcadc2a91f1e972c84c2d6970e5c84900a/src/Features/ActivityPanels.php#L65) array in PHP and pass the resulting value to the client where the request is made. But alas, timing issues mean changes in this PR make sense.

I went ahead and made an issue to track this https://github.com/woocommerce/woocommerce-admin/issues/6511. I think a comment in the code to the effect that this is temporary code until the solution is released in Woo Core would be great.

moon0326:
Closing this PR as timmy is no longer with Automattic and it looks like this PR has been open for while.

PR #659, opened 2021-03-03
Filtering wc-navigation root back url conditionally when referred by Calypso

joelclimbsthings:
As a further iteration on the work done in https://github.com/Automattic/wc-calypso-bridge/pull/634, we're moving this logic to the front-end so that we can only filter the URL conditionally, depending on the origin of the user. 

In short, this PR will alter the behavior for the "WordPress Dashboard" link via the new navigation. It will conditionally take the user back to Calypso only if that's where they originated, or retain current URL back to  `wp_admin` otherwise. 

_Notes:_
- This PR is dependent on [another PR in `wc-admin`](https://github.com/woocommerce/woocommerce-admin/pull/6505) to actually work

### Testing Instructions
- Checkout branch & following instructions in [the readme](https://github.com/Automattic/wc-calypso-bridge/blob/master/README.md) to put your store in eComm mode (including [new revised gist](https://gist.github.com/psealock/531205e2c3d37be1d8ac4d3ef4f346bc))
- Checkout branch from [companion PR](https://github.com/woocommerce/woocommerce-admin/pull/6505) in `wc-admin` and activate.
- Ensure new navigation is enabled.
- Navigate from WP Admin to any WooCommerce page, click around several times to various pages, and then hit "WordPress Dashboard" link in nav. 
- Observe that you're returned to WP Admin as expected.
- You need to simulate a referred click from Calypso. There may be a better way, but this is what I've been doing: go to `https://wordpress.com/home`, and then modify the DOM to include a link back to your Store. I used the  following script, pasted in the console: _(change out `YOUR_LOCAL_URL_HERE` for the URL to your local)_
```js
const el = document.createElement( 'div' );
el.classList.add( 'site', 'is-private' );
el.innerHTML = `<div class="site is-private">
<a
  class="site__content"
  title="afkp2.wordpress.com"
  aria-label="afkp2.wordpress.com"
  href="YOUR_LOCAL_URL_HERE/wp-admin/admin.php?page=wc-admin"
  ><div
    class="site-icon is-blank"
    style="height: 32px; width: 32px; line-height: 32px; font-size: 32px"
  >
    <svg
      xmlns="http://www.w3.org/2000/svg"
      viewBox="0 0 24 24"
      class="gridicon gridicons-globe"
      height="18"
      width="18"
    >
      <use
        xlink:href="/calypso/images/gridicons-84d04a83ed8c3cfc40de995e9bd32649.svg#gridicons-globe"
      ></use>
    </svg>
  </div>
  <div class="site__info">
      <div class="site__title">BACK TO LOCAL</div>
      <div class="site__domain">test.link</div>
    <span class="site__badge site__badge-private">Private</span>
  </div></a
>
</div>`;

document.querySelector( '.site-selector__sites' ).appendChild( el );
```
- Click on the link that was just created at the bottom of your site selector
image

- You'll be directed to your local WC Admin dashboard. Click around to various Woo Pages, and then click the "WordPress Dashboard" link.
- It should redirect you back to Calypso.


src/utils/storage.js
```diff
@@ -0,0 +1,28 @@
...
+				);
+			}
+		},
+
+		reset: () => {
+			storage.set( defaultValue );
+		},
+	};
+
+	return window.localStorage ? storage : null;
```
psealock, line 27: How about putting this check at the function level? I would want `storage( 'id' ).get()` to return `null` instead of fatal error if `window.localStorage` doesn't exist.

src/wc-navigation-root-url.js
```diff
@@ -0,0 +1,41 @@
...
+	} else if ( ! isWooPage ) {
+		rootUrlStorage.set( 'wcadmin' );
+	}
+
+	// Only continue to add filter if referred by Calypso
+	if ( ! ( rootUrlStorage.get() === 'calypso' ) ) {
+		return;
+	}
+
+	addFilter( 'woocommerce_navigation_root_back_url', 'plugin-domain', () =>
```
psealock, line 38:
```suggestion
	addFilter( 'woocommerce_navigation_root_back_url', 'wc-calypso-bridge', () =>
```

src/wc-navigation-root-url.js
```diff
@@ -0,0 +1,41 @@
...
+		document.referrer.indexOf( 'https://wordpress.com' ) === 0;
+
+	if ( fromCalypso ) {
+		rootUrlStorage.set( 'calypso' );
+	} else if ( ! isWooPage ) {
+		rootUrlStorage.set( 'wcadmin' );
+	}
+
+	// Only continue to add filter if referred by Calypso
+	if ( ! ( rootUrlStorage.get() === 'calypso' ) ) {
```
psealock, line 34: Maybe easier to read?

```suggestion
	if ( rootUrlStorage.get() !== 'calypso' ) {
```

class-wc-calypso-bridge.php
```diff
@@ -133,24 +132,32 @@ public function load_ecommerce_plan_ui() {
...
 			$script_asset['version'],
 			true
 		);
+
+		wp_add_inline_script( 'wc-calypso-bridge', 'window.wcCalypsoBridge = ' . wp_json_encode(
+			array(
+				'isWooPage' => $is_woo_page,
+				'homeUrl'   => get_home_url()
+			)
+		), 'before' );
```
psealock, line 160: There's a system for this kind of thing that WC Admin uses and Blocks as well to retrieve values with `@woocommerce/wc-admin-settings`. See it used in the homescreen:

https://github.com/woocommerce/woocommerce-admin/blob/ec8471dcadc2a91f1e972c84c2d6970e5c84900a/src/Features/Homescreen.php#L53

The intension was for extensions to also use it. I'm rather torn on whether or not we should use it here. Its the "right" way to do things, but that system is less than ideal because the `getSetting` method was a mistake because it manipulates the object on the window. We mitigated that to a certain extent with our data stores' hydration. I'm feeling like its probably best to avoid it, but just wanted to share some background in case you had any feelings either way.

psealock (commented):
Lovely to see this coming together @joelclimbsthings !

This is testing well for me. The correct redirect is happening even when I navigate to a legacy Woo page, very nice! I just have a few minor suggestions and one point to think about. Also there are git conflict marks in the `package-lock.json` that will need to be dealt with.

src/utils/storage.js
joelclimbsthings: Yes, that makes sense. Just refactored the `storage` module more along these lines.

class-wc-calypso-bridge.php
joelclimbsthings: Thanks for this insight @psealock , good to see the context here. It looks like that filter ultimately enables you to add properties to `wcSettings` on the `window`, although that's only available when you're on a Woo page. In this case I'd ideally have access when on _any_ page (including WP Admin). 

Also, it looks like this filter is applied here: https://github.com/woocommerce/woocommerce-gutenberg-products-block/blob/3c55aac1b599a5e4a8d2bf8ec56e89f6c56b3ad5/src/Assets/AssetDataRegistry.php#L130

But that also includes this comment, which makes it sound like we shouldn't be using it?
```

joelclimbsthings:
Thanks @psealock , good thoughts. I believe I've addressed all of the feedback, and restored `package-lock.json`. Also had one question about one of your comments.

class-wc-calypso-bridge.php
psealock: Ah thats right, we should be using `Automattic\WooCommerce\Blocks\Assets\AssetDataRegistry::class` but the old way `woocommerce_shared_settings` is still supported. 

>only available when you're on a Woo page

This would be a blocker then. Thanks for looking into it!

psealock (approved):
Nice, love the changes to `storage.js`. Thanks @joelclimbsthings :shipit:

psealock:
Have you seen the error messges on CI? I wonder if that has to do with our new dev environment with JS scripts now?

joelclimbsthings:
> Have you seen the error messges on CI? I wonder if that has to do with our new dev environment with JS scripts now?

Yeah, it looks like the failure just stems from the PHP linting issues. I think this will be best addressed as part of the in-process Housecleaning epic.

Issue #637, opened 2021-01-27
Various WC_Install::install() Methods Not Running on eCom Sites


timmyc:
This feedback came in via the WPCOM Happiness Townhall yesterday where @pmcpinto gave a presentation about the eCommerce Plan.

Post about this topic on the Atomic P2: p9o2xV-1eZ-p2

As the title states, the bug is that when a new ecommerce plan site is launched, the default WooCommerce pages ( Shop, Cart, Checkout Etc ) are not created, which is problematic for new Woo users.

I spent a little time digging in to where this logic is triggered, and the logic is housed in `WC_Install::install()` method which in-turn calls a method called [`maybe_create_pages`](https://github.com/woocommerce/woocommerce/blob/56e67b1ef8f4b521c349780e2dd38f84cf513358/includes/class-wc-install.php#L311). That method only checks to see if the site option for `woocommerce_db_version` is present, if it is, the install of Woo is considered an upgrade, and no default pages are set.

On a brand-new ecomm site created today, I verified that this behavior still exists, no default pages created, and after the site is provisioned, the db version looks correct:

![image](https://user-images.githubusercontent.com/22080/106062926-4dbff800-60ac-11eb-8fd4-351937e0cd8a.png)

I also confirmed that the page creation behavior is working as anticipated in a 4.9 install on a Jurassic Ninja site:

![image](https://user-images.githubusercontent.com/22080/106063178-a394a000-60ac-11eb-9969-62bd8bf02f2a.png)

## Missing Cron Jobs Too

I also wanted to see if any other logic in the `install()` method was not behaving as expected, so I also looked into [`WC_Install::create_cron_jobs()`](https://github.com/woocommerce/woocommerce/blob/56e67b1ef8f4b521c349780e2dd38f84cf513358/includes/class-wc-install.php#L512), and it too appears to not be running on install on the eCommerce Plan sites. So I'm a bit perplexed as parts of the install logic ( database creation and updating db version option ) are running, but so far at least 2 others are not.

__Cron jobs prefixed with `woocommerce_` on new JN Install__
![image](https://user-images.githubusercontent.com/22080/106063983-f0c54180-60ad-11eb-8e2d-c69bbc811669.png)

__Cron jobs prefixed with `woocommerce_` on eComm new Install__
![image](https://user-images.githubusercontent.com/22080/106064052-0e92a680-60ae-11eb-91cf-5a1f026a1519.png)

[...]

PR #662, opened 2021-03-12
Create WooCommerce pages if they are missing on a new installation

moon0326:
Fixes WooCommerce page issue in [636](https://github.com/Automattic/wc-calypso-bridge/issues/637) 

### Detailed test instructions

**Before getting started**:

1. Make sure to delete WooCommerce & WooCommerce Admin plugins. 
2. Make sure to delete `woocommerce_db_version` and `woocommerce_version` options from the `wp_options` table. You can use the following query.
```
delete from wp_options where option_name='woocommerce_version' or option_name='woocommerce_admin_version'
```
3. Make sure to delete WooCommerce pages (Cart, Checkout, My account, Shop) from the previous installation. If you have those pages in WordPress -> Pages, please delete them first.

Start with a fresh installation if you want to skip these steps.

**To test**:

1. Checkout this branch
2. Make sure you're on an eCommerce plan. If you're testing locally, use this [plugin](https://gist.github.com/psealock/531205e2c3d37be1d8ac4d3ef4f346bc).
3. Activate `wc-calypso-bridge` plugin (this plugin)
4. Navigate to WordPress -> Plugins -> Add New and install & activate WooCommerce
5. Confirm that you have Cart, Checkout, My account, and Shop pages in WordPress -> Pages

### DIFF
--- a/includes/class-wc-calypso-bridge-plugins.php
+++ b/includes/class-wc-calypso-bridge-plugins.php
@@ -39,6 +39,8 @@ class WC_Calypso_Bridge_Plugins {
 		add_action( 'update_option_active_plugins', array( $this, 'prevent_woocommerce_deactivation' ), 10, 2 );
 		add_action( 'current_screen', array( $this, 'prevent_woocommerce_deactiation_route' ), 10, 2 );
 		add_action( 'admin_notices', array( $this, 'prevent_woocommerce_deactiation_notice' ), 10, 2 );
+		add_action( 'woocommerce_admin_newly_installed', array( $this, 'maybe_create_wc_pages' ), 10, 2 );
+
 	}
 
 	/**
@@ -99,6 +101,18 @@ class WC_Calypso_Bridge_Plugins {
 		return $actions;
 	}
 
+	/**
+	 * Check WooCommerce pages (shop, cart, my-account, checkout) and create them if they don't exist.
+	 */
+	public function maybe_create_wc_pages() {
+		global $wpdb;
+
+		$post_count = $wpdb->get_var( "select count(*) from $wpdb->posts where post_name in ('shop', 'cart', 'my-account', 'checkout')" );
+
+		if ( 4 !== (int) $post_count ) {
+			WC_Install::create_pages();
+		}
+	}
 
 }
 $wc_calypso_bridge_plugins = WC_Calypso_Bridge_Plugins::get_instance();


includes/class-wc-calypso-bridge-plugins.php
```diff
@@ -99,6 +101,18 @@ public function remove_woocommerce_deactivation_link( $actions, $plugin_file ) {
 		return $actions;
 	}
 
+	/**
+	 * Check WooCommerce pages (shop, cart, my-account, checkout) and create them if they don't exist.
+	 */
+	public function maybe_create_wc_pages() {
+		global $wpdb;
+
+		$post_count = $wpdb->get_var( "select count(*) from $wpdb->posts where post_name in ('shop', 'cart', 'my-account', 'checkout')" );
```
timmyc, line 110: I like this check @moon0326 - I think its also worth noting that `create_pages()` calls `wc_create_page()` for each of the pages, which also [does a check](https://github.com/woocommerce/woocommerce/blob/23710744c01ded649d6a94a4eaea8745e543159f/includes/admin/wc-admin-functions.php#L74-L76) to ensure the page with the slug doesn't exist before creating the page.

timmyc:
I haven't fully tested this out yet, but it is looking good Moon. I think we can't fully test this without being able to re-create the buggy-state that causes this to happen on Atomic. So feeling like we might need to test as you have outlined, merge/release and test in production to verify all has been fixed.

includes/class-wc-calypso-bridge-plugins.php
moon0326: Thank you! Yup, I also double-checked that calling the function does not overwrite any existing pages.

moon0326:
> I haven't fully tested this out yet, but it is looking good Moon. I think we can't fully test this without being able to re-create the buggy-state that causes this to happen on Atomic. So feeling like we might need to test as you have outlined, merge/release and test in production to verify all has been fixed.

I agree. It's a bit tricky to test. I think the code itself is relatively safe to deploy.

pmcpinto:
@moon0326 we continue getting reports that this is happening (see this [comment](https://github.com/Automattic/wp-calypso/issues/39850#issuecomment-805639151)). 
Is it possible to merge it and ship it in the upcoming days?

moon0326:
> @moon0326 we continue getting reports that this is happening (see this [comment](https://github.com/Automattic/wp-calypso/issues/39850#issuecomment-805639151)).
> Is it possible to merge it and ship it in the upcoming days?

I will ask the team to review it today.

mattsherman:
I will review this shortly.

mattsherman (approved):
I have tested this and verified that the pages get created as expected. Nice work!

moon0326:
@pmcpinto We have released a new release with this change, but it did not fix the bug on wordpress.com for some reason. I'm going to take a look at this issue again.

moon0326:
@pmcpinto After investigating it a bit further, I've decided that this should be fixed in WooCommerce core. I've submitted a new issue https://github.com/woocommerce/woocommerce/issues/29491

pmcpinto:
👍 thanks for the heads up.

PR #663, opened 2021-03-22
Setup: Prevent Creative Mail and Crowdsignal Forms redirects when activating plugins

mattsherman:
This PR addresses https://github.com/Automattic/wp-calypso/issues/49925

It uses the same approach as originally used to prevent MailChimp redirects (https://github.com/Automattic/wc-calypso-bridge/pull/478).

## Test instructions

To reproduce the redirect locally, all one must do is install and activate the Creative Mail and Crowdsignal Forms plugins. It might be good to do this once before you test this branch. Then deactivate/delete the plugins.

1. Check out this branch
2. Install and activate the Creative Mail plugin
3. Verify that you end up on the Setup dashboard instead of the Creative Mail settings page
4. Install and activate Crowdsignal Forms plugin
5. Verify that you end up on the Setup dashboard instead of the Crowdsignal Forms plugin

Once this is merged and a wpcomsh release is done, we can test this in production to formally verify that it fixes https://github.com/Automattic/wp-calypso/issues/49925

### DIFF
--- a/README.md
+++ b/README.md
@@ -74,6 +74,8 @@ The ecommerce plan comes bundled with a number of plugins that this plugin integ
 - Marketing:
   - [MailChimp for WooCommerce](https://wordpress.org/plugins/mailchimp-for-woocommerce/)
   - [Facebook  for WooCommerce](https://woocommerce.com/products/facebook/)
+  - [Creative Mail](https://wordpress.org/plugins/creative-mail-by-constant-contact/)
+  - [Crowdsignal Forms](https://wordpress.org/plugins/crowdsignal-forms/)
 - Store Management:
 
   - [TaxJar -- Sales Tax Automation for WooCommerce](https://wordpress.org/plugins/taxjar-simplified-taxes-for-woocommerce/)
--- a/README.md
+++ b/README.md
@@ -60,43 +60,43 @@ If you would like to skip all of the above, [just download this gist](https://gi
 The ecommerce plan comes bundled with a number of plugins that this plugin integrates with if activated. To fully test this plugin's functionality, the following plugins can be installed.
 
 - Payments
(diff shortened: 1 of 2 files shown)


ilyasfoo (commented):
Overall changes look good! Some comments:

- I think adding Creative Mail and Crowdsignal to the readme.md under `Plugin Integrations` would be helpful.
- I was able to confirm that activating Creative Mail goes to WooCommerce home, but I wasn't able to get Crowdsignal redirects (even before checking out this branch). Might be something in my local preventing it from doing so. Would love to get another pair of 👀 to approve this!

mattsherman:
> * I think adding Creative Mail and Crowdsignal to the readme.md under `Plugin Integrations` would be helpful.

Done!

> * I was able to confirm that activating Creative Mail goes to WooCommerce home, but I wasn't able to get Crowdsignal redirects (even before checking out this branch). Might be something in my local preventing it from doing so. Would love to get another pair of 👀 to approve this!

Ah, I forgot to mention that. I too wasn't getting Crowdsignal redirects in my local env, though I have seen it in production. I'm not sure what's going on there, but I was able to get the redirects to happen by manually setting the option via WP-CLI after activating Crowdsignal (and then refreshing the page)...

```
wp option add crowdsignal_forms_do_activation_redirect true
```

ilyasfoo (approved):
>Ah, I forgot to mention that. I too wasn't getting Crowdsignal redirects in my local env, though I have seen it in production. I'm not sure what's going on there, but I was able to get the redirects to happen by manually setting the option via WP-CLI after activating Crowdsignal (and then refreshing the page)

Tried that and I confirm the redirect works for Crowdsignal. 

Thanks for cleaning up too! LGTM 🚢

PR #666, opened 2021-03-25
Delete woocommerce_*_page_id options before attempting to create WC pages and use wc_admin_daily action

moon0326:
This is a follow-up PR for https://github.com/Automattic/wc-calypso-bridge/pull/662

The previous PR did not work on ECOMMERCE plan. It looks `maybe_create_wc_pages` never gets called.

My best guess is that `wc-calypso-bridge` plugin [does not add](https://github.com/Automattic/wc-calypso-bridge/blob/master/wc-calypso-bridge.php#L20) any actions because WooCommerce is not available by the time it is activated.

1. wpcomsh gets installed & activated. `wc-calypso-bridge` does not add any actions at this time because WooCommerce is not available.
2. Other ecommerce plan plugins including WooCommerce gets installed & activated via a script, so `wc-calypso-bridge` does not get a chance to add actions including `woocommerce_admin_newly_installed`. 

This PR uses `wc_admin_daily` instead because it's executed by the scheduler. It should give `wc-calypso-bridge` enough time to add actions after WooCommerce gets installed. I think `wc_admin_daily` is the best action we can use for now.

### DIFF
--- a/includes/class-wc-calypso-bridge-plugins.php
+++ b/includes/class-wc-calypso-bridge-plugins.php
@@ -109,10 +109,10 @@ class WC_Calypso_Bridge_Plugins {
 		$post_count = $wpdb->get_var( "select count(*) from $wpdb->posts where post_name in ('shop', 'cart', 'my-account', 'checkout')" );
 
 		if ( 4 !== (int) $post_count ) {
-	    	//reset the woocommerce_*_page_id options
-            foreach (['shop', 'cart', 'myaccount', 'checkout'] as $page) {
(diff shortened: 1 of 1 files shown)


ilyasfoo (approved):
Great sleuthing here @moon0326! I've tested this using the instructions from #662 by zipping this branch and activated in a JN site. I've confirmed the pages are created after. I've also tested and confirmed `wc_admin_daily` does in fact run once after initial plugin activation. However, I did not attempt to reproduce the original issue.

Approving this, as I can't think of any other way than to test it in production. 🚢

PR #670, opened 2021-03-31
Added a new class to register actions even without WooCommerce detected

moon0326:
This PR is another attempt to fix ECOM page bug described in https://github.com/Automattic/wc-calypso-bridge/issues/637

I added a new class to register actions even without WooCommerce detected.

When a user signs up for an ecommerce plan (this is my best guess),
 
1. Install & Activate MU plugins. `wc-calypso-bridge` does not register anything because WooCommerce is not installed yet. See [here](https://github.com/Automattic/wc-calypso-bridge/blob/master/wc-calypso-bridge.php#L20).
2. Install & Activate WooCommerce and other plugins. Since the plugins are installed & activated via a script in a single session, `wc-calypso-bridge` does not get a chance to add any actions. I think that's the reason why both `woocommerce_admin_newly_installed` and `wc_admin_daily` callbacks never worked in my previous PRs.

This PR initializes `WC_Calypso_Bridge_Mu_Events` class before [this line](https://github.com/Automattic/wc-calypso-bridge/blob/master/wc-calypso-bridge.php#L20) to make sure the actions get registered.

### DIFF
--- a/includes/class-wc-calypso-bridge-mu-events.php
+++ b/includes/class-wc-calypso-bridge-mu-events.php
@@ -1,0 +1,59 @@
+<?php
+/**
+ * Events that must be registered even without WooCommerce detected.
+ *
+ * @package WC_Calypso_Bridge/Classes
+ * @since   1.0.0
+ * @version 1.0.0
+ */
+
+defined( 'ABSPATH' ) || exit;
+
+/**
+ * WC Calypso Bridge Mu Events
+ */
+class WC_Calypso_Bridge_Mu_Events {
+
+	/**
+	 * Class instance.
+	 *
+	 * @var WC_Calypso_Bridge_Themes_Setup instance
+	 */
+	protected static $instance = false;
+
+	/**
+	 * Get class instance
+	 */
+	public static function get_instance() {
+		if ( ! self::$instance ) {
+			self::$instance = new self();
+		}
+		return self::$instance;
+	}
+
+	/**
+	 * Constructor.
+	 */
+	private function __construct() {
+		add_action( 'woocommerce_admin_newly_installed', array( $this, 'maybe_create_wc_pages' ), 10, 2 );
+	}
+
+	/**
+	 * Check WooCommerce pages (shop, cart, my-account, checkout) and create them if they don't exist.
+	 */
+	public function maybe_create_wc_pages() {
+		global $wpdb;
+
+		$post_count = $wpdb->get_var( "select count(*) from $wpdb->posts where post_name in ('shop', 'cart', 'my-account', 'checkout')" );
+
+		if ( 4 !== (int) $post_count ) {
+			// reset the woocommerce_*_page_id options.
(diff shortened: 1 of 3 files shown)


timmyc:
Sorry to hear your last fix didn't stick... I'm a bit curious though, did it work a day later? Seems like the daily cron job should have kicked off and created the pages at some point? While that still doesn't fully resolve the bug, I'd just be really surprised if the pages didn't get created after the first time the job runs.

includes/class-wc-calypso-bridge-mu-events.php
```diff
@@ -0,0 +1,63 @@
...
+			self::$instance = new self();
+		}
+		return self::$instance;
+	}
+
+	/**
+	 * Constructor.
+	 */
+	private function __construct() {
+		add_action( 'woocommerce_admin_newly_installed', array( $this, 'maybe_create_wc_pages' ), 10, 2 );
```
timmyc, line 38: I thought we already tried this hook once before - and it seemed to never fire in the context of the ecomm plan?

timmyc:
So I'm trying to think of some other creative ways to trigger the page add logic you have here. The one notion that comes to mind is to perhaps piggyback on top of a flow that we know users will venture through when first setting up their store - like the OBW.

An example of how we are doing this currently can be [seen here](https://github.com/Automattic/wc-calypso-bridge/blob/master/includes/class-wc-calypso-bridge-filters.php#L38) where we filter off the CBD industry in the OBW. It wouldn't be 100% solid as users could skip the OBW entirely, but if we coupled that with the existing cron based fix, seems like we would get the pages created.

includes/class-wc-calypso-bridge-mu-events.php
moon0326: @timmyc We did, but it never got executed. It's a bit hard to explain and it's confusing, but the reason the hook did not get called was because `wc-calypso-bridge` never get a chance to add hooks due to [this line](https://github.com/Automattic/wc-calypso-bridge/blob/master/wc-calypso-bridge.php#L20).

With the change in this PR, the hook should work.

moon0326:
> So I'm trying to think of some other creative ways to trigger the page add logic you have here. The one notion that comes to mind is to perhaps piggyback on top of a flow that we know users will venture through when first setting up their store - like the OBW.
> 
> An example of how we are doing this currently can be [seen here](https://github.com/Automattic/wc-calypso-bridge/blob/master/includes/class-wc-calypso-bridge-filters.php#L38) where we filter off the CBD industry in the OBW. It wouldn't be 100% solid as users could skip the OBW entirely, but if we coupled that with the existing cron based fix, seems like we would get the pages created.

I think that's a guaranteed solution and reasonable as users who want to use the pages should go through the OBW anyway. I will update this PR with your suggestion as changes in this PR might not work. Thank you for the suggestion 👍

moon0326:
Closing this. I think a new branch is cleaner as we're just changing the action to filter.

PR #671, opened 2021-04-02
Use woocommerce_admin_onboarding_industries filter for the create page callback

moon0326:
This PR uses `woocommerce_admin_onboarding_industries` filter to check WC default pages and create them if they're missing.

## Testing Successful Page Creation

1. Delete the existing pages and options to reset the WC state.

```
delete from wp_posts where post_name in ('shop', 'cart', 'checkout', 'my-account');

delete from wp_options where option_name in ('woocommerce_task_list_tracked_completed_tasks', 'wc_pages_created_by_wc_calypso_bridge', 'woocommerce_onboarding_profile');
```

2. Navigate to WooCommerce -> Home
3. Click on the Store details task and finish the OBW. This isn't necessary for this particular test case, but we need it for the next test case. 
4. Navigate to WordPress -> Pages and confirm the shop, my account, checkout, and cart pages.
3. Confirm the `wc_pages_created_by_wc_calypso_bridge` option value is`yes`.
```
select option_value from wp_options where option_name = 'wc_pages_created_by_wc_calypso_bridge'
```

#### Case 1: wc_pages_created_by_wc_calypso_bridge
1. Delete `wc_pages_created_by_wc_calypso_bridge` option value.
```
delete from wp_options where option_name='wc_pages_created_by_wc_calypso_bridge';
```
2. Navigate to WooCommerce -> Home
3. Navigate WordPress -> Pages and observe the pages haven't updated. 

`maybe_create_wc_pages()` method should return at this [line](https://github.com/Automattic/wc-calypso-bridge/blob/e7cac2989793705165afc704ec7abad2d4fbcedb/includes/class-wc-calypso-bridge-plugins.php#L128) for this case.

#### Case 2: woocommerce_task_list_tracked_completed_tasks

1. Delete options 

```
delete from wp_options where option_name in ('woocommerce_task_list_tracked_completed_tasks', 'wc_pages_created_by_wc_calypso_bridge', 'woocommerce_onboarding_profile');
```
2. Navigate to WooCommerce -> Home
3. Navigate WordPress -> Pages and observe the pages haven't updated. 

`maybe_create_wc_pages()` method should return at this [line](https://github.com/Automattic/wc-calypso-bridge/blob/e7cac2989793705165afc704ec7abad2d4fbcedb/includes/class-wc-calypso-bridge-plugins.php#L135) for this case.

#### Case 3: Deleting one of the existing pages

1. Delete one of the existing pages and options.
```
delete from wp_options where option_name in ('woocommerce_task_list_tracked_completed_tasks', 'wc_pages_created_by_wc_calypso_bridge', 'woocommerce_onboarding_profile');

delete from wp_posts where post_name = 'shop';
```
2. Navigate to WooCommerce -> Home
3. Navigate WordPress -> Pages and confirm that the shop page hasn't been recreated. 

`maybe_create_wc_pages()` method should return at this [line](https://github.com/Automattic/wc-calypso-bridge/blob/e7cac2989793705165afc704ec7abad2d4fbcedb/includes/class-wc-calypso-bridge-plugins.php#L135) for this case.

### DIFF
--- a/includes/class-wc-calypso-bridge-plugins.php
+++ b/includes/class-wc-calypso-bridge-plugins.php
@@ -39,7 +39,7 @@ class WC_Calypso_Bridge_Plugins {
 		add_action( 'update_option_active_plugins', array( $this, 'prevent_woocommerce_deactivation' ), 10, 2 );
 		add_action( 'current_screen', array( $this, 'prevent_woocommerce_deactiation_route' ), 10, 2 );
 		add_action( 'admin_notices', array( $this, 'prevent_woocommerce_deactiation_notice' ), 10, 2 );
-		add_action( 'wc_admin_daily', array( $this, 'maybe_create_wc_pages' ), 10, 2 );
+		add_filter( 'woocommerce_admin_onboarding_industries', array( $this, 'maybe_create_wc_pages' ), 10, 2 );
 	}
 
 	/**
@@ -102,8 +102,12 @@ class WC_Calypso_Bridge_Plugins {
 
 	/**
 	 * Check WooCommerce pages (shop, cart, my-account, checkout) and create them if they don't exist.
+	 *
+	 * @param array $industries Array of industries.
+	 *
+	 * @return array
 	 */
-	public function maybe_create_wc_pages() {
+	public function maybe_create_wc_pages($industries) {
 		global $wpdb;
 
 		$post_count = $wpdb->get_var( "select count(*) from $wpdb->posts where post_name in ('shop', 'cart', 'my-account', 'checkout')" );
@@ -115,7 +119,9 @@ class WC_Calypso_Bridge_Plugins {
 			}
 			WC_Install::create_pages();
 		}
-	}
 
+		return $industries;
+	}
 }
+
 $wc_calypso_bridge_plugins = WC_Calypso_Bridge_Plugins::get_instance();
--- a/includes/class-wc-calypso-bridge-plugins.php
+++ b/includes/class-wc-calypso-bridge-plugins.php
@@ -101,24 +101,49 @@ class WC_Calypso_Bridge_Plugins {
 	}
 
 	/**
-	 * Check WooCommerce pages (shop, cart, my-account, checkout) and create them if they don't exist.
+	 * Check WooCommerce pages (shop, cart, my-account, checkout) and create them if the following conditions are met.
+	 *
+	 * 1. This is the first time running this method.
+	 * 2. User has not finished Store details task.
+	 * 3. Shop, cart, my-account, and checkout pages do not exist.
 	 *
 	 * @param array $industries Array of industries.
 	 *
 	 * @return array
 	 */
-	public function maybe_create_wc_pages($industries) {
+	public function maybe_create_wc_pages( $industries ) {
 		global $wpdb;
+
+		$option_name = 'wc_pages_created_by_wc_calypso_bridge';
+
+		// Abort if we have attempted to create the pages already.
+		if ( 'yes' === get_option( $option_name, 'no' ) ) {
+			return $industries;
+		}
+
+		// Abort if the user has completed store details task already.
+		$completed_tasks = get_option( 'woocommerce_task_list_tracked_completed_tasks', [] );
+		if ( in_array( 'store_details', $completed_tasks ) ) {
+			return $industries;
+		}
 
 		$post_count = $wpdb->get_var( "select count(*) from $wpdb->posts where post_name in ('shop', 'cart', 'my-account', 'checkout')" );
 
-		if ( 4 !== (int) $post_count ) {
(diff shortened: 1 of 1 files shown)


moon0326:
@timmyc Could you review this change when you get a chance, please 🙇? This change uses the filter you suggested in the previous PR. I think this is going to work finally.

timmyc:
> It looks like we fire woocommerce_admin_onboarding_industries when we navigate to WooCommerce -> Home. I think it's okay for the performance as this change applies to users with ECOM plan on wordpress.com and count query is generally super quick. I think this is better than not having the default pages.

Looked into this, and it seems like this is correct since apply_filters is called when the [setup checklist is still shown](https://github.com/woocommerce/woocommerce-admin/blob/2de2adf3672534412b531996f79cb2b14353a362/src/Features/Onboarding.php#L657-L661) on the home screen. So yeah a little perf hit, but once the setup checklist is completed, or hidden, this code path should no longer eval.

moon0326:
@timmyc Yup, I think this is safe to push out even with a little perf hit.

timmyc:
So based off our learnings from the user that had an issue with this similar logic... wondering if we should consider the following:

- Instead of making the check see that all four default pages exist, maybe make that more like if any one of the default pages exist we do not recreate them ( for possible scenarios where users perhaps delete the my account page since they don't offer registration ).
- After running the logic to setup pages, perhaps we should set a flag in the site options that the job has run and the default pages have been created. Then on subsequent runs of the logic, if that option exists, exit early.

moon0326:
> So based off our learnings from the user that had an issue with this similar logic... wondering if we should consider the following:
> 
> * Instead of making the check see that all four default pages exist, maybe make that more like if any one of the default pages exist we do not recreate them ( for possible scenarios where users perhaps delete the my account page since they don't offer registration ).
> * After running the logic to setup pages, perhaps we should set a flag in the site options that the job has run and the default pages have been created. Then on subsequent runs of the logic, if that option exists, exit early.

That's a really good suggestions 👍 

I'm gonna look into the problem later today and make changes.

moon0326:
@timmyc Hey Timmy, I've updated the code with your suggestions. I've also added a line to check `store_details` task completion as a precaution for the existing stores.

includes/class-wc-calypso-bridge-plugins.php
```diff
@@ -98,5 +99,56 @@ public function remove_woocommerce_deactivation_link( $actions, $plugin_file ) {
 		}
 		return $actions;
 	}
+
+
+	/**
```
timmyc, line 104: Thanks for the detailed doc block here Moon

includes/class-wc-calypso-bridge-plugins.php
```diff
@@ -98,5 +99,56 @@ public function remove_woocommerce_deactivation_link( $actions, $plugin_file ) {
...
+
+		// Abort if the user has completed store details task already.
+		$completed_tasks = get_option( 'woocommerce_task_list_tracked_completed_tasks', [] );
+		if ( in_array( 'store_details', $completed_tasks ) ) {
+			return $industries;
+		}
+
+		$post_count = $wpdb->get_var( "select count(*) from $wpdb->posts where post_name in ('shop', 'cart', 'my-account', 'checkout')" );
+
+		// Abort if we find any existing pages.
```
timmyc, line 133: Nice!

timmyc (commented):
Code is looking good Moon - can we update the PR description with some new test steps maybe to test out the different scenarios here - and/or maybe we should also write some unit test coverage around `maybe_create_wc_pages` to help ensure the various cases are covered correctly.

includes/class-wc-calypso-bridge-plugins.php
```diff
@@ -98,5 +99,56 @@ public function remove_woocommerce_deactivation_link( $actions, $plugin_file ) {
...
+		if ( 0 !== (int) $post_count ) {
+			return $industries;
+		}
+
+		// Reset the woocommerce_*_page_id options.
+		// This is needed as woocommerce_*_page_id options have incorrect values on a fresh installation
+		// for an ecom plan. WC_Install:create_pages() might not create all the
+		// required pages without resetting these options first.
+		foreach ( [ 'shop', 'cart', 'myaccount', 'checkout' ] as $page ) {
```
timmyc, line 142: TIL about this option mapping, great catch.

moon0326:
> Code is looking good Moon - can we update the PR description with some new test steps maybe to test out the different scenarios here - and/or maybe we should also write some unit test coverage around `maybe_create_wc_pages` to help ensure the various cases are covered correctly.

I will update the PR with detailed test instructions shortly. Thank you for the review 👍

timmyc (approved):
Thanks for the detailed test instructions Moon. All cases are passing as outlined for me. Also did the case of deleting one page ( My Account ) and ensured no pages are created.

PR #672, opened 2021-04-05
Remove wc_admin_daily hook

moon0326:
We received a report saying that our last merge to create WC pages is causing a problem for a user (https://a8c.slack.com/archives/C7YPW6K40/p1617638352126100)

I'm removing the change until we find the root cause.


timmyc (approved):
LGTM

PR #674, opened 2021-04-06
Add deployment steps

moon0326:
This PR adds deployment steps to README. I found myself keep looking for a link to the deployment steps in P2 :) I figured it would be nice to have it in README. I removed mention of wpcomsh as that is in a private repository.


ilyasfoo (approved):
I've crossed-check this with our P2, LGTM! 🚢

Issue #652, opened 2021-02-22
Hide free extensions promotion


pmcpinto:
**Describe the bug**
Some of the free extensions that we promote (Facebook and Mailchimp) are installed by default on the eComm plan. We should hide the extensions promotions in the alternative flows that we created: [#4579](https://github.com/woocommerce/woocommerce-admin/issues/4579) and [#5287](https://github.com/woocommerce/woocommerce-admin/issues/5287). We should hide the free extensions promotion in these flows 

**To Reproduce**

1. Create a new eComm plan
2. Select US as the country in the store detail step
3. In the Industry step select one of the following industries: Other, Food & Drinks, Fashion, Apparel, Accessories or Health & Beauty
4. See the free extensions promotion in the Business details step

timmyc:
@pmcpinto given the extensions step is such a moving target, would it make more sense to remove it entirely from the eCommerce plan? We can not currently do that, but after https://github.com/woocommerce/woocommerce-admin/issues/6352 lands we will be able to do so.

psealock:
I'm pretty sure this is handled by https://github.com/Automattic/wc-calypso-bridge/pull/675 which hides the free extensions tab in Business Details step.

Issue #635, opened 2021-01-23
Missing OBW Style Tweaks


timmyc:
During the great calypsoify removal of '20 - it appears some lines of css might not have made it over the fence into the pastures of _New Nav Nirvana_. @pmcpinto noticed the two following items:

```
Product types step - the prices toggle and information
Business details step - the free extensions promotion and information
```

I think those tweaks were originally shipped in the following PRs #610 #547 so we should be able to reimpliment/rehide pretty quickly.

timmyc:
This should be a quick win < 1 estimate.

PR #675, opened 2021-04-08
Add missing OBW Tweaks

psealock:
Fixes https://github.com/Automattic/wc-calypso-bridge/issues/635
Fixes https://github.com/Automattic/wc-calypso-bridge/issues/652

Some styles needed adjustment after de-Calypsoification

## Product Types Before

Screen Shot 2021-04-08 at 4 13 47 PM

## Product Types After

Screen Shot 2021-04-09 at 11 55 49 AM

## Business Details Before

Screen Shot 2021-04-08 at 4 21 35 PM

## Business Details After

Screen Shot 2021-04-08 at 4 31 30 PM

## Testing Instructions

1. Enable the OBW
2. See Product Types step and Business Details
3. Ensure the right things are hidden and there are no regressions

### DIFF
--- a/assets/css/ecommerce.css
+++ b/assets/css/ecommerce.css
@@ -1,11 +1,3 @@
-/* Hide business extensions in Profiler > Business Details */
-.woocommerce-profile-wizard__body
-	.woocommerce-profile-wizard__container
-	div
-	.woocommerce-profile-wizard__benefit {
-	display: none;
-}
-
 /* Hides Theme Tabs ( All | Free | Paid ) since we only show installed themes */
 .woocommerce-profile-wizard__body
 	.woocommerce-profile-wizard__themes-tab-panel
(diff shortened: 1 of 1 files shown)


psealock:
cc @timmyc to confirm the Business Details step since that has changed since the original issue was raised. Just double checking that removing the entire Free Features section is the intention.

psealock:
I'm just on my way out and realized this bit should also be removed because the context isn't there anymore. I'll fix tomorrow

Screen Shot 2021-04-08 at 4 13 33 PM

timmyc:
> cc @timmyc to confirm the Business Details step since that has changed since the original issue was raised. Just double checking that removing the entire Free Features section is the intention.

Great point Paul, thinking we might want to show the new tabbed biz features step at all, but @pmcpinto might be better suited to make that call.

psealock:
I fixed up the Product Types step to remove the footnote about billing because it didn't make sense without the billing context. This is ready for review

pmcpinto:
> Great point Paul, thinking we might want to show the new tabbed biz features step at all, but @pmcpinto might be better suited to make that call.

We can hide the business features tab. Thanks Paul!

ilyasfoo (approved):
LGTM and tests well! :shipit:

PR #680, opened 2021-05-27
Fix nav unification causing header overflow

ilyasfoo:
Fixes https://github.com/woocommerce/woocommerce-admin/issues/7035.

Nav unification caused WooCommerce header to overflow due to the width rule hardcoded with assumed navigation width. This PR fixes the issue for both business and e-commerce plan user.

### Testing instructions

1. In order to enable nav unification, make sure to have `wpcomsh` and Jetpack installed and connected.
2. If you're experiencing fatal error due to `wpcomsh` conflict, try using my [fork of the bridge helper](https://gist.github.com/ilyasfoo/3a898d3aca8c6a27acc851a28636332d).
2. Make sure new navigation feature is turned off.
3. Navigate to any WooCommerce page, observe a normal header.

### Screenshot:
Before:

image

After:

image


store-on-wpcom/assets/css/admin/nav-unification.css
```diff
@@ -0,0 +1,4 @@
+/* Fix header overflow https://github.com/woocommerce/woocommerce-admin/issues/7035 */
+.jetpack-connected .woocommerce-layout__header {
```
ilyasfoo, line 2: `.jetpack-connected` rule is used to override the `.woocommerce-layout__header` which style is enqueued after ours. The assumption is also correct, i.e: without jetpack connected, nav unification will not be enabled.

moon0326 (approved):
Nice work on this!! 👍  This LGTM and tested well 🎉 🚢

PR #682, opened 2021-06-03
Fix nav unification header overflow breaking mobile views

ilyasfoo:
Fixes https://github.com/woocommerce/woocommerce-admin/issues/7035 again.

This PR is a follow-up to https://github.com/Automattic/wc-calypso-bridge/pull/680, which broke the header for screen width < `961px`.

### Testing instructions

1. In order to enable nav unification, make sure to have `wpcomsh` and Jetpack installed and connected.
2. If you're experiencing fatal error due to `wpcomsh` conflict, try using my [fork of the bridge helper](https://gist.github.com/ilyasfoo/3a898d3aca8c6a27acc851a28636332d).
2. Make sure new navigation feature is turned off.
3. Navigate to any WooCommerce page, observe a normal header.

### Screenshots

**Before:**

image

**After:**

image

PR #685, opened 2021-06-17
Remove Canada Post notice suppression and header fix

### LINKED ISSUE #684: Users unable to connect to Canada Post
As reported in 4063940-zd-woothemes and p1623769959450600-slack-C7U3Y3VMY, ecommerce users are unable to connect to Canada Post account because the way to connect is via a notice that appears in wp-admin, but it's being removed [here](https://github.com/Automattic/wc-calypso-bridge/commit/53a7a6cb021e4c83b44fd3c2fc2f46050a85390e#diff-574140c5ebef5f5a93893e45669a487fc8f736639e6eb3b19f0869e1735c95beR81).

I don't have enough context to know what's the best course of action, more research is needed.

ilyasfoo:
Fixes #684.

This PR does the following:

- Removes the notice suppression for Canada Post since the extension requires a connection via woocommerce.com that can only be accessed via notice.
- Fix for header UI bug for e-commerce users. This was introduced in https://github.com/Automattic/wc-calypso-bridge/pull/682 where some e-commerce users with WooCommerce Navigation feature turned on but at the same time has nav unification flag enabled. The fix adds another check to make sure that Navigation feature is turned off before apply the original CSS fix.

## Testing instructions

1. Set the store's address to somewhere in Canada.
2. Install & activate `WooCommerce Canada Post Shipping` extension.
3. Observe the  "Connect Your Canada Post" notice in plugins page.

### Canada Post notice:

image

### Header UI:

Before
image

After
image

### DIFF
--- a/class-wc-calypso-bridge.php
+++ b/class-wc-calypso-bridge.php
@@ -173,7 +173,7 @@ class WC_Calypso_Bridge {
 			add_filter( 'admin_footer_text', array( $this, 'update_woocommerce_footer' ) );
 			add_action( 'admin_enqueue_scripts', array( $this, 'add_ecommerce_plan_styles' ) );
 
-			// Nav unification fixes.
+		// Nav unification fixes.
 		if ( function_exists( 'wpcomsh_activate_nav_unification' )
 			&& wpcomsh_activate_nav_unification( false )
 			&& ! Loader::is_feature_enabled( 'navigation' ) ) {
--- a/class-wc-calypso-bridge.php
+++ b/class-wc-calypso-bridge.php
@@ -174,7 +174,7 @@ class WC_Calypso_Bridge {
 			add_action( 'admin_enqueue_scripts', array( $this, 'add_ecommerce_plan_styles' ) );
 
 			// Nav unification fixes.
-    if ( function_exists( 'wpcomsh_activate_nav_unification' )
+		if ( function_exists( 'wpcomsh_activate_nav_unification' )
 			&& wpcomsh_activate_nav_unification( false )
 			&& ! Loader::is_feature_enabled( 'navigation' ) ) {
 				add_action( 'admin_enqueue_scripts', array( $this, 'add_nav_unification_styles' ) );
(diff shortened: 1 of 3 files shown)


ilyasfoo:
I've found out that Canada Post original setup task was removed in https://github.com/Automattic/wc-calypso-bridge/pull/561/files#diff-f550a8568c500f26c3c463f03901ca0c4c422f6922ffb68fe1a1c184e60b1b7eL401-L411. I've also looked at other tasks in the list and I think it's safe to say that no other extensions are affected by notice suppression since most of them would have their respective task built-in wc-admin.

class-wc-calypso-bridge.php
```diff
@@ -172,7 +174,9 @@ public function load_ui_elements() {
 			add_action( 'admin_enqueue_scripts', array( $this, 'add_ecommerce_plan_styles' ) );
 
 			// Nav unification fixes.
-			if ( function_exists( 'wpcomsh_activate_nav_unification' ) && wpcomsh_activate_nav_unification( false ) ) {
+		if ( function_exists( 'wpcomsh_activate_nav_unification' )
```
adrianduffell, line 177: Is the indentation off here?

store-on-wpcom/inc/wc-calypso-bridge-nav-unification-fixes.php
```diff
@@ -24,7 +26,9 @@
  */
 function load_ui_elements() {
   if ( is_wc_calypso_bridge_page() ) {
-    if ( function_exists( 'wpcomsh_activate_nav_unification' ) && wpcomsh_activate_nav_unification( false ) ) {
+    if ( function_exists( 'wpcomsh_activate_nav_unification' )
+      && wpcomsh_activate_nav_unification( false )
```
adrianduffell, line 30: Also looks like spaces used for indentation here.

adrianduffell (approved):
@ilyasfoo This tested well and code looks good!

I noticed some indentation might be off but pre-approving 🚀

class-wc-calypso-bridge.php
ilyasfoo: Thanks for catching this! I've fixed indentation in 692e24c16cee7f99ca392375de45c9877dd0b344

PR #699, opened 2021-07-05
Add TypeScript Support

adrianduffell:
Adds TypeScript support, using a similar setup to WooCommerce Admin.

### Testing Instructions.

- Run `npm start`
- Change an existing `.js` file to `.tsx`, for example `src/payments-welcome/index.tsx`
- See that it builds ok
- Make an intentional type error, for example `const PaymentsWelcomePage = () : number => (`
- See that a build error occurs.


samueljseay (approved):
Great stuff! I've tested and it works well. 🚀

Issue #690, opened 2021-06-29
Automatically redirect merchants to the KYC flow after they install the extension


samueljseay:
The WC Pay welcome screen has a "Get Started" button.

![image](https://user-images.githubusercontent.com/9312929/124141299-e32c4180-dabb-11eb-963b-c0f9c0a7f8e5.png)

This button should install WCPay while showing loading/progress, then redirect to `wcpaySettings.connectUrl` global var after installation.

samueljseay:
I'm estimating this as a 2, because I think based on discussion that once everything else is in place this will be a relatively small task.

@adrianduffell @ilyasfoo could you confirm?

adrianduffell:
Yeah 2 sounds good to me 👍

ilyasfoo:
Adding a detail we had in sprint planning call, we can try to use either the wp.data store `wc/admin/plugins` (the most optimal) or directly to API `/wc-admin/plugins/install`. Business step in OBW would be a good reference

ilyasfoo:
@pmcpinto The existing track in WCPay used to connect is `wcadmin_wcpay_connect_account_clicked`. Do you want to have a different track for when users connect via the wcpay in core experiment? Or maybe an additional prop to differentiate?

pmcpinto:
@ilyasfoo I think that probably having a different event or `wcadmin_page_view` `path` prop is enough. @oneillo what do you think?

oneillo:
@pmcpinto @ilyasfoo - I like the idea of using a new `wcadmin_page_view` prop to distinguish when a user sees the version of this screen in the experiment and keeping the existing track for connecting WCPay the same. I think it better reflects the situation; that a user is seeing this screen that they may not have seen previously, and if they connect, they are taking the same action they would have taken under normal circumstances.

ilyasfoo:
@pmcpinto @oneillo Just to make sure I understood this correctly, do we want to add a new prop `wcadmin_page_view` with the `path` as its value into `wcadmin_wcpay_connect_account_clicked` track event?

Also, note that this new prop will only exist when users connect via this project. The original track in WCPay will not contain this new prop.

pmcpinto:
> @pmcpinto @oneillo Just to make sure I understood this correctly, do we want to add a new prop wcadmin_page_view with the path as its value into wcadmin_wcpay_connect_account_clicked track event?
> Also, note that this new prop will only exist when users connect via this project. The original track in WCPay will not contain this new prop.

[...]

PR #700, opened 2021-07-06
Install WCPay and redirect merchant KYC flow

ilyasfoo:
This PR adds the following:
- Install and activate WCPay plugin when click on "Get started"
- Redirect to KYC after successful activation
- Remove redundant pageview track `CONNECT_ACCOUNT_VIEW` (wcadmin already tracks `wcadmin_page_view`)
- Adds local state for displaying errors
- Change `ConnectPageOnboarding` to use props for better testability

What it doesn't do:
- Detect if WCPay plugin is already installed (I think this is better handled at disabling menu instead)

Todo in a follow-up:
- Potentially add a prop to a track to differentiate between this project's connect and WCPay original connect. Ongoing discussion [here](https://github.com/Automattic/wc-calypso-bridge/issues/690#issuecomment-874401258).

### Testing instructions
- Make sure WCPay is not installed.
- Go to the new welcome page, click on "Get started".
- Observe the loading animation and wait until it finishes, it should automatically redirect to KYC.
- Don't need to complete KYC, instead look at the plugins page and notice that WooCommerce Payments plugin is installed & activated.

### DIFF
--- a/class-wc-calypso-bridge.php
+++ b/class-wc-calypso-bridge.php
@@ -157,7 +157,8 @@ class WC_Calypso_Bridge {
 		wp_add_inline_script( 'wc-calypso-bridge', 'window.wcCalypsoBridge = ' . wp_json_encode(
 			array(
 				'isWooPage' => $is_woo_page,
-				'homeUrl'   => get_home_url()
+				'homeUrl'   => get_home_url(),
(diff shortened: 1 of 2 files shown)


adrianduffell (approved):
@ilyasfoo Nice work, this tested well and LGTM! 🚀 

> Detect if WCPay plugin is already installed (I think this is better handled at disabling menu instead)

Good call. I'll add some details about this to the menu item issue.

PR #705, opened 2021-07-07
Call activate promo REST API 

### LINKED ISSUE #689: Reuse one of the financial incentives implemented for the new existing store acquisition experiment
Stores that set up WC Pay should have the "no transaction fees for 3 months" promo applied to their account.

moon0326:
Fixes #689 

This PR calls activate promo REST API to activate the WC Pay incentive promotion.

### Testing instructions

1. Make sure you don't have `wcpay-promo-2021-6-incentive-2` note in your `wp_wc_admin_notes` table. If you have it already, make sure the status is `unactioned` 
2. Navigate to the welcome page.
3. Click "Get started" button. You should be redirected. 
4. Open `wp_wc_admin_notes` table and confirm that the `wcpay-promo-2021-6-incentive-2` status is now `actioned` 

You can also finish the KYC and confirm your account has the promotion applied.

### Screenshots

Screen Shot 2021-07-07 at 2 44 00 PM

### DIFF
--- a/includes/payments/class-wc-payments-controller.php
+++ b/includes/payments/class-wc-payments-controller.php
@@ -27,7 +27,7 @@ class WC_Payments_Controller extends WC_REST_Controller {
 	 *
 	 * @var string
 	 */
-	protected $namespace = 'wc-calypso-brdige/v1';
+	protected $namespace = 'wc-calypso-bridge/v1';
 
 	/**
 	 * Route base.
@@ -56,8 +56,7 @@ class WC_Payments_Controller extends WC_REST_Controller {
 	 * Override this method if custom permissions required.
 	 */
 	public function check_permission() {
-		return true;
-//		return current_user_can( 'manage_woocommerce' );
+		return current_user_can( 'manage_woocommerce' );
 	}
 
 	/**
@@ -88,7 +87,12 @@ class WC_Payments_Controller extends WC_REST_Controller {
 			$note->set_status( Note::E_WC_ADMIN_NOTE_ACTIONED );
 			$data_store->create( $note );
 		}
-		return true;
+
+		return rest_ensure_response(
+			array(
+				'success' => true
+			)
+		);
 	}
 
 }
--- a/includes/payments/class-wc-payments-controller.php
+++ b/includes/payments/class-wc-payments-controller.php
@@ -27,7 +27,7 @@ class WC_Payments_Controller extends WC_REST_Controller {
 	 *
 	 * @var string
 	 */
-	protected $namespace = 'wc/v3';
+	protected $namespace = 'wc-calypso-brdige/v1';
 
 	/**
 	 * Route base.
@@ -40,9 +40,9 @@ class WC_Payments_Controller extends WC_REST_Controller {
 	 * Register routes.
 	 */
 	public function register_routes() {
-		register_rest_route( $this->namespace, '/' . $this->rest_base . '/post-install', array(
+		register_rest_route( $this->namespace, '/' . $this->rest_base . '/activate-promo', array(
 			array(
-				'methods'             => WP_REST_Server::READABLE,
+				'methods'             => 'POST',
 				'callback'            => array( $this, 'activate_promo_note' ),
 				'permission_callback' => array( $this, 'check_permission' ),
 			),
@@ -56,7 +56,8 @@ class WC_Payments_Controller extends WC_REST_Controller {
 	 * Override this method if custom permissions required.
 	 */
 	public function check_permission() {
-		return current_user_can( 'manage_woocommerce' );
+		return true;
+//		return current_user_can( 'manage_woocommerce' );
 	}
 
 	/**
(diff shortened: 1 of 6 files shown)


includes/payments/class-wc-payments-controller.php
```diff
@@ -0,0 +1,98 @@
...
+ * @package WooCommerce/API
+ */
+class WC_Payments_Controller extends WC_REST_Controller {
+
+	/**
+	 * Endpoint namespace.
+	 *
+	 * @var string
+	 */
+	protected $namespace = 'wc-calypso-bridge/v1';
```
moon0326, line 30: I renamed the namespace from `wc/v3` to `wc-calypso-bridge/v1` as we're in wc-calypso-bridge, but we can also keep `wc/v3`.

includes/payments/class-wc-payments-controller.php
```diff
@@ -0,0 +1,98 @@
...
+	 */
+	protected $rest_base = 'payments';
+
+	/**
+	 * Register routes.
+	 */
+	public function register_routes() {
+		register_rest_route( $this->namespace, '/' . $this->rest_base . '/activate-promo', array(
+			array(
+				'methods'             => 'POST',
```
moon0326, line 45: I changed the API name from `post-install` to `activate-promo` and methods to `POST` as well.

includes/payments/class-wc-payments-controller.php
```diff
@@ -0,0 +1,98 @@
...
+	public function activate_promo_note() {
+		$promo_name = 'wcpay-promo-2021-6-incentive-2';
+		$data_store = WC_Data_Store::load( 'admin-note' );
+		$add_where_clause = function( $where_clause ) use ( $promo_name ) {
+			return $where_clause . " AND name = '$promo_name'";
+		};
+
+		add_filter( 'woocommerce_note_where_clauses', $add_where_clause );
+		$notes = $data_store->get_notes();
+		remove_filter( 'woocommerce_note_where_clauses', $add_where_clause );
```
adrianduffell, line 75: I _think_ `remove_filter` is unable to remove anonymous functions like `$add_where_clause `. A bit more info [here](https://stackoverflow.com/questions/31240432/anonymous-functions-in-wordpress-hooks).

If that's the case, I think the where clause would continue to run on other notes 😬 

Would it be better to use a class method for the filter?

adrianduffell (commented):
@moon0326 This is looking good, I just left a question about the `woocommerce_note_where_clauses` filter.

includes/payments/class-wc-payments-controller.php
becdetat: I think that SO answer is talking about the case where `remove_filter` is being called in a different scope to `add_filter`, where the anonymous function  instance isn't available. Since this is calling `remove_filter` with the same instance, IMO this should work. I am a rookie however so don't trust me.

moon0326: Thank you @adrianduffell 👍 This was very interesting! I wrote the following test locally to confirm that it actually removes the filter. @becdetat I think you're right that it can be removed as long as the function can be referenced back. 

```

public function test_filter_gets_removed() {
	global $wp_filter;
	$promo_name = 'wcpay-promo-2021-6-incentive-2';
	$data_store = WC_Data_Store::load( 'admin-note' );
	$add_where_clause = function( $where_clause ) use ( $promo_name ) {
		return $where_clause . " AND name = '$promo_name'";
	};

	add_filter( 'woocommerce_note_where_clauses', $add_where_clause );
	$data_store->get_notes();
	$this->assertArrayHasKey('woocommerce_note_where_clauses', $wp_filter);

	$wheres_with_filter = $data_store->get_notes_where_clauses();
	$this->assertEquals(' AND is_deleted = 0 AND name = \'wcpay-promo-2021-6-incentive-2\'', $wheres_with_filter);

	remove_filter( 'woocommerce_note_where_clauses', $add_where_clause );

	$wheres_without_filter = $data_store->get_notes_where_clauses();
	$this->assertEquals(' AND is_deleted = 0', $wheres_without_filter);

	$this->assertArrayNotHasKey('woocommerce_note_where_clauses', $wp_filter);
}
```

src/payments-welcome/index.tsx
```diff
@@ -102,22 +103,45 @@ const ConnectPageOnboarding = ({
...
+			const installAndActivateResponse = await installAndActivatePlugins([
+				'woocommerce-payments',
+			]);
+
+			if (installAndActivateResponse?.success) {
+				activatePromo();
+			} else {
+				renderErrorMessage(installAndActivateResponse.message);
+			}			
+		} catch (e: any) {
```
ilyasfoo, line 143: Thanks for adding a catch-all here ❤️

ilyasfoo (approved):
Changes look perfect, couldn't have done it any better myself. It even works when I bailed the Jetpack connection the first time & continued via WCPay original interface.

Tested well, ship it! :shipit:

includes/payments/class-wc-payments-controller.php
adrianduffell: Awesome stuff, thanks for investigating it!

Issue #707, opened 2021-07-09
Add payments setting


adrianduffell:
Users should be able to re-enable the Payments menu item after dismissing it with the "No, thanks" button.

A new WooCommerce setting needs to be added  for this called "Payments". We perhaps also need to switch our existing usage of the `wc_calypso_bridge_payments_dismissed` option to use this setting instead?

The exit survey states:

> This will remove WooCommerce Payments from the navigation. In order to enable it again, go to WooCommerce Settings > Payments

PR #708, opened 2021-07-12
Re-enable WCPay menu in payments settings

ilyasfoo:
Fixes https://github.com/automattic/wc-calypso-bridge/issues/707

This PR adds an admin notice in Payments settings page that contains a link to the welcome page with a param to trigger re-enable of the menu. The re-enable is done via `wc-admin/options` API. It also adds a URL utility function that's copied from wc-admin.

I used an admin note because I can't think of any better way yet, suggestions are welcome. Note the texts used are copied from existing strings available. I expect to change this if required.

image

### Testing instructions

1. Go to the WCPay welcome page, click on "No thanks" and remove the menu.
2. Go to Settings > Payments.
3. Observe an admin note is added per screenshot attached.
4. Click on "Learn more".
5. It should navigate back to the welcome page.
6. Refresh the page to observe that the WCPay menu is re-added on admin menu.


ilyasfoo:
Hi @pmcpinto @elizaan36, need your help to check if this change is okay. Basically, it's an option to re-enable WCPay when the user already clicked on "No thanks".  I need to check:
1. If its okay to make it as an admin note instead of a checkbox since no existing checkbox is in payments page (I might try to reposition this note to below the payments table since we feel that it's a bit obstructive).
2. The text content of the note and action button.

elizaan36:
hey @ilyasfoo, there's no need for an admin note. Since we recently introduced the recommended payments extensions card, WooCommerce Payments is always available in the core Payments settings screen, even when it's not installed.

ilyasfoo:
Thanks for the feedback @elizaan36! Does that mean we don't have to add anything in the Settings screen if the user clicked on "No thanks"? I'm curious because the exit survey states:

>This will remove WooCommerce Payments from the navigation. In order to enable it again, go to WooCommerce Settings > Payments

Let me know you can confirm this, thanks!

adrianduffell:
@ilyasfoo I think @elizaan36 is referring to this on the payments settings screen  https://github.com/woocommerce/woocommerce-admin/pull/6760. Apologies for the bad direction in the issue, TIL about this feature.

However I'm confused because I don't see it appear locally with the bridge. I thought maybe it's  because `woocommerce_allow_marketplace_suggestions` is disabled but removing that line seemed to have no affect. Any ideas?

https://github.com/Automattic/wc-calypso-bridge/blob/c184d9be5ed9ffb3f79a640e2b54fd8daf9ec6f0/includes/class-wc-calypso-bridge-hide-alerts.php#L41

ilyasfoo:
@adrianduffell Hmm but still I feel the wording isn't as accurate in the exit survey. Maybe something like "You can still enable WooCommerce payments via the settings page" would be better IMO. Though also caveat users will not get the discounted rate if they enable via that and did not perform action on inbox.

The recommended card actually appears a bit later at the bottom of the page. You'd need to wait a few seconds. I did not change anything for it to appear in my local with bridge.

image

adrianduffell:
@ilyasfoo Ah right, I still don't see it after waiting but I guess it's just something to do with my local set up. Sorry for the confusion 😅 

I agree the survey intro message is not quite accurate. From the current survey message wording, I'd expect to be able to re-enable the menu item and see the screen with the FAQs again. However the button on this screen will proceed with WC Pay installation directly.

@elizaan36  What do you think of Ilyas’ proposed wording above? 

> "You can still enable WooCommerce payments via the settings page"

Also, is it ok, as Ilyas pointed out, that enabling payments from the settings page won't apply the promotion?

elizaan36:
It's fine that enabling the promotion from settings won't apply the promotion since this will be a very small edge case of users who dismiss out and then go to reenable in settings.

Ilyas' wording sounds perfect but maybe we can add something about that - 

> This will remove WooCommerce Payments from the navigation. You can enable it again in WooCommerce Settings > Payments, however the promotion will not apply.

ilyasfoo:
Alright, I've changed the text.

image

adrianduffell (approved):
LGTM! Thanks for improving the wording @ilyasfoo @elizaan36  🚀

PR #710, opened 2021-07-14
Add css fix for CardBody bug

moon0326:
Fixes p1626286300278800-slack-C9K5T0XSP

Please refer to the slack thread for more details.

A little background: 

1. Gutenberg updated `Card` component with a new [isSCrollable](https://github.com/WordPress/gutenberg/blob/trunk/packages/components/src/card/card-body/hook.js#L26) props. The prop's default value is `true`. This causes a visual glitch with `CardBody` component. 

Screen Shot 2021-07-14 at 4 33 01 PM

2. We can either fix this issue in Gutenberg or WC Admin, but both require a code push and users have to update one of the plugins.

This PR fixes the problem for wordpress.com users while we're working on a fix in WC Admin.

### DIFF
--- a/assets/css/ecommerce.css
+++ b/assets/css/ecommerce.css
@@ -20,3 +20,7 @@
 	.components-tab-panel__tabs {
 	display: none;
 }
+
+.woocommerce-layout__main .components-scrollable {
+	overflow: unset;
+}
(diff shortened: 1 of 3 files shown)


ilyasfoo (approved):
I tested this on my WPCOM site, works great! Thanks, @moon0326  🚢

ciampo:
Hey @moon0326 , just a reminder that Gutenberg 11.2.0 has been released (as a standalone plugin, and on both simple and atomic sites), which means that you should be able to revert this hotfix in case you haven't done it already!

PR #715, opened 2021-07-16
Fix payments not working for business plan and move things around

ilyasfoo:
This PR aims to fix the following problems:

1. WCPay menu did not show up for business plan users. This is because our frontend tooling was loaded in `class-wc-calypso-bridge.php`, which actually is only called for Ecommerce plan users.
2. Duplicated codes were made to handle business and ecom users, i.e: nav unification fixes.

This PR would bridge (🥁 ba-dum-tss) shared functionalities between business and ecom plan by sharing a single class. This class would now handle our frontend tooling registration and nav unification fixes.

Apart from that, this PR also moves WCPay-related script registration to `includes/class-wc-calypso-bridge-payments.php`, which seem to be more appropriate.

### Testing instructions

1. Load up this branch and observe everything still works regarding related features.
2. Optionally, for a better coverage please follow the tutorial in p90Yrv-2j7-p2 to test in WPCOM production. If possible, both business and ecommerce plan site is tested.


moon0326 (changes requested):
Thank you @ilyasfoo for this PR 🎉 It looks good and tested well on prod. One thing I noticed is that the WC Pay icon doesn't render its icon unless you're on the welcome page because `wc-payments.css` is not loaded. The icon looks good once you're on the new welcome page.

Could we load `wp-payments.css` globally for both eCom and business?

class-wc-calypso-bridge-shared.php
```diff
@@ -0,0 +1,140 @@
...
+
+    // @todo This should rely on the navigation screens instead.
+    $connect_files = glob( dirname( __FILE__ ) . '/includes/connect/*.php' );
+    foreach ( $connect_files as $connect_file ) {
+      include_once $connect_file;
+    }
+
+		if ( is_wc_calypso_bridge_page() ) {
+		// Nav unification fixes.
+		if ( function_exists( 'wpcomsh_activate_nav_unification' )
```
moon0326, line 124: I think we have formatting issue here

moon0326 (approved):

class-wc-calypso-bridge-shared.php
ilyasfoo: Fixed formatting in ceaf3e3

PR #717, opened 2021-07-20
Fix/716 wc pay menu icon and notification

### LINKED ISSUE #716: Fix WC Pay menu icon and notification
I've found two issues with the WC Paymenu while testing on wp.com.

1. The WC Pay menu icon is not displayed correctly on "My Home" menu due to missing CSS. 
2. The notification badge isn't displayed.
3. I also noticed that we render two WC Pay menu when the navigation feature is turned off.

moon0326:
Fixes #716 

This PR fixes the following problems.

1. The WC Pay menu icon is not displayed correctly on "My Home" page due to missing CSS.

I've added base64 image, but kept the CSS as well. It was a bit quirky. When I pass `icon` option to `wc_admin_register_page`, it was removed when I go to one of the WC pages. Keeping CSS can fix this problem.

2. The notification badge isn't displayed.

It turned out that the dashboard page expects a bit different notification format than what we used.

New format: `<span class="wcpay-menu-badge awaiting-mod count-1"><span class="plugin-count">1</span></span>`. Notice we have a nested span.

3. It also fixes duplicated WC Pay menu on "My Home" page.

**Note**: You might see overlapped WC Pay menu icon if you test it locally due to having both base64 image and CSS. I haven't found a way to workaround it yet, but I think it's acceptable since this is going to be on .com only.

### Testing instructions

Follow the tutorial in p90Yrv-2j7-p2 to test in WPCOM production. Upload this branch to both business and e-commerce sites and test both.

Confirm you see the correct menu icon and notification badge on "My Home" (dashboard) and WooCommerce -> Home or


includes/class-wc-calypso-bridge-payments.php
```diff
@@ -105,21 +109,23 @@ public function register_payments_welcome_page() {
...
-		add_menu_page(
-			__( 'Payments', 'wc-calypso-bridge' ),
-			__( 'Payments', 'wc-calypso-bridge' ),
-			'view_woocommerce_reports',
-			'admin.php?page=wc-admin&path=/payments-welcome',
-			null,
-			null,
-			'55.7' // After WooCommerce & Product menu items.
-		);
+		if ( 'yes' === get_option('woocommerce_navigation_enabled', 'no') ) {
```
ilyasfoo, line 112: Nice catch!

ilyasfoo (changes requested):
Thanks for making this fix @moon0326! I think the SVG icon is flipped horizontally since the dollar sign looks reversed 😆 

This worked well in Chrome, Firefox and Safari. However in Brave browser (on the wpcomstaging site, not calypso), it wierdly becomes like this:

image

I notice that it does not have `#toplevel_page_wc-admin-path--payments-welcome div.wp-menu-image::before` css applied to it.

includes/class-wc-calypso-bridge-payments.php
moon0326: Thank you for testing 👍 That's interesting. That vertical gray is the reason why I kept the CSS. For some reason, passing `icon` property was removed when I visit one of the WC pages.

I'll look into it

moon0326: Just checked it with Brave and it worked fine for me.

Screen Shot 2021-07-21 at 7 48 40 PM

The icon also looks identical to WC Pay plugin icon

Screen Shot 2021-07-21 at 7 49 32 PM

Could you run `npm run build` or `npm run start` and check again please?

ilyasfoo: After rebuild and cleared cache, it worked. I think it might've been my cache, sorry!

ilyasfoo: I forgot to mention, the reversed icon can be seen Calypso

image

ilyasfoo (approved):
I confirm the SVG icon is correct in Calypso now. LGTM! 🚢

PR #719, opened 2021-07-22
Add translated FAQs

ilyasfoo:
This PR uses i18n to translate all FAQ strings. It also lints formatting for modified files and removed `href.li` redirector since we don't need them.

Note that I did not put too much thought into the naming of the string variables, if you do have any better suggestions please let me know! The strings file also grew considerably, but I think we can address splitting strings in the future if required.

### Testing Instructions

1. Review all texts and links and make sure everything works as it was previously.

### Screenshots

image

### DIFF
--- a/src/payments-welcome/faq.tsx
+++ b/src/payments-welcome/faq.tsx
@@ -1,7 +1,7 @@
 /**
  * External dependencies.
  */
-import { Icon, help } from '@wordpress/icons'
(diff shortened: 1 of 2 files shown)


adrianduffell (approved):
LGTM! 🚀 

Thanks for taking this from me. I appreciate it was a tedious job to do 😜

PR #722, opened 2021-07-27
Add condition to disable WCPay menu when store is not US english

ilyasfoo:
This PR adds another condition to display WCPay menu, which is the store's WordPress instance language needs to be in US English. This change should be temporary and removed when we have translations ready.

### Testing Instructions

2. Go to Settings > General.
3. Set site language to anything other than `English (United States)`
3. Observe that the WCPay menu disappears.
4. Optionally, set it back to English (United States) and see it appears again.


moon0326 (approved):
@ilyasfoo Thank you for the PR 👍 This LGTM and tested well 🎉 🚀

PR #723, opened 2021-07-27
Fix missing WCPay icon on ecom plan users

ilyasfoo:
It was discovered that ecom plan users users would have a different classname appended to menu. As such, the WCPay icon in `wc-payments.css` did not get applied and resulted in the following:

This PR would fix this problem by adding another full class path.

### Testing Instructions

1. Set up a new ecommerce plan site in WordPress.com
2. After setting up, revert the site (you can obtain the revert link in wpcom-dev-blogs)
2. Follow the guide in p90Yrv-2j7-p2 to transfer the site to atomic, obtain sftp credentials, and upload bridge to the site
3. Observe the payments icon displayed properly in wp-admin (if it's showing WooCommerce navigation, click on "Back to WordPress Dashboard")

### DIFF
--- a/assets/css/wc-payments.css
+++ b/assets/css/wc-payments.css
@@ -10,8 +10,8 @@
 }
 
 /* stylelint-disable selector-id-pattern */
-#toplevel_page_wc-admin-path--payments-welcome
-div.wp-menu-image::before {
+#toplevel_page_wc-admin-path--payments-welcome div.wp-menu-image::before,
+#toplevel_page_admin-page-wc-admin-path--payments-welcome div.wp-menu-image::before {
 	font-family: 'WCPay' !important;
 	content: '\e900';
 	width: 24px;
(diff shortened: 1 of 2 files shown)


moon0326:
@ilyasfoo  I don't know what has happened, but it looks like SVG image is being rendered now. 

I see overlapped WC Pay icon on both business and ecommerce plans.

Screen Shot 2021-07-27 at 10 50 10 AM

ilyasfoo:
I just checked again and yes it now renders it. How awkward, I've already cleared cache multiple times yesterday and didn't work. Thanks for checking this @moon0326!

ilyasfoo:
Reopening this in case if we're able to set the SVG icon to only show on calypso

moon0326 (approved):
LGTM and tested well 🚀 🚢

PR #728, opened 2021-08-24
Update deps for php8

lsl:
PHP8 requires a newer version of phpunit. This PR updates each dev dep.

Need to test this. There are a couple of major version bumps there.


lsl:
@ilyasfoo hold off looking too closely at this, I'll try get the tests running tomorrow, looks like we're still doing phpunit6 things going by the ci failure looking at phpunit6-compat files.

PR #729, opened 2021-08-24
Hide onboarding checklist in wp-admin

lsl:
Hides the onboarding setup checklist from wp-admin.

Testing Setup checklist:
- Enable woop with a define in your wp-config.php `define( 'WPCOM_ENABLE_WOOP', true );`
- Load https://localhost/wp-admin/index.php with and without the define set. When set, the onboarding checklist should not be shown.

Before:
Screenshot 2021-08-24 at 16-21-27 Dashboard ‹ Test — WordPress

After:
Screenshot 2021-08-24 at 16-20-53 Dashboard ‹ Test — WordPress

Related: https://github.com/Automattic/wp-calypso/issues/55422

### DIFF
--- a/includes/woop.php
+++ b/includes/woop.php
@@ -30,7 +30,7 @@ function wc_calypso_bridge_woop_init() {
 	}
 
 	if ( is_admin() ) {
-		// require_once dirname( __FILE__ ) . '/woop/some-admin-feature.php';.
+		require_once dirname( __FILE__ ) . '/woop/hide-onboarding.php';
 	} else {
 		// require_once dirname( __FILE__ ) . '/woop/some-feature.php';.
 	}
--- a/includes/woop/hide-onboarding.php
+++ b/includes/woop/hide-onboarding.php
@@ -1,0 +1,12 @@
+<?php
+/**
+ * Hide onboarding flows.
+ *
+ * @since 1.8.1
+ * @package WC_Calypso_Bridge
+ */
+
+defined( 'ABSPATH' ) || exit;
+
+// Hide the 'WooCommerce Setup' card from wp-admin.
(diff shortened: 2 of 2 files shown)


ilyasfoo:
Hi @lsl, in the PHP init, you'd have to also check for woocommerce like so: https://github.com/Automattic/wc-calypso-bridge/blob/977561dc5ca655a45b3801f6d2641f537a0f4a3a/includes/class-wc-calypso-bridge-payments.php#L60-L69

This bit us in the butt before, with cases where users had installed WC but deactivated the plugin. As to why we're not checking it earlier in the execution, well I don't have an answer for that right now 😬

lsl:
Nice catch, I was wondering about things like this, each loader does something slightly different, I'll add this and check for similar gotchas I might be missing.

includes/yesno.php
```diff
@@ -0,0 +1,25 @@
...
+ * @since 1.8.1
+ * @package WC_Calypso_Bridge
+ */
+
+/**
+ * Always return 'yes'.
+ *
+ * @return string yes
+ */
+function wc_calypso_bridge_return_yes() {
```
ilyasfoo, line 14: May I know what these are for?

lsl: Filters like this https://github.com/Automattic/wc-calypso-bridge/pull/729/files#diff-64414c498da63dbd91258d202730af80ce94415aafd9a2217970343ff62201c3R12 there will be a few of them and I'd rather just comment the reason for adding the filter than having a named function for each thing we disable. 

The required disable logic is also reversed quite often which means you end up with functions that look like typos:

```
function disable_feature_a() {
    return 'no';
}
function disable_feature_b() {
    return 'yes';
}
```

PR #730, opened 2021-08-24
Restrict Payment gateways on premium plans

deBhal:
This PR restricts available payment methods for premium users, allowing only WCPay and core Woocommerce gateways.

#### Testing Instructions

- Install an extra payment gateway on a local install. e.g. [paypal](https://wordpress.org/plugins/woocommerce-paypal-payments/)
- In wp-config.php shim APD and set some defines so that plan lookups go through your shim:
```
define( 'IS_ATOMIC', true );
define( 'ATOMIC_CLIENT_ID', 2 ); // WP-on-atomic client id
define( 'ATOMIC_SITE_ID', 149647132 ); // wooptest3.wpcomstaging.com

// Shim APD
class Atomic_Persistent_Data {
	public $site_sticker_woop = true;
	public $WPCOM_PLAN = 'free';
	public function __get( $name ) {
		return null;
	}
}
```
- *Test:* Alter the WPCOM_PLAN variable to different plan levels, `business`, `ecommerce`, `free`, `premium`, `personal`. Only business and ecommerce plans should display paypal.
- *Test:* Disable the woop feature entirely by changing the site_sticker_woop value to `null`.

#### Unittests
```
./docker/phpunit
```

#### Before:
Screenshot 2021-09-17 at 13-32-53 WooCommerce settings ‹ test — WordPress

#### After:
Screenshot 2021-09-17 at 13-33-22 WooCommerce settings ‹ test — WordPress


lsl:
Had to do a manual rebase, retested, I think it still works.

includes/woop.php
```diff
@@ -32,5 +32,7 @@ function wc_calypso_bridge_woop_init() {
 	if ( is_admin() ) {
 		require_once dirname( __FILE__ ) . '/woop/hide-onboarding.php';
 	}
+
+	require_once dirname( __FILE__ ) . '/woop/limit-payment-gateways.php';
```
lsl, line 36: Bit of a ramble touching on this code made me think about: 

I wouldn't mind figuring out how to properly separate admin code, a `woop/admin` dir maybe? include all and rely on the include to do the is_admin check? probably some code wants to assume that without defining things the front end can access.

includes/woop/limit-payment-gateways.php
```diff
@@ -0,0 +1,33 @@
...
+				return in_array( get_class( $payment_gateway ), $class_whitelist, true );
+			}
+
+			return false;
+		}
+	);
+}
+
+if ( wc_calypso_bridge_is_personal_plan() ) {
+	add_filter( 'woocommerce_payment_gateways', 'wc_calypso_bridge_woop_limit_personal_payment_gateways' );
```
lsl, line 32: I dropped the `1, 11` param attachment because it looked like a typo (order looked wrong) but now I'm wondering if that's what was enabling wc-payments to break through this logic. Do we need to add them here or change the priority back again?

wc-calypso-bridge.php
```diff
@@ -34,6 +34,24 @@
...
+	 * Returns if a site is a personal plan site (i.e. WOOP).
+	 * The `at_options` array is created during provisioning. Usually it is 'business' or 'ecommerce'
+	 * To Test: update_option( 'at_options', array( 'plan_slug' => 'personal' ) );
+	 * or: wp option set at_options --format=json '{ "plan_slug": "personal" }'
+	 *
+	 * @return bool True if the site is a personal site.
+	 */
+	function wc_calypso_bridge_is_personal_plan() {
+		if ( class_exists( 'Atomic_Plan_Manager' ) && method_exists( 'Atomic_Plan_Manager', 'current_plan_slug' ) ) {
+			return Atomic_Plan_Manager::current_plan_slug() === Atomic_Plan_Manager::PERSONAL_PLAN_SLUG;
```
lsl, line 48: Having to gate access to this class isn't ideal. I'm tempted to say just hard code the slugs anywhere outside of wpcomsh. And in turn use a single wrapper function to get the slug in bridge that can be filtered before use, we then centralize the gating to one function. We're creating plan changes overhead with these named functions regardless so we may as well just skip the complexity and hard code the slugs directly.

Alternatively we could extract the dependency (or a stub) out with composer.

lsl:
Added a couple of comments but they're really just things I was wondering about when I crossed this code, not something for this PR but I wouldn't mind getting your input on my thoughts for future PR's / refactors.

lsl:
@ilyasfoo fyi the bootrstrap.php changes

PR #731, opened 2021-08-27
Fix tests, update tests/bootstrap.php to match woocommerce changes

lsl:
This PR updates the calypso-bridge bootstrap with new loading logic from [woocommerce](https://github.com/woocommerce/woocommerce/blob/trunk/tests/legacy/bootstrap.php).

Let me know if this isn't the right approach but it looks like this is where we got the original bootstrapping logic from (modified to include our own tests once upstream test framework is loaded.

Tests are currently failing due to changes to woocommerce and movement of legacy code. 

e.g. https://travis-ci.org/github/Automattic/wc-calypso-bridge/builds/774724440

#### Before:
before

#### After:
after

#### Testing instructions:

You need mysql running locally. Personally I'm using [mysql from ganon](https://github.com/Automattic/ganon/blob/ac81af0e363807f2ac87fd628bc4983933e0bbbb/docker-compose.yml#L9) but any mysql instance on localhost should work for this setup.

```
# Cleanup previous install attempts
rm -rf /tmp/wordpress-tests-lib /tmp/wordpress
mysql -h127.0.0.1 -P3306 -uroot -proot -e"drop database wordpress_tests"

# Run the setup install process, installs into /tmp and a database named wordpress_tests accessed through the loopback
./bin/install-wp-tests.sh wordpress_tests root root 127.0.0.1:3306
```

On my system, in order to get the woocommerce `npm install` command to successfully execute in install-wp-tests.sh I needed this patch:

```diff
diff --git a/bin/install-wp-tests.sh b/bin/install-wp-tests.sh
index 70c150b..892db6b 100755
--- a/bin/install-wp-tests.sh
+++ b/bin/install-wp-tests.sh
@@ -181,6 +181,8 @@ install_deps() {
 
        # Bring in WooCommerce Core dependencies
        cd "woocommerce"
+       source /usr/share/nvm/nvm.sh
+       nvm use "lts/*"
        npm install
        composer install --no-dev
```

If you have the right version of node installed you probably won't need this (12 and 14 seem to work, 16 fails). I'm aware this is less than ideal but you only need to run the install script successfully one time (if you're /tmp doesn't get cleared on reboot!) and fixing this will require a fair bit of refactoring of the install script to allow more diverse setups. I also think it's worth dockerizing this but taking a look into that yesterday I decided this isn't the PR for that.

To run the tests
```
./vendor/bin/phpunit
```

### DIFF
--- a/phpunit.xml
+++ b/phpunit.xml
@@ -7,7 +7,6 @@
 	convertNoticesToExceptions="true"
 	convertWarningsToExceptions="true"
 	verbose="true"
-	syntaxCheck="true"
 	>
 	<testsuites>
 		<testsuite name="WooCommerce Calypso Bridge Test Suite">
--- a/tests/bootstrap.php
+++ b/tests/bootstrap.php
@@ -1,6 +1,8 @@
 <?php
 /**
  * WooCommerce Calypso Bridge Unit Tests Bootstrap
+ *
+ * Loosely based on woocommerce/tests/legacy/bootstrap.php.
  */
 class WC_Calypso_Bridge_Unit_Tests_Bootstrap {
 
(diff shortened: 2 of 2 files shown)


ilyasfoo (approved):
I've tried this in my docker instance, and the tests successfully ran. This does add some complexity to bootstrap, but I don't think we have any other way around that's simpler at this moment.

I'd say my only comment is if it's possible to also update README.md on how you got it working for your local, would be super.

Appreciate your effort in fixing our tests @lsl! 🚢

lsl:
> I'd say my only comment is if it's possible to also update README.md on how you got it working for your local, would be super.

Lets do that in another PR, the instructions in this PR are pretty messy / confusing, I think I'm going to want to dockerize it for ease of use.

ilyasfoo:
@lsl That'll be great!

PR #740, opened 2021-09-03
Add docker for running phpunit / phpcs locally

lsl:
This PR adds a `docker` directory and some helper scripts to run `phpunit` and `phpcs` locally during development.

- Made the implicit dependency on yoast phpunit polyfills explicit per https://core.trac.wordpress.org/changeset/51598.
- Replicated most of `./bin/install-wp-test.sh` in the Dockerfile directly. This worked out because the all the required db bootstrapping is done by wp tests on startup and database creation is handled by the mysql image in its own entrypoint.
- Added `docker-compose.yml` to manage mysql / wordpress `depends_on` dependency for running tests.
- Added `phpunit`, `phpcs`, `phpcs-changed` scripts. There is also a `docker-compose` script but you shouldn't need it unless you want to run a `down` command to clear things. `phpcs-changed` was added due to there being a lot of pre-existing lint errors.
- Added readme.md to outline usage / note about getting things running in osx.

#### Testing:
```
./docker/docker-compose build
./docker/phpunit
./docker/phpcs
./docker/phpcs-changed
```

#### Usage:
docker-after

### DIFF
--- a/composer.json
+++ b/composer.json
@@ -29,7 +29,8 @@
     "wp-coding-standards/wpcs": "1.1.0",
     "woocommerce/woocommerce-sniffs": "*",
     "wimg/php-compatibility": "9.0.0",
-    "dealerdirect/phpcodesniffer-composer-installer": "^0.7.0"
+    "dealerdirect/phpcodesniffer-composer-installer": "^0.7.0",
+    "yoast/phpunit-polyfills": "1.0.1"
   },
   "scripts": {
     "test": [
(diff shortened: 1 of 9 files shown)


lsl:
@ilyasfoo could you take this for a spin when you get a chance? Curious if all the scripts / docker stuff work on mac without me having to mess wtih things further. I suspect you might need to drop mariadb for mysql but I'm trying to avoid it.

@joshuatf / @psealock just fyi as you're suggested reviewers but I think that's just because I had to touch `composer.json`.

Secondary question: Any problem with merging this? Is there a way to omit these files from the package on build (if we even do that?)

lsl:
Merging this into woop branch so I can use it to test #730 changes.

Issue #748, opened 2021-09-09
Calpyso menu items are out of sync on eCom plans


moon0326:
WooCommerce displays two different menus on Calpyso and WPAdmin

`wc-calypso-bridge` enables the `navigation` feature [by default ](https://github.com/Automattic/wc-calypso-bridge/blob/master/includes/class-wc-calypso-bridge-setup.php#L40) on eCom plans.

When the `navigation` is enabled, we should not see the legacy coupon menu **WooCommerce -> Coupons**.

Calypso pages use the admin menu API (https://public-api.wordpress.com/wpcom/v2/sites/:site-id/admin-menu/?_envelope=1) to render the menus. The problem is that [this line](https://github.com/Automattic/wc-calypso-bridge/blob/master/includes/class-wc-calypso-bridge-setup.php#L40) does not get hit when the API is requested and it renders **WooCommerce -> Coupons** since the feature is not enabled.

PR #749, opened 2021-09-09
Set woocommerce_navigation_enabled to yes on eCom plans

moon0326:
Fixes https://github.com/Automattic/wc-calypso-bridge/issues/748 and https://github.com/woocommerce/woocommerce-admin/issues/7603

This PR fixes the out-of-sync menu problem by setting `woocommerce_navigation_enabled` menu to `yes` on eCom plans.

### DIFF
--- a/includes/class-wc-calypso-bridge-setup.php
+++ b/includes/class-wc-calypso-bridge-setup.php
@@ -35,6 +35,7 @@ class WC_Calypso_Bridge_Setup {
 	 * Constructor.
 	 */
 	private function __construct() {
+		$this->add_navigation_option();
 		add_filter( 'default_option_woocommerce_onboarding_profile', array( $this, 'set_business_extensions_empty' ) );
 		add_filter( 'option_woocommerce_onboarding_profile', array( $this, 'set_business_extensions_empty' ) );
 		add_filter( 'default_option_woocommerce_navigation_enabled', array( $this, 'enable_navigation_by_default' ) );
@@ -134,6 +135,12 @@ class WC_Calypso_Bridge_Setup {
 	 */
 	public function enable_navigation_by_default() {
 		return 'yes';
+	}
+
+	public function add_navigation_option() {
+		if ( false === get_option( 'woocommerce_navigation_enabled' ) ) {
+			update_option( 'woocommerce_navigation_enabled', 'yes' );
+		}
 	}
 
 	/**


mattsherman:
@moon0326 This forces navigation to be enabled, right? Is there agreement that we want to force navigation for the eComm plan? It would be good to get confirmation on that? /cc @pmcpinto Do you have any thoughts, or know who we should check with to verify this change?

Also, if we force enable navigation, we should remove the navigation feature toggle from the UI, so that users don't think that they can disable it.

moon0326:
> @moon0326 This forces navigation to be enabled, right? Is there agreement that we want to force navigation for the eComm plan? It would be good to get confirmation on that? /cc @pmcpinto Do you have any thoughts, or know who we should check with to verify this change?
> 
> Also, if we force enable navigation, we should remove the navigation feature toggle from the UI, so that users don't think that they can disable it.

@mattsherman AFAIK, we already enable the navigation by default for the eCom plan. See https://github.com/Automattic/wc-calypso-bridge/pull/749/files#diff-6913cd5881364970fe5a0344296a03005ccab72b20b844d5e040e649ba91531aR41

This PR actually sets the option value for the Calpyso pages.

pmcpinto:
> @moon0326 This forces navigation to be enabled, right? Is there agreement that we want to force navigation for the eComm plan? It would be good to get confirmation on that? /cc @pmcpinto Do you have any thoughts, or know who we should check with to verify this change?

@mattsherman @moon0326 thanks for the ping. One of the goals with the new navigation launch on eComm was to get quantitative data on the % of merchants disabling the new experience. Ideally, we should:
- Allow customers who disabled the new navigation to continue using the "old" version
- Continue providing a setting where it's possible to disable/enable the new navigation
Let me know if this is doable.

moon0326:
> Also, if we force enable navigation, we should remove the navigation feature toggle from the UI, so that users don't think that they can disable it.

Since the change checks for `false === get_option(...)`, you can still disable the feature as before in **WooCommerce -> Settings -> Adavanced -> Features**

mattsherman:
> Since the change checks for `false === get_option(...)`, you can still disable the feature as before in **WooCommerce -> Settings -> Adavanced -> Features**

Ah, right... I forgot that was how `get_option()` works!

Why do we need to do this in two places? Is there not a single place we could take care of it?

moon0326:
> > Since the change checks for `false === get_option(...)`, you can still disable the feature as before in **WooCommerce -> Settings -> Adavanced -> Features**
> 
> Ah, right... I forgot that was how `get_option()` works!
> 
> Why do we need to do this in two places? Is there not a single place we could take care of it?

So the problem is that [includes/class-wc-calypso-bridge-setup.php](https://github.com/Automattic/wc-calypso-bridge/blob/f25d89887b6eca8f1144ff5a14f8a2d9d6e3dde8/includes/class-wc-calypso-bridge-setup.php) does not get hit from Calpyso pages on an eCom plan, which means that the pages will render an incorrect set of menus. 

By [setting](https://github.com/Automattic/wc-calypso-bridge/blob/f25d89887b6eca8f1144ff5a14f8a2d9d6e3dde8/includes/class-wc-calypso-bridge-setup.php) `woocommerce_navigation_enabled` option in the database, Calypso pages should render the menu as expected. This is a good place to set the option since the function will be called when the users install WC during the onboarding process.

Details are explained in the first linked [issue](https://github.com/Automattic/wc-calypso-bridge/issues/748) as well.

moon0326:
Hey @mattsherman ✋  Could you give it another look please? Thanks!

PR #752, opened 2021-09-21
Hide the wccom Marketplace menu for plans that don't have the ability to install plugins

lsl:
Hide the Marketplace / extensions page.

#### Before
Screenshot 2021-09-21 at 17-00-05 Home ‹ WooCommerce ‹ test — WooCommerce

#### After
Screenshot 2021-09-21 at 16-59-53 Home ‹ WooCommerce ‹ test — WooCommerce

#### Testing Instructions

- View wc-admin home
- Reload the page with / without the plugin enabled

Fixes https://github.com/Automattic/wp-calypso/issues/55428


lsl:
@olaolusoga @ianstewart hiding the woo marketplace using woo's provided hook also hides the subscription page, this isn't something I was aware of / thought about and raises a few questions:

- Will the dotcom marketplace be able to handle woo subscriptions? 
- It seems a bit questionable to turn access to subscriptions off for new customers, is this something we're ok with?
- How should we handle existing woo subscriptions? (I'm assuming it ties heavily into the marketplace pages with various links back etc)

lsl:
decision: hide for premium users

louwie17:
Hi @lsl, this PR has been sitting around for some time, will these changes still be relevant?

lsl:
Sorry missed your ping, still relevant, not needed until we start looking at premium plans on atomic that won't have the ability to install extensions.

obenland:
@cpapazoglou Is this one that Bespin could look at?

cpapazoglou:
> @cpapazoglou Is this one that Bespin could look at?

Yes, we can handle **removing the "Marketplace" menu for sites not on business / ecommerce plans** where users shouldn't be able to install plugins (at least from external sources like WooCommerce.com).

Also, the Marketplace / My Subscriptions WooCommerce menus will be obsolete when we have added all the wc extensions to wpcom marketplace. 

BTW, it seems that "My Subscriptions" is now part of the "Marketplace" screen rather than one more submenu under "WooCommerce".
SS 2022-01-18 at 12 00 22

@obenland are there any insights for when this issue should be resolved?

obenland:
Not that I'm aware of. As soon as you're reasonable able to get to it?

PR #765, opened 2021-11-10
Remove calls to old navigation note

ilyasfoo:
This PR removes the old navigation note since it has never been shown before and will start showing when cron events are fixed in WPCOM, which will cause confusion for users.

### Testing Instructions

1. Install [Crontrol](https://wordpress.org/plugins/wp-crontrol/) plugin
2. Make sure the note named `wc-admin-navigation-learn-more` does not exist in notes table in database
2. Go to Tools > Cron Events, manually run `wc_calypso_bridge_daily`
3. Go to WooCommerce > Home, make sure the note with content `Introducing a streamlined, commerce-first navigation experience,` does not appear

### DIFF
--- a/includes/class-wc-calypso-bridge-notes.php
+++ b/includes/class-wc-calypso-bridge-notes.php
@@ -41,11 +41,6 @@ class WC_Calypso_Bridge_Notes {
 	 * Include notes and initialize note hooks.
 	 */
 	public function init() {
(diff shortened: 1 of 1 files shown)


adrianduffell (approved):
@ilyasfoo This tested well and LGTM. Thanks for catching this! 🚀

PR #770, opened 2021-12-03
Update check to see if feature exists before adding

### LINKED ISSUE #747: Navigation: Feature toggle is shown twice when wc-admin is installed
In Ecom plan stores, when wc-admin is installed, there are two checkboxes appearing to enable or disable the new nav.
Screen Shot 2021-09-02 at 2 43 13 PM

It's required to uncheck both to disable the navigation, then the nav disappears altogether. A refresh will bring back the unified nav but it's quite disorienting.
Screen Shot 2021-09-09 at 1 51 58 PM

There should only be one checkbox to enable/disable the new nav, and the transition to the WPCOM nav shouldn't require a manual refresh.

joshuatf:
Fixes #747 

The previous check was checking for an array key, but as far as I can tell, WCA features are and have been using an array of feature keys not indexed by the key itself.

### Before
Screen Shot 2021-12-03 at 2 57 47 PM

### After
Screen Shot 2021-12-03 at 3 08 29 PM

### Testing

1. Set up this repo to mimic the bridge dev mode and ecomm plans - https://github.com/Automattic/wc-calypso-bridge
2. Install and activate the latest `main` from WCA
3. Navigate to the WC Settings->Advanced->Features.
4. Make sure only one toggle is shown for the navigation
5. Optionally, install a really old version of WCA and/or WooCommerce to make sure a single toggle is still shown.

### DIFF
--- a/.travis.yml
+++ b/.travis.yml
@@ -8,7 +8,7 @@ matrix:
     env: WP_VERSION=latest WP_MULTISITE=0 WP_CORE_DIR=/tmp/wordpress RUN_PHPCS=1 COMPOSER_DEV=1
 
 before_install:
-  - nvm install 'lts/*'
+  - nvm install 14
   # Composer 2.0.7 introduced a change that broke the jetpack autoloader in PHP 7.0 - 7.3.
   - composer self-update 2.0.6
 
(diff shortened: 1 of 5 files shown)


louwie17 (approved):
Tested great, nice work @joshuatf

PR #774, opened 2022-01-21
 Add docker for running phpunit / phpcs locally

lsl:
Replaces #740.

This PR adds a `docker` directory and some helper scripts to run `phpunit` and `phpcs` locally during development.

- ~Made the implicit dependency on yoast phpunit polyfills explicit per https://core.trac.wordpress.org/changeset/51598.~ Done elsewhere since #740 
- Replicated most of `./bin/install-wp-test.sh` in the Dockerfile directly. This worked out because the all the required db bootstrapping is done by wp tests on startup and database creation is handled by the mysql image in its own entrypoint.
- Added `docker-compose.yml` to manage mysql / wordpress `depends_on` dependency for running tests.
- Added `phpunit`, `phpcs`, `phpcs-changed` scripts. There is also a `docker-compose` script but you shouldn't need it unless you want to run a `down` command to clear things. `phpcs-changed` was added due to there being a lot of pre-existing lint errors.
- Added readme.md to outline usage / note about getting things running in osx.

#### Testing:
```
./docker/docker-compose build
./docker/phpunit
./docker/phpcs
./docker/phpcs-changed
```

#### Usage:
docker-after

### DIFF
--- a/docker/Dockerfile
+++ b/docker/Dockerfile
@@ -9,11 +9,14 @@ RUN apk -q --update --no-cache add tini shadow less git subversion nodejs npm \
     php7-phar php7-session php7-simplexml php7-tokenizer php7-pecl-xdebug php7-xml \
     php7-xmlwriter php7-zip php7-zlib
 
+# Woo requires pnpm
+RUN npm install -g pnpm
+
 # Install wp-cli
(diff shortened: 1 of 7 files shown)


moon0326:
Hi @lsl ✋ 

Is this PR still needed?

lsl:
No, I don't need. Closing out but feel free to take over if its something you'd like to add.

PR #783, opened 2022-04-14
Use feature check instead of plan manager check

obenland:
This change will allow wpcomsh to get rid of the Plan Manager class entirely.

Using a feature check to determine the plan is not ideal, but it's the best we can do right now.

Props @mreishus.

### DIFF
--- a/wc-calypso-bridge.php
+++ b/wc-calypso-bridge.php
@@ -37,14 +37,12 @@ define( 'WC_MIN_VERSION', '3.0.0' );
 if ( ! function_exists( 'wc_calypso_bridge_is_ecommerce_plan' ) ) {
 	/**
 	 * Returns if a site is an eCommerce plan site or not.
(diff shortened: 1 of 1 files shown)


moon0326 (commented):
Hi ✋ 

Code LGTM, but I'm not sure if I tested it correctly.

1. I signed up for a new Pro plan.
2. Converted the blog to a dev blog via https://mc.a8c.com/atomic/wpcom-dev-blogs/
3. SSHed into the server and modified `wc-calypso-bridge.php`
4. `wpcom_site_has_feature( \WPCOM_Features::ECOMMERCE_MANAGED_PLUGINS )` returned false for me.

Is this change for the legacy e-commerce plan only?

obenland:
> Is this change for the legacy e-commerce plan only?

Correct, yes, only the eCommerce plan has that feature

mmtr (approved):

moon0326 (approved):

PR #785, opened 2022-04-18
Update test install script to use the new WC monorepo setup

moon0326:
This PR fixes failing CI by updating the install script to use the new WC monorepo setup.


chihsuan (approved):
Looks good. Thanks for fixing this!

PR #788, opened 2022-05-17
Make sure WC is active

### LINKED ISSUE #787: Fatal error when WooCommerce is deactivated
`wc_calypso_bridge_daily` event throws a PHP fatal error when WC is installed but deactivated.

We only load `wc-calypso-bridge` when it [detects](https://github.com/Automattic/wc-calypso-bridge/blob/master/wc-calypso-bridge.php#L20) WooCommerce, but we also need to check whether the plugin is activated or not. Otherwise, there is a chance that `wc-calypso-brdige` doesn't have access to certain classes when WC is deactivated.

Discussion: p1652716487529899-slack-C7U3Y3VMY

moon0326:
Fixes #787 

This PR adds a check to make sure WC is active to prevent any undesired errors by accessing WC classes. 

### Test Instructions

1. Clone `trunk` branch
2. Active both `WooCommerce` and `wc-calpyso-brdige`
3. Deactivate `WooCommerce`
4. Install [WP Crontrol](https://wordpress.org/plugins/wp-crontrol/)
5. Navigate to `Tools -> Cron Events`  and run `wc_calypso_bridge_daily` event
6. Notice a fatal PHP error.
7. Clone this branch
8. Repeat the same steps. The fatal error should not show up again.

If you can test it on a dev blog, that would be nice (not required)

### DIFF
--- a/wc-calypso-bridge.php
+++ b/wc-calypso-bridge.php
@@ -17,7 +17,13 @@ if ( ! defined( 'ABSPATH' ) ) {
 	return;
 }
 
-if ( ! file_exists( WP_PLUGIN_DIR . '/woocommerce/woocommerce.php' ) ) {
+if ( ! function_exists( 'is_plugin_active' ) ) {
+    require_once ABSPATH . 'wp-admin/includes/plugin.php';
+}
+
+$wc_plugin_path = 'woocommerce/woocommerce.php';
+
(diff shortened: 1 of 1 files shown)


ilyasfoo (approved):
Tested well, LGTM! 🚢

PR #791, opened 2022-07-15
Only apply the rule when nav unification is rendered

moon0326:
Fixes https://github.com/Automattic/wp-calypso/issues/65522

This PR only applies the CSS when there is nav-unification (272px width menu).


lupus2k (approved):
I made a comment about this [here](https://github.com/Automattic/wp-calypso/issues/65522#issuecomment-1185839659) and looks almost the same fix.
LGTM!

moon0326:
> I made a comment about this [here](https://github.com/Automattic/wp-calypso/issues/65522#issuecomment-1185839659) and looks almost the same fix. LGTM!

Thank you! I just saw your comment and added `.jetpack-connected` selector as well.

jorgeatorres (approved):
LGTM 🚢

Issue #797, opened 2022-08-29
`pre_option_woocommerce_allow_tracking` hook doesn't work for WCTracker


moon0326:
This [line](https://github.com/Automattic/wc-calypso-bridge/blob/master/includes/class-wc-calypso-bridge-tracks.php#L55) was added to set `woocommerce_allow_tracking` to `yes` for both Tracks events (pixel) and WCTracker (corn) use cases. However, it turned out that the change didn't work for WCTracker.

1. WCTracker is initialized in [class-woocommerce.php](https://href.li/?https://github.com/woocommerce/woocommerce/blob/trunk/plugins/woocommerce/includes/class-woocommerce.php#L548), which is included in [woocommerce.php](https://href.li/?https://github.com/woocommerce/woocommerce/blob/trunk/plugins/woocommerce/woocommerce.php#L34) very early.

2. wc-calypso-bridge adds the `pre_option_woocommerce_allow_tracking` hook via [init](https://href.li/?https://github.com/Automattic/wc-calypso-bridge/blob/master/wc-calypso-bridge.php#L64) hook, which runs after the WCTracker initialization. This makes the get_option always return no

Possible solution: Instead of using the pre-hook, set the value to `yes` instead.

PR #798, opened 2022-08-29
Always set woocommerce_allow_tracking to yes

moon0326:
Fixes https://github.com/Automattic/wc-calypso-bridge/issues/797

This PR sets `woocommerce_allow_tracking` to `yes` in database instead of using `pre_option_woocommerce_allow_tracking` hook to fix the issue described in https://github.com/Automattic/wc-calypso-bridge/issues/797

### Testing instructions

1. Start with a fresh install. Do not start OBW and make sure `woocommerce_allow_tracking` is set to `no`
2. Checkout this branch and enable `wc-calypso-bridge` plugin.
3. Confirm `woocommerce_allow_tracking` is set to `yes` in `wp_options`

### DIFF
--- a/includes/class-wc-calypso-bridge-tracks.php
+++ b/includes/class-wc-calypso-bridge-tracks.php
@@ -53,19 +53,19 @@ class WC_Calypso_Bridge_Tracks {
 		add_filter( 'woocommerce_apply_tracking', '__return_true' );
 		add_filter( 'woocommerce_apply_user_tracking', '__return_true' );
 
-        $this->enable_tracking();
+		$this->enable_tracking();
 	}
 
 	/**
-     * Set woocommerce_allow_tracking to yes.
-     *
+	 * Set woocommerce_allow_tracking to yes.
+	 *
 	 * @return void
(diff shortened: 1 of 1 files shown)


ilyasfoo (approved):
Tested well, LGTM!

manospsyx:
@ilyasfoo @moon0326 as far I understand, your diagnosis suggests that this is an issue with hooking priority. Instead of modifying the database value directly, it might be best to continue to filter the return value of `get_option` dynamically.

My understanding is that (if it's not already done) eventually, we'll want to dynamically change the return value depending on the merchant's privacy preferences under `https://wordpress.com/me/privacy` -- which means that it would be handy to continue to have a way to filter the return value "on the fly".

moon0326:
> 

@manospsyx Good suggestion. I'll look into it and see if we can adjust the priority.

moon0326:
@manospsyx Opened a new PR [here](https://github.com/Automattic/wc-calypso-bridge/pull/802). Thank you for the suggestion 👍

PR #802, opened 2022-08-30
Use pre_option_ hook for woocommerce_allow_tracking

### LINKED ISSUE #801: Use pre_option_ hook for woocommerce_allow_tracking
We've recently merged https://github.com/Automattic/wc-calypso-bridge/pull/798 to set `woocommerce_allow_tracking` value directly in the D.B.

However, as pointed out by @manospsyx [here](https://github.com/Automattic/wc-calypso-bridge/pull/798), it would be best if we keep the pre_option_hook

Look into see if there is a way to keep `pre_option_` hook.

moon0326:
Closes #801 

This PR moves `pre_option_woocommerce_allow_tracking` out of `init` hook and adds it right after the track class inclusion.

It's guaranteed to run before WooCommerce for being a MU plugin.

### Testing Instructions

Let's test it as a MU plugin to mimic Atomic env as close as possible.

1. Create `mu-plugins` in `wp-content` if it doesn't exist.
2. Move `wc-calypso-bridge` directory to the `mu-plugins` directory.
3. Create `wc-calypso-bridge-loader.php` in the `mu-plugins` with the following content.
```
<?php
/**
 * Plugin Name: WC Calypso Brdige Loader
 * Description: WordPress.com provided functionality & tools pre-installed and activated on all Atomic Sites
 *
 * @package wc-calypso-bridge
 */

require_once WPMU_PLUGIN_DIR . '/wc-calypso-bridge/wc-calypso-bridge.php';
```

4. Create a fresh site and start OBW.
5.  Click `Continue` on the first step and make sure the tracking confirmation modal doesn't show up.

### DIFF
--- a/includes/class-wc-calypso-bridge-tracks.php
+++ b/includes/class-wc-calypso-bridge-tracks.php
@@ -52,21 +52,14 @@ class WC_Calypso_Bridge_Tracks {
 		// Always opt-in to Tracks, WPCOM user tracks preferences take priority.
 		add_filter( 'woocommerce_apply_tracking', '__return_true' );
 		add_filter( 'woocommerce_apply_user_tracking', '__return_true' );
-
-		$this->enable_tracking();
 	}
 
 	/**
-	 * Set woocommerce_allow_tracking to yes.
-	 *
-	 * @return void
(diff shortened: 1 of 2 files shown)


chihsuan (approved):
Nice approach. Tested well and looks good! 👍

PR #811, opened 2022-09-15
Skip the Onboarding Profiler in the Ecommerce Plan

### LINKED ISSUE #808: Skip the Onboarding Profiler in the Ecommerce Plan
### Description

The goal of this project is to:

- [x] Skip the Core Profiler for all merchants on the Ecommerce Plan.
- [x] Give merchants a way to enter their store’s address by preserving the existing Add store details Task in a pending state.
- [x] Change the behavior of the Add Store Details Task to take users to the WooCommerce > General settings, and fill in their Store Address.
- [x] Deliver a smooth experience that directs the user’s attention to the Store Address section, and takes them back to the Task List once they have entered an address.

More at peapX7-jn-p2

xristos3490:
Closes #808 

### Description

This PR aims to:
- Skip the WooCommerce setup wizard/profiler
- Redirect the user to the setup task list once they've entered a full address _(and they are in the "onboarding" flow)_.


includes/class-wc-calypso-bridge-woocommerce-admin.php
```diff
@@ -35,6 +35,39 @@ public static function factory() {
...
+		add_action( 'admin_init', array( $this, 'redirect_store_details_onboarding' ) );
+	}
+
+	/**
+	 * Handle the store location's onboarding redirect when user provided a full address.
+	 *
+	 * @since 1.9.4
+	 * @return void
+	 */
+	public function redirect_store_details_onboarding() {
```
manospsyx, line 47: `class-wc-calypso-bridge-setup.php` seems like a better place for this -- you may also be able to clean up a bunch of stuff there now that the Profiler is gone.

includes/class-wc-calypso-bridge-setup.php
```diff
@@ -45,6 +47,30 @@ private function __construct() {
...
+	public function setup_admin() {
+
+		// Make sure setup runs only once.
+		$is_initialized = 'yes' === get_option( 'woocommerce_atomic_initialized', 'no' );
+		if ( $is_initialized ) {
+			return;
+		}
+
+		// Skip the WooCommerce setup wizard.
+		update_option( 'woocommerce_onboarding_profile', array( 'skipped' => true ) );
```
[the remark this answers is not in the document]
xristos3490: Alright! Using the hooks API here is better than hardcoding the option value. 

If we need to **keep and update existing data** from the profiler (if any), the `pre_option_` hook doesn't provide the actual option value; this means that we might need to use the `option_{$option_name}` hooks.

includes/class-wc-calypso-bridge-setup.php
```diff
@@ -86,6 +102,38 @@ public function prevent_redirects_on_activation( $location, $status ) {
...
+		$store_city     = get_option( 'woocommerce_store_city' );
+		$store_postcode = get_option( 'woocommerce_store_postcode' );
+
+		if ( ! empty( $store_address ) && ! empty( $store_city ) && ! empty( $store_postcode ) ) {
+			$should_redirect = true;
+		}
+
+		if ( $should_redirect ) {
+			wp_safe_redirect( admin_url( 'admin.php?page=wc-admin' ) );
+		}
```
obenland, line 134: There seems to be a mismatch between `$should_redirect_home` and `$should_redirect`? Is that needed at all?
```suggestion
		$store_address  = get_option( 'woocommerce_store_address' );
		$store_city     = get_option( 'woocommerce_store_city' );
		$store_postcode = get_option( 'woocommerce_store_postcode' );

		if ( ! empty( $store_address ) && ! empty( $store_city ) && ! empty( $store_postcode ) ) {
			wp_safe_redirect( admin_url( 'admin.php?page=wc-admin' ) );
		}
```

includes/class-wc-calypso-bridge-setup.php
```diff
@@ -35,16 +35,32 @@ public static function get_instance() {
 	 * Constructor.
 	 */
 	private function __construct() {
+		add_action( 'admin_init', array( $this, 'redirect_store_details_onboarding' ) );
```
obenland, line 38: Rather than the generic `admin_init`, is there [a more specific `load-$page_hook` action](https://github.com/WordPress/wordpress-develop/blob/trunk/src/wp-admin/admin.php#L217-L237) that could be hooked into instead?

includes/class-wc-calypso-bridge-setup.php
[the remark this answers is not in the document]
xristos3490: I ended up using the `pre_option_` here for manipulating the value at runtime. Existing data should remain available for direct DB access if needed.

includes/class-wc-calypso-bridge-setup.php
xristos3490: Great idea! The `load-woocommerce_page_wc-settings` hooks seem perfect for this job.

includes/class-wc-calypso-bridge-setup.php
xristos3490: Nice catch! 

> Is that needed at all?

We can definitely work without it ;) 

Cleaned up in https://github.com/Automattic/wc-calypso-bridge/pull/811/commits/93fab1a089c5f27343c713dac97c6f1dedcff966

PR #826, opened 2022-10-05
Limit activity panels to Woo Home page

### LINKED ISSUE #825: Limit activity panels to WooCommerce Home page
### Description

Currently, the WooCommerce Header Bar and associated [activity panels](https://github.com/woocommerce/woocommerce/blob/trunk/plugins/woocommerce-admin/docs/woocommerce.com/activity-panels.md) are injected to every WooCommerce page. These elements:

- Introduce shortcuts to store management functions (process orders, moderate reviews, restock products) in locations where the user navigates to with clear goals (e.g. Settings, Orders, Products, Subscriptions, Analytics). 
- Add a lot of noise and take up screen real estate. 
- Are inconsistent with the design system of WP admin pages, further contributing to a cluttered experience.

That said, the Woo Header Bar and activity panels are a very useful addition to the WooCommerce Home page, which is where the user needs/wants to have an overview of what's happening in their store. The presence of divergent UIs/elements can be justified in the Home page, which introduces its own design system.

manospsyx:
closes #825 

Before:

Screen Shot 2022-10-05 at 12 40 57 PM
Screen Shot 2022-10-05 at 12 42 52 PM

After:

Screen Shot 2022-10-05 at 12 41 32 PM
Screen Shot 2022-10-05 at 12 42 09 PM

Note that limiting the Header Bar to the WooCommerce Home required a revision to the Home page heading styles: Since the Header Bar is now a Home-only element, heading sizes should be optimized relative to each other within this context.

### DIFF
--- a/includes/class-wc-calypso-bridge-woocommerce-admin-features.php
+++ b/includes/class-wc-calypso-bridge-woocommerce-admin-features.php
@@ -4,7 +4,7 @@
  *
  * @package WC_Calypso_Bridge/Classes
  * @since   1.0.0
- * @version 1.9.4
+ * @version 1.9.5
  */
 
 defined( 'ABSPATH' ) || exit;
@@ -81,15 +81,16 @@ class WC_Calypso_Bridge_WooCommerce_Admin_Features {
 	/**
 	 * Add is-woocommerce-home body class.
 	 *
-	 * @param string $classes Body classes.
+	 * @since   1.9.5
 	 *
+	 * @param string $classes Body classes.
 	 * @return string
 	 */
 	public function filter_woocommerce_body_classes( $classes ) {
 
 		$page = PageController::get_instance()->get_current_page();
 
-		if ( $page && 'woocommerce-home' === $page[ 'id' ] ) {
+		if ( $page && 'woocommerce-home' === $page['id'] ) {
 			$classes .= ' is-woocommerce-home';
 		}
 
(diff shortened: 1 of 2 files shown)


includes/class-wc-calypso-bridge-woocommerce-admin-features.php
```diff
@@ -72,6 +78,42 @@ public function filter_woocommerce_admin_features( $features ) {
...
+	 * Add is-woocommerce-home body class.
+	 *
+	 * @since   1.9.5
+	 *
+	 * @param string $classes Body classes.
+	 * @return string
+	 */
+	public function filter_woocommerce_body_classes( $classes ) {
+
+		$page = PageController::get_instance()->get_current_page();
```
xristos3490, line 91: Hmm, I was wondering how this fix would behave when the `woocommerce_admin_disabled` filter is ON. 

I would suggest being more defensive here and checking if the `\Automattic\WooCommerce\Admin\PageController` class exists first.

xristos3490: Does this filter prevent the PageController from being loaded? Not sure, actually.

xristos3490: Just tested it on a WoA site. Everything seems OK!

xristos3490 (approved):

PR #827, opened 2022-10-05
Hide write button in global bar

### LINKED ISSUE #824: Hide Write button in global bar

PanosSynetos:
closes #824

This should not be merged yet, as we cannot hide the `Write` button in Calypso pages. The fix only works in `wp-admin` and `frontend` when logged in.

### Testing

- Navigate to any WP "native" admin pages, and the `write` button should not appear.
- Navigate to home (when logged in), and the `write` button should not appear.
- Navigate to any calypso pages, and should still exist

### DIFF
--- a/class-wc-calypso-bridge.php
+++ b/class-wc-calypso-bridge.php
@@ -4,7 +4,7 @@
  *
  * @package WC_Calypso_Bridge/Classes
  * @since   1.0.0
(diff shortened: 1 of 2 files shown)


xristos3490:
I've also opened a PR to hide this button on the Calypso side here https://github.com/Automattic/wp-calypso/issues/69068

PR #833, opened 2022-10-07
Enable logger to check woocommerce installation hooks

### LINKED ISSUE #832: Enable logger to check woocommerce installation hooks
Before releasing #822 (Default to block based checkout) we need to be able to check if execution passes from certain hooks.

As discussed with @manospsyx , the hooks would be
- `woocommerce_admin_onboarding_industries`
- `woocommerce_newly_installed`
- `woocommerce_create_pages`

PanosSynetos:
closes #832 

Tested by hooking into `woocommerce_init` - The log file was created as seen in screenshot

image

### DIFF
--- a/class-wc-calypso-bridge.php
+++ b/class-wc-calypso-bridge.php
@@ -76,6 +76,7 @@ class WC_Calypso_Bridge {
 		 */
 		add_filter( 'woocommerce_create_pages', static function ( $pages ) {
 			self::log_message( 'Filter: woocommerce_create_pages passed.' );
+			add_option( 'wc_calypso_bridge_debug_woocommerce_create_pages', time() );
 
 			return $pages;
 		}, PHP_INT_MAX );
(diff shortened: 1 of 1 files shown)


class-wc-calypso-bridge.php
```diff
@@ -34,6 +34,13 @@ class WC_Calypso_Bridge {
 	 */
 	protected static $instance = null;
 
+	/**
+	 * Logger instance.
+	 *
+	 * @var WC_Logger $logger
+	 */
+	protected static $logger = null;
```
xristos3490, line 42: I would suggest using the native `wc_get_logger` function that includes some kind of static management and doesn't instantiate the logger each time. You can see an example in gift cards here: https://github.com/woocommerce/woocommerce-gift-cards/blob/master/woocommerce-gift-cards.php#L377-L387

PanosSynetos: Thank you Chris - I applied your comments!

xristos3490:
Great thinking, Panos! Looks good! I liked the `time()` as a value here!

xristos3490 (approved):

PR #839, opened 2022-10-12
Remove OBW hook and create pages on `woocommerce_installed`

### LINKED ISSUE #838: Hook `maybe_create_wc_pages` on a more appropriate place
Currently, `maybe_create_wc_pages` is hooked into `woocommerce_admin_onboarding_industries`

Due to the fact that we've removed OBW, we shouldn't hook onto OBW related places.

Going back to the changes, I see that there were many different approaches to solve this issue

- March 12, 2021, `woocommerce_admin_newly_installed` https://github.com/Automattic/wc-calypso-bridge/commit/0706cefba29a916de5be03c0357dfeb2b5f85d2a
- March 12, 2021, `wc_admin_daily` - https://github.com/Automattic/wc-calypso-bridge/commit/699afe1914fefa17fc23e22bd022892edb1e9afd
- April 2, 2021, `woocommerce_admin_onboarding_industries` - https://github.com/Automattic/wc-calypso-bridge/commit/de19148ee452214d4c922ced78d011ac8c402b45

After the investigation I've done, it seems like using `woocommerce_installed` is a better fit for this code (and it's not wrapped around any WC Version checking (like `woocommerce_admin_newly_installed` )

PanosSynetos:
closes #838

### DIFF
--- a/includes/class-wc-calypso-bridge-plugins.php
+++ b/includes/class-wc-calypso-bridge-plugins.php
@@ -137,7 +137,6 @@ class WC_Calypso_Bridge_Plugins {
 			delete_option( "woocommerce_{$page}_page_id" );
 		}
 
-		do_action( 'wc_calypso_bridge_maybe_create_pages' );
 		WC_Install::create_pages();
 		update_option( $option_name, 'yes' );
 	}
--- a/includes/class-wc-calypso-bridge-plugins.php
+++ b/includes/class-wc-calypso-bridge-plugins.php
@@ -4,7 +4,7 @@
  *
  * @package WC_Calypso_Bridge/Classes
  * @since   1.0.0
- * @version 1.9.5
+ * @version 1.9.6
  */
 
 defined( 'ABSPATH' ) || exit;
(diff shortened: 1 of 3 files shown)


includes/class-wc-calypso-bridge-plugins.php
```diff
@@ -108,48 +108,38 @@ public function remove_woocommerce_deactivation_link( $actions, $plugin_file ) {
...
 		// Reset the woocommerce_*_page_id options.
 		// This is needed as woocommerce_*_page_id options have incorrect values on a fresh installation
-		// for an ecom plan. WC_Install:create_pages() might not create all the
+		// for an ecomm plan. WC_Install:create_pages() might not create all the
 		// required pages without resetting these options first.
 		foreach ( [ 'shop', 'cart', 'myaccount', 'checkout' ] as $page ) {
 			delete_option( "woocommerce_{$page}_page_id" );
 		}
 
+		do_action( 'wc_calypso_bridge_maybe_create_pages' );
```
xristos3490, line 140: Why do we need this? Isn't the `wc_pages_created_by_wc_calypso_bridge` option enough to make a decision?

PanosSynetos: I wanted to log `time()` and `wc_pages_created_by_wc_calypso_bridge` gives a yes/no .
I also wanted to keep it consistent with the rest of the loggers

PanosSynetos: OK, removing it, pushing and merging! Thanks!

PR #846, opened 2022-10-27
Remove delete link from managed ecommerce plugins

### LINKED ISSUE #845: Prevent the deletion of Ecommerce managed plugins
### Description

The goal is to hide the "Delete" action in all of the plugins that are managed by the Ecommerce plan. Moreover, it would be nice to indicate that the plugins are managed with a notice(like wpcomsh does or with a similar UI).

More at

PanosSynetos:
closes #845


xristos3490 (approved):
Great work, @PanosSynetos! From a coding standpoint, it looks good! Let's get the green light from @manospsyx for the plugin list and merge!

PR #849, opened 2022-10-31
Suppress wc payments refund policy inbox

PanosSynetos:
closes #847

PR #844, opened 2022-10-19
Pre-configure Jetpack for Ecommerce Plan users

manospsyx:
closes #834 

Important: 

- Since we have moved Stats under **Jetpack > Stats**, we will need to make a PR in Calypso to remove the sparkline for Ecommerce Plan sites: See https://github.com/Automattic/wp-calypso/commit/50c979cbed8821111d276f8be43bb62f9604badf
- The list of available Jetpack modules is modified for new sites only. This is handled via https://github.com/Automattic/wc-calypso-bridge/pull/844/files#diff-b724d74a3dc5a84c65d2e6c648b01548a2e48e05ba69d28ba5166b2682493e29R71
- Before merging this, rename and update this constant: https://github.com/Automattic/wc-calypso-bridge/pull/844/files#diff-7b47aa048b7d3c28605b4b71d206617f94ab77abe288d55b8e6d288266eae8d5R22
- Also see https://a8c.slack.com/archives/C02GL7EDYT1/p1666179875913849

cc @xristos3490

PR #823, opened 2022-10-04
Default to block-based cart/checkout

PanosSynetos:
closes #822 #848 

* Create WooCommerce pages with a one-time operation
* Delete wc-refund-returns-page inbox note, so it can be recreated with the correct refund page ID
* Set default to block-based cart/checkout
* Display note about the new block-based cart/checkout if the site is active for more than 2 days
* Display wc-refund-returns-page inbox note if the site is active for more than 5 days

PR #853, opened 2022-11-03
Prepare to release 1.9.8

PanosSynetos:
This PR bumps the version to 1.9.8 to release

* #844
* #823 
  * Create WooCommerce pages with a one-time operation
  * Delete wc-refund-returns-page inbox note, so it can be recreated with the correct refund page ID
  * Set default to block-based cart/checkout
  * Display note about the new block-based cart/checkout if the site is active for more than 2 days
  * Display wc-refund-returns-page inbox note if the site is active for more than 5 days
* #827
* #846
* #849

### DIFF
--- a/composer.json
+++ b/composer.json
@@ -1,6 +1,6 @@
 {
   "name": "automattic/wc-calypso-bridge",
-  "version": "v1.9.7",
+  "version": "v1.9.8",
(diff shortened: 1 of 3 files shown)


jimjasson (approved):
Changes look good!

PR #858, opened 2022-11-21
Improved limiting activity panels to Woo Home page

### LINKED ISSUE #857: Improve limiting activity panels to WooCommerce Home
Currently, the activity panel is properly hidden when visiting
- wp-admin pages
- wc-admin pages

and it's properly displayed when you visit directly `admin.php?page=wc-admin`

But if you first visit a wc-admin page (ie `admin.php?page=wc-admin&path=%2Fcustomers`), and then click on the menu item to go to home `admin.php?page=wc-admin` the activity panel is hidden (should be visible).

Related issue: #825

PanosSynetos:
closes #857

### DIFF
--- a/includes/class-wc-calypso-bridge-woocommerce-admin-features.php
+++ b/includes/class-wc-calypso-bridge-woocommerce-admin-features.php
@@ -108,9 +108,20 @@ class WC_Calypso_Bridge_WooCommerce_Admin_Features {
 	public function filter_woocommerce_body_classes_js() {
 		?>
 		<script>
-			( function() {
+			document.addEventListener( 'DOMContentLoaded', function() {
+
+				// Bail out earlier.
+				if ( ! document.body.classList.contains( 'woocommerce_page_wc-admin' ) ) {
+					return;
+				}
+
 				let url = location.href;
-				document.body.addEventListener( 'click', () => {
+				document.body.addEventListener( 'click', ( event ) => {
+
+					// Bail out early.
+					if ( 'A' !== event.target.tagName ) {
+						return;
+					}
 
 					requestAnimationFrame( () => {
 						// URL has changed - let the magic happen.
@@ -130,7 +141,7 @@ class WC_Calypso_Bridge_WooCommerce_Admin_Features {
 					} );
 
 				}, true );
-			} )();
+			} );
 		</script>
 	<?php }
 
--- a/includes/class-wc-calypso-bridge-woocommerce-admin-features.php
+++ b/includes/class-wc-calypso-bridge-woocommerce-admin-features.php
@@ -4,7 +4,7 @@
  *
  * @package WC_Calypso_Bridge/Classes
  * @since   1.0.0
- * @version 1.9.5
+ * @version 1.9.9
  */
 
 defined( 'ABSPATH' ) || exit;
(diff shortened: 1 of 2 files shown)


includes/class-wc-calypso-bridge-woocommerce-admin-features.php
```diff
@@ -97,6 +98,53 @@ public function filter_woocommerce_body_classes( $classes ) {
...
+	 *
+	 * @return void
+	 */
+	public function filter_woocommerce_body_classes_js() {
+		?>
+		<script>
+			document.addEventListener( 'DOMContentLoaded', function() {
+
+				// Bail out earlier.
+				if ( ! document.body.classList.contains( 'woocommerce_page_wc-admin' ) ) {
```
xristos3490, line 114: I believe this won't work as expected (?) 

It falls in the same pitfall as the one that engaged this PR. React will always change the page contents, not the initial body classes, as the user navigates through other react pages.

includes/class-wc-calypso-bridge-woocommerce-admin-features.php
```diff
@@ -97,6 +98,53 @@ public function filter_woocommerce_body_classes( $classes ) {
...
+					return;
+				}
+
+				let url = location.href;
+				document.body.addEventListener( 'click', ( event ) => {
+
+					// Bail out early.
+					if ( 'A' !== event.target.tagName ) {
```
xristos3490, line 122: What about `button`s? Perhaps omitting this change would work better for us. We can't be sure what triggered the page change in the DOM.

xristos3490: My initial thought was to limit this script to the wp-admin context. Thus searching for the `#root` element in the DOM is enough. 

Hint: We might not even need the DOM Ready if the `#root` element exists at first paint; I would also test this perf.

xristos3490 (changes requested):

includes/class-wc-calypso-bridge-woocommerce-admin-features.php
PanosSynetos: As we discussed

> `document.addEventListener( 'DOMContentLoaded', function() {`

Removed as it's not needed

> if ( ! document.body.classList.contains( 'woocommerce_page_wc-admin' ) ) {

Works fine, and it doesn't fall into the same pitfall.

Here's a video demonstrating that it works as expected


https://user-images.githubusercontent.com/2484390/204244914-15ffeb51-c161-4931-82e4-401de74bbee0.mp4

includes/class-wc-calypso-bridge-woocommerce-admin-features.php
PanosSynetos: Indeed, great observation about the buttons

> My initial thought was to limit this script to the wp-admin context. Thus searching for the #root element in the DOM is enough.

Checking for the existence of `woocommerce_page_wc-admin` works as expected, and there is no need for the `#root` element

xristos3490 (approved):
Great work, Panos! Let's merge it!

PanosSynetos:
Cool, thanks - I'll merge it once we're ready to release it

PR #813, opened 2022-09-21
Disable extension activation notices (PB, GC)

PanosSynetos:
closes #812

PR #860, opened 2022-11-24
Suppressed activation notices for Back In Stock Notifications

### LINKED ISSUE #859: Suppress Back In Stock activation notice
As described in the project peapX7-o6-p2 we need to suppress the following notice

image

PanosSynetos:
Closes #859


manospsyx (approved):
Looks fine to me, Pano. Can you double-check that we don't need to add anything on the Extension's side?

PanosSynetos:
Yes, double checked it - I followed the same mechanism/approach that we used for #813

PR #862, opened 2022-11-29
Fix fatal error when setting jetpack defaults for new sites

### LINKED ISSUE #861: Fatal error when setting jetpack defaults one time job
We get a fatal error when a site is created, as at that point WooCommerce is not yet activated

```
Uncaught Error: Class "Automattic\WooCommerce\Admin\WCAdminHelper" not found in 
/wordpress/plugins/wpcomsh/3.4.19/vendor/automattic/
wc-calypso-bridge/includes/class-wc-calypso-bridge-setup.php:245
```

https://github.com/Automattic/wc-calypso-bridge/blob/master/includes/class-wc-calypso-bridge-setup.php#L245

PanosSynetos:
closes #861

### DIFF
--- a/includes/class-wc-calypso-bridge-setup.php
+++ b/includes/class-wc-calypso-bridge-setup.php
@@ -4,7 +4,7 @@
  *
  * @package WC_Calypso_Bridge/Classes
  * @since   1.0.0
- * @version 1.9.8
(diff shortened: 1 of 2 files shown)


xristos3490 (approved):
Looks good, Panos!

PR #865, opened 2022-12-02
Introduce locking to one time operations

### LINKED ISSUE #863: One time operations may run multiple times
### Description

One time operations are running multiple times, for various reasons
- simultaneous ajax calls
- multiple wp-cli activate plugins

Sensitive ones, like the woocommerce create pages, results into multiple cart/checkout pages.

PanosSynetos:
Closes #863


xristos3490 (approved):
Great work, Panos! The DB locking looks 🔥 !

PR #900, opened 2023-01-10
Delete WooCommerce pages before recreating them (new sites only)

### LINKED ISSUE #898: Delete ecommerce related pages before recreating them (new sites only)
For more info see slack thread p1672831281426299-slack-C03Q87XT1QF

TLDR: Tsubaki is creating cart/checkout/refunds pages and then calypso bridge is calling `WC_Install::create_pages();` to create all the related ecomm pages.

All these pages should be removed before recreating them.

PanosSynetos:
This PR closes #898

### DIFF
--- a/includes/class-wc-calypso-bridge-setup.php
+++ b/includes/class-wc-calypso-bridge-setup.php
@@ -195,7 +195,7 @@ class WC_Calypso_Bridge_Setup {
 				 */
 				foreach ( [ 'shop', 'cart', 'my-account', 'checkout', 'refund_returns' ] as $page ) {
 					$page = get_page_by_path( $page, ARRAY_A );
-					if ( $page ) {
+					if ( is_array( $page ) && isset( $page['ID'] ) ) {
 						wp_delete_post( $page['ID'], true );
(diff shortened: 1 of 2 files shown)


includes/class-wc-calypso-bridge-setup.php
```diff
@@ -187,11 +187,27 @@ public function woocommerce_create_pages_callback() {
...
 			try {
+				/*
+				 * Force delete all WooCommerce pages. Some themes create them, and we end up with duplicates.
+				 *
+				 * `My Account` page, has slug `my-account`.
+				 * @see WC_Install::create_pages()
+				 */
+				foreach ( [ 'shop', 'cart', 'my-account', 'checkout', 'refund_returns' ] as $page ) {
+					$page = get_page_by_path( $page, ARRAY_A );
+					if ( $page ) {
```
xristos3490, line 198: Looks good and tested well! I would suggest being a bit more defensive here. Something along these lines:

```suggestion
					if ( is_array( $page ) && isset( $page['ID'] ) ) {
```

PR #908, opened 2023-01-16
Revert the default value of the ecommerce_new_woo_atomic_navigation_enabled filter

xristos3490:
### Specification

This PR brings back the `ecommerce_new_woo_atomic_navigation_enabled` filter as a safety net for easy disabling the Ecommerce navigation and for debugging purposes.


PanosSynetos (approved):
👌 Tested on my ecommerce dev site and works as expected (with the filter returning true and false)

@xristos3490 has also tested on a business dev site

Issue #911, opened 2023-01-17
Free trial site-wide customer facing banner


ilyasfoo:
P2: pdibGW-1iQ-p2
Figma: oRIYBgxHeLDtlVrTjINlxf-fi-646%3A29343&t=tA0y6iDCXjGfC6f6-0

![Image](https://user-images.githubusercontent.com/3747241/212799668-eed5bf29-612b-4540-8631-8af0c97f2f97.png)

Add a permanent, customer facing bottom banner with a CTA to pick a free trial plan. I'd recommend to just enqueue a script and style to bridge.

### Acceptance Criteria

1. Add a customer facing banner that appears on all pages.
2. When clicking on "pick a plan", it should redirect to the plans page.

moon0326:
Hey @ilyasfoo,

Do you have link to the `pick a plan` link?

chihsuan:
> Do you have a link to the pick a plan link?

@moon0326 I think it should be a link to this page:

![Screenshot 2023-01-23 at 11 36 37](https://user-images.githubusercontent.com/4344253/213961840-769252a6-f3e8-4ff1-99b1-91e362cb7848.png)

ilyasfoo:
@moon0326 Apologies for missing this. It goes to the plans page as @chihsuan mentioned.

PR #917, opened 2023-01-19
Add plan picker banner for free trial

moon0326:
Fixes https://github.com/Automattic/wc-calypso-bridge/issues/911

### Changes

1. Added free trial plan check to load frontend class conditionally.  The check logic was suggested in p1674072849311499-slack-C0Q664T29

2. Added a plan picker banner at the bottom.

### Test Instructions

1. Make sure to copy over the latest changes in [woocmmerce-calypso-bridge-helper.php](https://gist.github.com/moon0326/cac46c70a2cee81b61faef517fef7178) and the plugin is activated.
2. Clone this branch
3. Navigate to the frontend of your local env. Confirm the banner is not rendered.
4. Change [this line](https://gist.github.com/moon0326/cac46c70a2cee81b61faef517fef7178#file-a-woocommerce-calypso-bridge-helper-php-L33) to false in your local env.
5. Go back to the frontend of your local env and confirm the banner is rendered.


wc-calypso-bridge.php
```diff
@@ -55,6 +55,21 @@ function wc_calypso_bridge_is_ecommerce_plan() {
 	}
 }
 
+if( ! function_exists( 'wc_calypso_bridge_can_install_plugins' )) {
+	/**
+	 * Returns true if a site has install-plugins feature
+	 *
+	 * @return bool
+	 */
+	function wc_calypso_bridge_can_install_plugins() {
```
ilyasfoo, line 64: Great work, @moon0326! Thanks for finding out how to detect free trial plan. I feel these feature detection logic could be refactored into a better structure, but probably not in this PR

wc-calypso-bridge.php
```diff
@@ -100,3 +115,16 @@ function wc_calypso_bridge_init() {
 require_once dirname( __FILE__ ) . '/class-wc-calypso-bridge-frontend.php';
 
 require_once dirname( __FILE__ ) . '/includes/class-wc-calypso-bridge-woocommerce-admin-features.php';
+
+/**
+ * At this point, user is confirmed to have ECOM plan.
+ * Include Free Trial classes if the user is missing install-plugins feature.
+ *
+ * Free trial plan is a subset of ECOM plan without some features.
+ * One of them is install-plguins.
```
ilyasfoo, line 124:
```suggestion
 * One of them is install-plugins.
```

wc-calypso-bridge.php
ilyasfoo: Also, I think the detection would be widely used, I think it's better to name it as `wc_calypso_bridge_is_free_trial_plan`?

moon0326: > Also, I think the detection would be widely used, I think it's better to name it as wc_calypso_bridge_is_free_trial_plan?

That's a good point. My concern was that feature checking requires a [DB access](https://github.com/Automattic/wpcomsh/blob/trunk/wpcom-features/functions-wpcom-features.php#L57) as far as I understand. I wanted to make sure we only use it when necessary. 

I think we can make a class that caches the result for the current request so we can safely call it multiple times, but that refactor can wait for a little 😃

moon0326:
@Automattic/somewherewarm Hi team ✋

I noticed you are working on peapX7-1pT-p2 to refactor/optimize plan detection logic.

I just wanted to cross-check with you on this [logic](https://github.com/Automattic/wc-calypso-bridge/pull/917/files#diff-01916d5466e0b3deaed366497e01f4f72f8af94fdabc2a3fe1cc068d5af55eb6R128) to detect free trial.

Any feedback would be appreciated 🙏

wc-calypso-bridge.php
ilyasfoo: >I think we can make a class that caches the result for the current request so we can safely call it multiple times, but that refactor can wait a little 😃 I think we can work on that in a separate PR. What do you think?

I think that can work!

includes/class-wc-calypso-bridge-ecom-free-trial-frontend.php
```diff
@@ -0,0 +1,56 @@
...
+	 */
+	final public static function get_instance() {
+		if ( is_null( self::$instance ) ) {
+			self::$instance = new self();
+		}
+
+		return self::$instance;
+	}
+	
+	public function __construct(){
```
xristos3490, line 32: Should we add this notice for shop managers and administrators, only? 🤔

xristos3490 (commented):
Thanks for getting this workstream started, @moon0326!

> I think we can make a class that caches the result for the current request so we can safely call it multiple times, but that refactor can wait a little 

FYI, that's something we plan on working on during this refactor!

> I noticed you are working on peapX7-1pT-p2 to refactor/optimize plan detection logic. 

That's true! But we are at the discovery phase, so bringing this PR now is okay for us.

> I just wanted to cross-check with you on this [logic](https://github.com/Automattic/wc-calypso-bridge/pull/917/files#diff-01916d5466e0b3deaed366497e01f4f72f8af94fdabc2a3fe1cc068d5af55eb6R128) to detect free trial. It's checking for install-plugins feature after confirming ECOM plan.

Given that we'll be adding a runtime cache for these checks, it would be best to have each function work independently. But, for now, this works just fine! Also, looking at the Diff, it is clear that the INSTALL_PLUGINS is a good path for determining the free trial plan! I've added some notes on this for the upcoming refactor! 

I've just commented on whether we should add this notice, or perhaps use a different copy for non-admins. Thoughts?

includes/class-wc-calypso-bridge-ecom-free-trial-frontend.php
moon0326: Thank you for the detailed feedback @xristos3490 👍 

> Should we add this notice for shop managers and administrators, only?

That's a good point! I think the notice should be for just shop managers and administrators.

I'll confirm with design team and make a change 👍

moon0326: Updated in https://github.com/Automattic/wc-calypso-bridge/pull/917/commits/e57816c1ccfa263b0984fce9455c4dfc516c8c0e

moon0326:
Hey @xristos3490 ✋ 

We're aiming to launch some of the free trial changes on Feb 20th (see [comment](https://ghidorahp2.wordpress.com/2023/02/13/cycle-33-kick-off-async/#comment-1110)). We need to have free trial detection for those issues.

It sounds like we're deploying #926 early next week, according to this [comment](https://github.com/Automattic/wc-calypso-bridge/pull/937#issuecomment-1423959642).

What do you think about merging changes in this PR and refactoring it to use the functions in your PR later?

PanosSynetos:
@moon0326 the plan is to merge #926 today (in a couple of hours) and ask the atomic team to deploy it as soon as possible so we have enough time to fix any issues that might arise.

All manual testing so far, look good.

xristos3490:
> What do you think about merging changes in this PR and refactoring it to use the functions in your PR later?

Done! Let me know what you think! 

Regarding the free-trial banner, while testing, I noticed that the banner falls into the store-wide notice. I would opt for fixing this, as the store-wide notice is enabled by default.

See:
notice


Another solution might be to replace the existing store-wide notice functionality and replace the message. But that would require making the "pick a plan" link into a button instead of inline text. Thoughts?

xristos3490:
Also, this store-wide message design needs to be fixed, right? A design input here may help!

See

moon0326:
> > What do you think about merging changes in this PR and refactoring it to use the functions in your PR later?
> 
> Done! Let me know what you think!

Thank you so much! Looks fantastic! I'll make changes to use the new functions.

> Another solution might be to replace the existing store-wide notice functionality and replace the message. But that would require making the "pick a plan" link into a button instead of inline text. Thoughts?

Good catch! I'll make a change to replace the store-wide notice text.

For the button change, I think we need to confirm with @verofasulo

xristos3490:
> Should we disable store-wide notice in free trial instead

There is an option during onboarding that lets the merchant write a custom store notice here:

notice

I would suggest keeping the experience intact, and perhaps coming up with a solution that stacks the notices together.
I'd love @verofasulo's input here as well!

moon0326:
> > Should we disable store-wide notice in free trial instead
> 
> There is an option during onboarding that lets the merchant write a custom store notice here:
> 
> notice
> 
> I would suggest keeping the experience intact, and perhaps coming up with a solution that stacks the notices together. I'd love @verofasulo's input here as well!

Sounds good. Let's wait for @verofasulo's input.

FlorianBru:
I wouldn't recommend stacking both notifications. Is there an option to move the store notice to the top?
If not, I suggest moving the side-wide free trial banner to the top.
@moon0326 @xristos3490

moon0326:
> Hello there ✋
> 
> I wouldn't recommend stacking both notifications. Is there an option to move the store notice to the top? If not, I suggest moving the side-wide free trial banner to the top.
> 
> @moon0326 @xristos3490

Thank you for the feedback 👍 That's a good idea.

moon0326:
@xristos3490 @FlorianBru 

So the store notice can appear at the top or bottom depending on the activated theme.

I've made this [change](https://github.com/Automattic/wc-calypso-bridge/pull/917/commits/472ae03443ed5f4ba8fd88183d40990e3b470e1c) to place the store notice at the top all the time regardless of activated theme for ONLY admin. Customers won't see the difference since we don't show the free trial banner to them.

**Customers view:**
Screen Shot 2023-02-22 at 12 28 00 PM

**Admin view**
Screen Shot 2023-02-22 at 12 27 52 PM

FlorianBru:
Hey @moon0326, do you think it's also possible to move the free trial notification to the top and leave the notice at the bottom? This way we would stay consistent regarding the notice positioning.

includes/class-wc-calypso-bridge-frontend-free-trial.php
```diff
@@ -0,0 +1,70 @@
...
+	final public static function get_instance() {
+		if ( is_null( self::$instance ) ) {
+			self::$instance = new self();
+		}
+
+		return self::$instance;
+	}
+	
+	public function __construct() {
+
```
ilyasfoo, line 33: Nit: Extra space
```suggestion
```

class-wc-calypso-bridge.php
```diff
@@ -123,6 +123,8 @@ public function includes() {
 		require_once WC_CALYPSO_BRIDGE_PLUGIN_PATH . '/includes/class-wc-calypso-bridge-free-trial-payment-task.php';
 		require_once WC_CALYPSO_BRIDGE_PLUGIN_PATH . '/includes/free-trial/class-wc-calypso-bridge-free-trial-orders-changes.php';
 		require_once WC_CALYPSO_BRIDGE_PLUGIN_PATH . '/includes/free-trial/class-wc-calypso-bridge-free-trial-wc-payments.php';
+		require_once WC_CALYPSO_BRIDGE_PLUGIN_PATH . '/includes/class-wc-calypso-bridge-frontend-free-trial.php';
+
```
ilyasfoo, line 127: Nit: Extra space
```suggestion
```

class-wc-calypso-bridge.php
```diff
@@ -123,6 +123,8 @@ public function includes() {
 		require_once WC_CALYPSO_BRIDGE_PLUGIN_PATH . '/includes/class-wc-calypso-bridge-free-trial-payment-task.php';
 		require_once WC_CALYPSO_BRIDGE_PLUGIN_PATH . '/includes/free-trial/class-wc-calypso-bridge-free-trial-orders-changes.php';
 		require_once WC_CALYPSO_BRIDGE_PLUGIN_PATH . '/includes/free-trial/class-wc-calypso-bridge-free-trial-wc-payments.php';
+		require_once WC_CALYPSO_BRIDGE_PLUGIN_PATH . '/includes/class-wc-calypso-bridge-frontend-free-trial.php';
```
ilyasfoo, line 126: Nit: Do you think this file could have a more representative naming? I think we could either have a generic controller that does multiple simple modifications for free trial, or specific modifications just for the customer screen.

ilyasfoo (approved):
Tested well, @moon0326! Nice work. Some nitpicky comments, but overall LGTM! 🚢

class-wc-calypso-bridge.php
moon0326: @ilyasfoo Good suggestion. Um..I can't think of a good naming since we don't have many use cases yet. I'm considering moving this class into `free-trial` and treating it like any other free-trial classes. What do you think?

ilyasfoo: @moon0326 Sounds good!

moon0326: @ilyasfoo Updated in https://github.com/Automattic/wc-calypso-bridge/pull/917/commits/26ac35a7fd469ef905cd3dc291600a2989771403

ilyasfoo: @moon0326  Excellent, lgtm!

FlorianBru:
Hey @moon0326 , thanks for positioning the trial notification at the top. There is still something we can do here to enhance the experience. The banner should behave like the navigation bar at the top of the screen.

At the moment the trial notification is hiding parts of the actual site, this should be changed.

moon0326:
> Hey @moon0326 , thanks for positioning the trial notification at the top. There is still something we can do here to enhance the experience. The banner should behave like the navigation bar at the top of the screen.
> 
> At the moment the trial notification is hiding parts of the actual site, this should be changed.

Thank you for the feedback 👍 

Actually, the trial notification is a sticky banner just like the admin bar, but I can see what you mean. 

I think we can add margin to the body element to push down the main content when the trial notification is shown.

What do you think?

**Before:**
Screen Shot 2023-03-01 at 11 32 36 AM

**After:**
Screen Shot 2023-03-01 at 11 32 29 AM

xristos3490:
> I think we can add margin to the body element to push down the main content when the trial notification is shown.

Sounds like a good idea! We need to ensure that this works with Tsubaki theme too.

Also, this part:
```css
/**
* Store notice can be at the top or bottom depending on the activated theme.
* Force it to be at the bottom all the time to place the free trial banner at the bottom.
*/
body.has-free-trial-plan-picker .woocommerce-store-notice {
	top: auto !important;
}

**Comment by FlorianBru** [comment]
>I think we can add margin to the body element to push down the main content when the trial notification is shown.

@moon0326 Yep, I also feel this could do the trick.

File: includes/free-trial/class-wc-calypso-bridge-free-trial-plan-picker-banner.php [hunk 7]
```diff
@@ -0,0 +1,69 @@
...
+	/**
+	 * @return void
+	 */
+	public function add_plan_picker_banner() {
+		$status      = new \Automattic\Jetpack\Status();
+		$site_suffix = $status->get_site_suffix();
+		$link = sprintf( "https://wordpress.com/plans/%s", $site_suffix );
+
+		$text = sprintf( __("
+			At the moment you are the only one who can see your store. To let everyone see your store, you simply need to&nbsp;<a href='%s'>pick a plan</a>.
```
ilyasfoo, line 59: I just spotted a different copy in figma lD0gTB4IqGsmSmsnjZNaP6-fi-31%3A8549 
```suggestion
			At the moment you are the only one who can see your store. To make your store available to everyone, please&nbsp;<a href='%s'>upgrade to a paid plan</a>.
```

moon0326: Thank you!

moon0326:
@xristos3490 @FlorianBru Updated in https://github.com/Automattic/wc-calypso-bridge/pull/917/commits/cc4bf4f14312d322aa161d513c1dfef57d174e7f

I've also tested it on WPCOM with a few different themes. Here are screenshots for your reference.

**Bibimbap**
Screen Shot 2023-03-02 at 1 14 01 PM

**Calyx**
Screen Shot 2023-03-02 at 1 12 35 PM

**Tsubaki**
Screen Shot 2023-03-02 at 1 11 29 PM

adrianduffell (approved):
Tested perfect, thanks @moon0326 ! 🚀

adrianduffell:
Merging on behalf of @moon0326

PR #921, opened 2023-01-24
Arrange menu order for Mailpoet and Automatewoo

xristos3490:
### Specification

This PR removes MailPoet from the list of managed plugins and arranges the new Mailpoet and AutomateWoo menu items into the Woo Navigation.

### DIFF
--- a/includes/class-wc-calypso-bridge-ecommerce-admin-menu.php
+++ b/includes/class-wc-calypso-bridge-ecommerce-admin-menu.php
@@ -4,7 +4,7 @@
  * Class Ecommerce_Atomic_Admin_Menu.
  *
  * @since   1.9.8
- * @version 1.9.17
+ * @version 1.9.18
  *
  * The admin menu controller for Ecommerce WoA sites.
  */
--- a/includes/class-wc-calypso-bridge-ecommerce-admin-menu.php
+++ b/includes/class-wc-calypso-bridge-ecommerce-admin-menu.php
@@ -46,11 +46,10 @@ class Ecommerce_Atomic_Admin_Menu extends \Automattic\Jetpack\Dashboard_Customiz
 	 * Groups WooCommerce items.
 	 */
 	public static function menu_order( $menu_order ) {
-
 		// Initialize our custom order array.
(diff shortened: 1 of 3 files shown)


includes/class-wc-calypso-bridge-ecommerce-admin-menu.php
```diff
@@ -46,11 +46,10 @@ public function __construct() {
 	 * Groups WooCommerce items.
 	 */
 	public static function menu_order( $menu_order ) {
-
 		// Initialize our custom order array.
 		$woocommerce_menu_order = array();
 
-		// Get the index of our managed pages.
+		// Get the index of our managed pages for core.
 		$orders_index              = array_search( 'edit.php?post_type=shop_order', $menu_order, true );
```
PanosSynetos, line 53: Do we need the code indexes? The variables are unused.

includes/class-wc-calypso-bridge-ecommerce-admin-menu.php
```diff
@@ -59,8 +58,12 @@ public static function menu_order( $menu_order ) {
 		$extensions_index          = array_search( 'woocommerce', $menu_order, true );
 		$woocommerce_sep_index     = array_search( 'separator-woocommerce', $menu_order, true );
 		$woocommerce_sep_top_index = array_search( 'wc-calypso-bridge-separator-top', $menu_order, true );
+
+		// Get the index of our managed pages for integrations.
 		$payments_connect_index    = array_search( 'wc-admin&path=/payments/connect', $menu_order, true );
```
PanosSynetos, line 63: Do we care about the indexes in general in this function? Isn’t `in_array` enough? (Unless `array_search` is more performant)

PanosSynetos (commented):
I tested and it works as expected :) (wouldn't expect anything else from you :) )

I do have two questions though, let me know when you get a chance so I can release (tomorrow - as we're too close to the atomic deployment and we there's no need to rush)

PanosSynetos (approved):
👌 Manually tested and works as expected

PR #925, opened 2023-01-26
Disable ecommerce menu and relevant JS fixes when SSO is disabled

### LINKED ISSUE #923: eCommerce Navigation: Home redirect not working when Jetpack SSO is off
## Quick summary

When Jetpack SSO is deactivated and a customer has not authenticated WP Admin log in for their site, then the redirection to WooCommerce Home does not work and it displays a WP Admin log in page. If the Coming Soon mode is enabled, then it actually redirects to a Coming Soon page. 

## Steps to reproduce
1. Make sure you do not have a WP Admin log in cookie for your eCommerce plan site in your browser.
2. Deactivate Jetpack SSO via Calypso by going to Settings -> Security. 
3. Click a WP Admin link in the left sidebar.

## What you expected to happen
Not quite sure. The possible solution is to redirect to Calypso Home instead of WooCommerce Home, when Jetpack SSO is deactivated. Though, a customer should not be redirected to a log in page. 

## What actually happened

**Coming Soon Enabled**
Customer is redirected to a Coming Soon page and no WP Admin log in is shown.
*This seems to depend on a cookie if I am correct, so might not always show.* 

**Coming Soon Disabled**
Customer is redirected to a WP Admin log in page. 

### Context
p1674197330375039-wpcom-support

### Platform (Simple, Atomic, or both?)
Atomic

### Reproducibility
Consistent

### Severity
Some (< 50%)

### Available workarounds?
Yes & No. If you know the direct URLs for Calypso pages, you can access them. However, customer cannot normally manage their site in Calypso, meaning they cannot access vital pages like Hosting Configuration. 

@Automattic/somewherewarm

xristos3490:
Fixes #923 

### Specification

This PR adds detection for Jetpack SSO to disable all menu fixes in the JS. 

There are two issues here:
- When navigating to Calypso home, the "My Home" menu item isn't selected as active: 

Screenshot 2023-07-07 at 2 59 07 PM

- When navigating to the WC Admin home in wp-admin, the "My Home" and the "WooCommerce" menus are both active.

Screenshot 2023-07-07 at 3 00 48 PM

### Validation

Steps for testing:
- Navigate to **Settings > Security** and disable the "Allow users to log in to this site using WordPress.com accounts" option.
- (In a new tab) Navigate to WooCommerce Home. 
- Notice that the menu is working and is highlighted as expected based on the screenshot above.
- Now navigate to `wordpress.com/home/[siteslug_with_sso_disabled]` 
- Notice that the "My home" is highlighted as expected based on the screenshot above.
- Also, notice that the ecommerce menu is not applied to any context.


class-wc-calypso-bridge-shared.php
```diff
@@ -111,7 +111,7 @@ public function add_scripts() {
 		$params      = array(
 			'isEcommercePlan'              => (bool) wc_calypso_bridge_has_ecommerce_features(),
 			'isEcommercePlanTrial'         => (bool) wc_calypso_bridge_is_ecommerce_trial_plan(), // This is true for ecommerce trial only.
-			'isWooNavigationEnabled'       => (bool) apply_filters( 'ecommerce_new_woo_atomic_navigation_enabled', 'yes' === get_option( 'wooexpress_navigation_enabled', 'yes' ) ),
+			'isWooNavigationEnabled'       => (bool) apply_filters( 'ecommerce_new_woo_atomic_navigation_enabled', 'yes' === get_option( 'wooexpress_navigation_enabled', 'yes' ) ) && \Jetpack::is_module_active( 'sso' ),
```
ilyasfoo, line 114: We should check for existence of `Jetpack` class

ilyasfoo (changes requested):
Overall tested well! I have a comment about being more defensive with assuming Jetpack class

class-wc-calypso-bridge-shared.php
xristos3490: Thanks for the review, @ilyasfoo! There are additional instances where the Jetpack class is presumed, but I have implemented defensive checks in this section. We can continue to evaluate any remaining assumptions as we move forward.

ilyasfoo (approved):
@xristos3490  Awesome, looks great! 🚢

PR #926, opened 2023-01-27
Refactor Ecommerce project and implement plans detection controller

### LINKED ISSUE #932: Refactor to clear debt and support multiple plans
The goal of this epic is to:

1. Clean up and refactor the Calypso Bridge codebase to (i) support multiple plans; and (ii) remove unused/legacy code. A separate project with be created to communicate the work that needs to be done in detail, and resolve open questions.

2. Review all improvements shipped to the Ecommerce Plan since September 2022 and cherry pick those relevant to the (i) Business Plan with WooCommerce, Free Trial and (iii) Small Ecommerce Plan in Calypso Bridge, Calypso, and Jetpack. WooCommerce stores on the Business Plan will only inherit a subset of the merchant experience optimizations that we have shipped, and none of the plugins/Extensions bundled in Hosted Woo plans.

xristos3490:
### Specification

This PR aims to refactor how _The Bridge_ works to accommodate multiple plans. The new global functions for detecting a 
plan are:
- `wc_calypso_bridge_has_ecommerce_features()` 
This will return `true` for all ecommerce-related plans such as the Ecommerce and Ecommerce Trial.

- `wc_calypso_bridge_is_ecommerce_plan()` 
This one existed before. Returns `true` if site has the specific Ecommerce plan. e.g., `false` for the trial plan.

- `wc_calypso_bridge_is_ecommerce_trial_plan()` 
Returns `true` if the site has an Ecommerce Trial plan.

- `wc_calypso_bridge_is_business_plan()` 
Returns `true` if the site has a business plan.

### Updates to Ecommerce sites

No updates or changes to existing or new ecommerce sites are expected from this PR.

### Updates to Business sites

- The "Coupon moved" notice will be deprecated and removed (if exists).
- A [bunch of inbox notes](https://github.com/Automattic/wc-calypso-bridge/pull/926/files#diff-cf4261fea98ea63a9d2123c45a5d479d59f957e496b6f5bc1baf8e0708db38b8R56-R81) will be removed from WC Home.
- Force enable woocommerce tracker.

### Validation

Using three sites, one business, one ecom trial, and one ecommerce, notice the following:
- Jetpack active modules list should be the same in business sites.
- The Cart/Checkout pages should be block-based in ecommerce sites and the legacy ones in new and existing business sites.
- The Onboarding Setup Wizard should show up on new business sites.
- The WooCommerce Navigation feature should remain active in existing business sites that already use this feature.
- All sites include the "Add a domain" and "Launch your store" tasks in the WC home setup list.

closes #932


class-wc-calypso-bridge-dotcom-features.php
```diff
@@ -0,0 +1,110 @@
...
+	protected static $is_ecommerce_trial_plan = null;
+
+	/**
+	 * Determine if site is Ecommerce and cache it.
+	 *
+	 * @var bool
+	 */
+	public static function is_ecommerce_plan() {
+		if ( is_null( self::$is_ecommerce_plan ) ) {
+			self::$is_ecommerce_plan = (bool) function_exists( 'wpcom_site_has_feature' ) ? wpcom_site_has_feature( \WPCOM_Features::ECOMMERCE_MANAGED_PLUGINS ) : false;
```
daledupreez, line 79: A few thoughts here:
1. Both the eCommerce trial and the main eCommerce plan have this feature, so either the logic or the naming needs to be changed.
2. For the case where the function doesn't exist yet, we have no way of knowing that the value was set before `wpcom_site_has_feature()` was defined, and the value should be checked. (This is an extreme edge case, but I want to flag it -- there may well be cases where `wpcom_site_has_feature()` is never defined, so we do need to pick our battles.)
3. This is a nit, but there is a _lot_ of logic on this one line! Pushing this to a multiline condition would make it easier for me to read and understand.
```suggestion
			self::$is_ecommerce_plan = (bool) function_exists( 'wpcom_site_has_feature' )
				? wpcom_site_has_feature( \WPCOM_Features::ECOMMERCE_MANAGED_PLUGINS )
				: false;
```

class-wc-calypso-bridge.php
```diff
@@ -46,421 +41,172 @@ class WC_Calypso_Bridge {
...
 
 	/**
-	 * Remove WooCommerce footer text
+	 * Loads language files for the plugin.
+	 *
+	 * @since 2.0.0
 	 */
-	public function remove_woocommerce_core_footer_text() {
-		add_filter( 'woocommerce_display_admin_footer_text', '__return_false' );
+	public function load_transalation() {
```
daledupreez, line 189: Nit: this should be `load_translation()`.
```suggestion
	public function load_translation() {
```

includes/class-wc-calypso-bridge-addons.php
```diff
@@ -62,7 +76,9 @@ public function addons_menu() {
 	 * Init the addons page.
 	 */
 	public function addons_page() {
+		require_once WC_CALYSPO_BRIDGE_PLUGIN_PATH . '/includes/class-wc-calypso-bridge-addons-screen.php';
```
daledupreez, line 79: Typo in the constant -- this probably applies in other places too.
```suggestion
		require_once WC_CALYPSO_BRIDGE_PLUGIN_PATH . '/includes/class-wc-calypso-bridge-addons-screen.php';
```

class-wc-calypso-bridge-dotcom-features.php
```diff
@@ -0,0 +1,110 @@
...
+ * WC Calypso Bridge DotCom Features class.
+ */
+class WC_Calypso_Bridge_DotCom_Features {
+
+	/**
+	 * Is Ecommerce plan.
+	 *
+	 * @var bool
+	 */
+	protected static $is_ecommerce_plan = null;
```
daledupreez, line 56: Might it be cleaner to structure these variables and the related checks in a way that resembles the `WPCOM_Features` approach? We could even push the primary logic directly into `WPCOM_Features` so these are direct features in the system.

I ask this because `is_ecommerce_plan()` is already ambiguous given the existence of both the eCommerce and trial plans, and will only become more so as we introduce WooExpress plans. It won't be clear to future developers if `is_ecommerce_plan()` would reasonably include those plans or not.

Perhaps we can do this using naming that is more self-descriptive, like `$has_ecommerce_features` or `$has_ecommerce_optimizations`.
```suggestion
	protected static $has_ecommerce_features = null;
	protected static $has_ecommerce_optimizations = null;
```

class-wc-calypso-bridge-dotcom-features.php
```diff
@@ -0,0 +1,110 @@
...
+	}
+
+	/**
+	 * Determine if site is Ecommerce Trial and cache it.
+	 *
+	 * @var bool
+	 */
+	public static function is_ecommerce_trial_plan() {
+		if ( is_null( self::$is_ecommerce_trial_plan ) ) {
+			self::$is_ecommerce_trial_plan = self::is_ecommerce_plan() && (bool) function_exists( 'wpcom_site_has_feature' ) ? wpcom_site_has_feature( \WPCOM_Features::INSTALL_PLUGINS ) : false;
```
daledupreez, line 92: This check feels incorrect to me: this should be true if we have an eCommerce (or the eCommerce features) and we _don't_ have permissions to install a plugin.

I think we probably need a dedicated feature to specifically flag that a site is on a free trial. But I think that's something we can handle in a follow-up.

class-wc-calypso-bridge-dotcom-features.php
```diff
@@ -0,0 +1,110 @@
...
+	}
+
+	/**
+	 * Determine if site is Business and cache it.
+	 *
+	 * @var bool
+	 */
+	public static function is_business_plan() {
+		if ( is_null( self::$is_business_plan ) ) {
+			self::$is_business_plan = (bool) function_exists( 'wpcom_site_has_feature' ) ? wpcom_site_has_feature( \WPCOM_Features::CONCIERGE_BUSINESS ) : false;
```
daledupreez, line 105: I think this one should work.

daledupreez (commented):
I've dropped some comments with a focus on the feature checks, and I've captured my specific concerns in those comments.

More broadly, I am super-concerned at how large this PR is. It feels like it's going to be really hard to review and test the changes effectively (even without the hassle of getting it to dev sites with the respective plans) given the sheer size and scope of the changes.

Might it be possible to break all the changes down into some smaller, more focused PRs? I know this will involve dev work, but it may be worth it to streamline the testing and give us more confidence about each change. We can defer actual deployment until we have the changes in place, but the actual process of getting there may be more complicated if we only have this one PR to manage all feedback and testing.

manospsyx:
> More broadly, I am super-concerned at how large this PR is. It feels like it's going to be really hard to review and test the changes effectively (even without the hassle of getting it to dev sites with the respective plans) given the sheer size and scope of the changes.

> Might it be possible to break all the changes down into some smaller, more focused PRs? I know this will involve dev work, but it may be worth it to streamline the testing and give us more confidence about each change. We can defer actual deployment until we have the changes in place, but the actual process of getting there may be more complicated if we only have this one PR to manage all feedback and testing.

@daledupreez with the Woo meetup ahead, and limited headroom to break this down, I think we need to make a conscious decision around what we will sacrifice. As they say -- "Speed, quality/risk, cost, pick 2".

I would trust Chris with shipping high quality code, always, and therefore I am willing to optimize for speed and lower cost here, but take some quality risk. That risk will be mitigated by releasing the code on the week after the Woo meetup -- which means that if there are any quirks, we'll have them ironed out in time for launch.

If we delay and ship this at launch, we'll be still moving some (but less) risk further down. I don't think that breaking this up into more readable/reviewable chunks will mitigate all of that risk -- in other words, I think it's a better tradeoff to launch earlier and be as proactive as we can. 

If we could delay the Free Trial launch, I'd be happy to take a more conservative route, but if that's not on the table, we have to make some tradeoffs here and take the safety of "testing this in production" ahead of time. It's really unfortunate that we have zero unit or e2e tests here to be more confident; I have called that out previously. As much as I'd love to make every codebase follow the standards I've had for ours, I don't have the capacity to do it -- so something's gotta give if we want to move quickly :(

class-wc-calypso-bridge-dotcom-features.php
xristos3490: That's correct! The statement is missing a `!` operator!

> I think we probably need a dedicated feature to specifically flag that a site is on a free trial. But I think that's something we can handle in a follow-up.

Yup! I was thinking about this, too. For now, I've implemented the check logic based on the discussion at p1674072849311499-slack-C0Q664T29

class-wc-calypso-bridge-dotcom-features.php
xristos3490: This seems tricky! 

> Might it be cleaner to structure these variables and the related checks in a way that resembles the WPCOM_Features approach? 

Could you elaborate on this? Are you thinking of a features-array that constructs different combinations? If so, we could revisit this class as we add plans+logic into the system. 

For now, I've added a static class helper to help us with the runtime caching and exposed the checks in global functions. I thought we could revisit the static class in the long term and always have the global functions as a way to interact with it. 

> I ask this because is_ecommerce_plan() is already ambiguous given the existence of both the eCommerce and trial plans, and will only become more so as we introduce WooExpress plans. It won't be clear to future developers if is_ecommerce_plan() would reasonably include those plans or not.

That's a great idea! The `wc_calypso_bridge_is_ecommerce_plan()` naming suggests this is a strict boolean check. I really liked the idea of renaming it to `wc_calypso_bridge_has_ecommerce_features()`. The only drawback is that the `wc_calypso_bridge_is_ecommerce_plan()` exists from the start. 

So moving to a different plan checking will require us to either 
- Introduce the new `wc_calypso_bridge_has_ecommerce_features()` and deprecate the previous one, or;
- Keep them both and make the `wc_calypso_bridge_is_ecommerce_plan()` strict. 

For example (pseudocode for demonstration):
```php
function wc_calypso_bridge_is_ecommerce_plan() {
    has_ecommerce_features && can_install_plugins;
}
```

Thoughts?

xristos3490: I would also love to know more about the race condition where the `wpcom_site_has_feature()` won't be available! I think the best approach we can take is to do a sanity check before spinning up this entire project and bail early if we can't make any assumptions about the plan used. 

For example, we could add an early exit in the plugin's main file:
```php
// [...]
if ( ! function_exists( 'wpcom_site_has_feature' ) {
    return;
}

// The Bridge Main Controller.
require_once WC_CALYPSO_BRIDGE_PLUGIN_PATH . '/class-wc-calypso-bridge-dotcom-features.php';
require_once WC_CALYPSO_BRIDGE_PLUGIN_PATH . '/class-wc-calypso-bridge.php';
```

PR #933, opened 2023-02-02
add: homescreen free trial banner

### LINKED ISSUE #927: Free trial extension: Homescreen header banner
## User Story

> As a free trial user, 
> I want to see a banner in homescreen,
> so that I'm informed that I'm on a free trial and
> so that I can go to the relevant page when I need more information.

## Acceptance Criteria

```
Feature: Header banner
Given the user is on a free trial plan
When the user views WooCommerce homescreen
Then show header banner
```

```
Feature: Header banner CTA click
Given the user is on a free trial plan
When the user clicks on "Learn more"
Then navigate to plans page ex: "https://wordpress.com/plans/sitenamehere"
```

```
Feature: Header banner dismiss
Given the user is on a free trial plan
When the user clicks on close `x` button
Then dismiss the header permanently (to be confirmed oRIYBgxHeLDtlVrTjINlxf-fi-646:29343)
```

## Other details

Figma: oRIYBgxHeLDtlVrTjINlxf-fi-646%3A29343

Image

Related extensibility PR: https://github.com/woocommerce/woocommerce/pull/36467

rjchow:
Fixes #927

### Changes
- Added the free trial informational banner in the homescreen header slotfill
- "Learn More" link is populated using ` window.wcCalypsoBridge.siteSlug `, not sure if this is the most appropriate way?
- Dismissal history is stored in options

### Testing Instructions

1. Make sure to copy over the latest changes in [woocmmerce-calypso-bridge-helper.php](https://gist.github.com/moon0326/cac46c70a2cee81b61faef517fef7178) and the plugin is activated.
2. Clone this branch
3. Navigate to the frontend of your local env. Confirm the banner is rendered.
4. Check that "Learn More" button correctly brings you to wordpress.com plan selection
5. Check that dismissal works and is persisted

https://user-images.githubusercontent.com/27843274/216316747-667bf96b-d96c-4627-8b71-ba0e88a72b6a.mp4

### DIFF
--- a/src/index.js
+++ b/src/index.js
@@ -17,6 +17,7 @@ import { PaymentGatewaySuggestions } from './payment-gateway-suggestions';
 import { Tax } from './free-trial/tax';
 import { TaskListCompletedHeaderFill } from './task-completion/fill.tsx';
 import './index.scss';
+import { CalypsoBridgeHomescreenBanner } from './homescreen-banner';
 
 wcNavFilterRootUrl();
 
@@ -97,10 +98,13 @@ if ( !! window.wcCalypsoBridge.isEcommercePlanTrial ) {
 		),
 	} );
 
+	registerPlugin( 'wc-calypso-bridge-homescreen-slotfill-banner', {
+		render: CalypsoBridgeHomescreenBanner,
+		scope: 'woocommerce-admin',
+	} );
 }
 
 if ( !! window.wcCalypsoBridge.isEcommercePlan ) {
-
 	// Filter wc admin pages.
 	addFilter( 'woocommerce_admin_pages_list', 'wc-calypso-bridge', ( pages ) => {
 
--- a/src/homescreen-banner/index.js
+++ b/src/homescreen-banner/index.js
@@ -1,0 +1,69 @@
+/**
+ * External dependencies
+ */
+import { Fill, Card, Button } from '@wordpress/components';
+import { __ } from '@wordpress/i18n';
+import { OPTIONS_STORE_NAME } from '@woocommerce/data';
+import { useSelect, useDispatch } from '@wordpress/data';
+/**
+ * Internal dependencies
+ */
+import './style.scss';
+export const WC_CALYPSO_BRIDGE_HOMESCREEN_BANNER_HIDDEN =
+	'wc_calypso_bridge_homescreen_banner_hidden';
+
+export const CalypsoBridgeHomescreenBanner = () => {
+	const { updateOptions } = useDispatch( OPTIONS_STORE_NAME );
+	const { isModalHidden } = useSelect( ( select ) => {
+		const { getOption, hasFinishedResolution } =
+			select( OPTIONS_STORE_NAME );
+
+		return {
+			isModalHidden:
+				getOption( WC_CALYPSO_BRIDGE_HOMESCREEN_BANNER_HIDDEN ) ===
+					'yes' ||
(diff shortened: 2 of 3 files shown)


src/index.js
```diff
@@ -34,18 +35,23 @@ if ( !! window.wcCalypsoBridge.isEcommercePlan ) {
 		),
 	} );
 
+	registerPlugin( 'wc-calypso-bridge-homescreen-slotfill-banner', {
```
moon0326, line 38: `window.wcCalypsoBridge.isEcommercePlan`  only detects ECOM plan. 

I think once this [PR](https://github.com/Automattic/wc-calypso-bridge/pull/926/files#diff-c9506099f3f037eaa3fc4b48305a4d641f4204266a5494640c4876e942ea711a) gets approved, we should add `window.wcCalypsoBridge.isFreeTrialPlan` and use that instead.

package.json
```diff
@@ -28,6 +28,10 @@
 	},
 	"dependencies": {
 		"@automattic/color-studio": "^2.5.0",
+		"@wordpress/components": "^19.5.0",
```
moon0326, line 31: These deps should be available already from [WooCommerceDependencyExtractionWebpackPlugin](https://github.com/Automattic/wc-calypso-bridge/blob/master/webpack.config.js#L11)

src/index.js
rjchow: Thanks! Good catch 😅

package.json
rjchow: I think we should still explicitly declare them as dependencies though? But yes we should set up the dep extraction properly so that it isn't re-included in the bundle (not sure if it is).

adrianduffell (approved):
Thanks @rjchow, this tested well and looks like it will be good blueprint to follow for the rest of the customizations.

package.json
adrianduffell: I think we'll need the dependencies declared or dev tooling, although admittedly things like Jest are not established right now.

xristos3490: > we should set up the dep extraction properly so that it isn't re-included in the bundle

That's a great idea, and it should also help with the JS linters!

> things like Jest are not established right now.

FYI, If we have a robust way of testing JS modules, we could definitely add it as part of the new GitHub Actions introduced in #926.

src/index.js
ilyasfoo: @rjchow The [PR](https://github.com/Automattic/wc-calypso-bridge/pull/926/files#diff-c9506099f3f037eaa3fc4b48305a4d641f4204266a5494640c4876e942ea711a) is now merged

chihsuan:
@rjchow I'm not sure if it's just me, but there is a blue underline on my atomic site (with Gutenberg disabled). I think we need to disable `text-decoration`. 🤔 

Screenshot 2023-02-22 at 15 38 53

Also, I see you changed `package-lock.json` but I don't see a package.json change. Is it expected?

rjchow:
> @rjchow I'm not sure if it's just me, but there is a blue underline on my atomic site (with Gutenberg disabled). I think we need to disable `text-decoration`. 🤔
> 
> Screenshot 2023-02-22 at 15 38 53
> 
> Also, I see you changed `package-lock.json` but I don't see a package.json change. Is it expected?

Fixed both issues! There was a package-lock.json change because a previous commit added some packages and I did a full clean of the package-lock.json after removing the packages 😅

chihsuan (approved):
Awesome! Tested well. 🚀

PR #930, opened 2023-01-30
Free trial payment restrictions

PanosSynetos:
Testing instructions

See pdqkbG-2oa-p2 for testing instructions

Additionally, make sure that when the `Set up payments` task is complete, you don't see any other payment gateway suggestions, other than

- PayPal
- Stripe
- WC Pay
- Bacs
- COD
- Cheque

wp-admin/admin.php?page=wc-admin&task=payments

image

-----
closes #924 

As described in

- peapX7-1rw-p2 ,
- peapX7-1xk-p2

in free trial, payments will be blocked for non-admin users

PR #936, opened 2023-02-07
Add free trial admin notice on WooCommerce -> Orders

### LINKED ISSUE #918: Free trial Orders screen admin notices
P2: pdibGW-1iQ-p2

## User Story

> As a free trial user, 
> I want to get a notice bar when viewing orders listing screen
> so that I'm aware of the limitations of free trial and able to proceed to purchase a plan when needed

## Acceptance Criteria

image

image

```
Feature: Admin notice on orders listing screen
Given the user is on a free trial plan
When the user views WooCommerce Orders screen
Then show admin notice with contents as above screenshot
```

```
Feature: Click on "Sign up for a plan" button
Given the user is on a free trial plan
When the user clicks on "Sign up for a plan"
Then redirect the user to plans page, ex: "https://wordpress.com/plans/sitenamehere"
```

moon0326:
Fixes #918 

This PR adds an admin notice on the Orders page when user is in free trial mode.

#### Test Instructions.

1. Checkout this branch
2. Change [this line](https://gist.github.com/moon0326/cac46c70a2cee81b61faef517fef7178#file-a-woocommerce-calypso-bridge-helper-php-L34) to false to force free trial mode
3. Navigate to `WooComemrce -> Orders`
4. You should see the notice.

Upon checking the latest figma oRIYBgxHeLDtlVrTjINlxf-fi-1090%3A41064, I noticed that the button copy changed to `Upgrade now`.


adrianduffell (approved):
This tested perfect! Thanks @moon0326 🚀

class-wc-calypso-bridge-frontend-free-trial.php
```diff
@@ -0,0 +1,61 @@
...
+
+		add_action('wp_enqueue_scripts', array($this, 'add_styles'));
+	}
+
+	/**
+	 * @return void
+	 */
+	public function add_plan_picker_banner()
+	{
+		$link = 'https://wordpress.com/';
```
PanosSynetos, line 47: The link should point to the correct plan URL - As discussed with @xristos3490 the most robust way is the following

```php
$site_slug = ( new \Automattic\Jetpack\Status() )->get_site_suffix();
$plan_url  = 'https://wordpress.com/plans/' . $site_slug;
```

PanosSynetos: I see that there is a function in this PR `get_action_url` that is already doing the same

PanosSynetos (changes requested):
Hey team - I added my comments.

I was wondering, instead of going full custom with the notice, couldn't we have used the native wc admin notices functionality?

I guess the button would be an issue; My question is honest, not judging the approach :)

class-wc-calypso-bridge-frontend-free-trial.php
moon0326: Thank you @PanosSynetos 👍 

I wasn't aware of `get_action_url` when I committed this change. I'll update it soon.

moon0326:
>  I was wondering, instead of going full custom with the notice, couldn't we have used the native wc admin notices functionality?

That's a good suggestion. I'll look into that!

> Oh and the notice's text in Figma, is different

Looks like it has been updated. I'll apply that.

moon0326:
@PanosSynetos  Updated the latest copy in https://github.com/Automattic/wc-calypso-bridge/pull/936/commits/7fe7e20656a1b778f081b09e376561da0b63fed5

PanosSynetos:
I'll check and re-review after the rebase. Planned for later in the day

PanosSynetos:
On my end, we're good, except some minor comments.

We're not using `<?=` anywhere else in calypso bridge, nor in wpcomsh. It would be better to keep it consistent.

 @xristos3490 I have a similar `admin_notice` on my #930 PR related to the free trial restrictions. https://github.com/Automattic/wc-calypso-bridge/pull/930/files#diff-f005d442494e72fae6c1d1df015bffb521313d257bca26b6df286bab08e5d880R329

Should we harmonize them? Should I follow the same styling as @moon0326 ?

xristos3490:
> I have a similar admin_notice on my https://github.com/Automattic/wc-calypso-bridge/pull/930 PR related to the free trial restrictions.

Design feedback would be great here. I actually don't see the notice in Settings in the Figma design. 

cc @FlorianBru Do we need the notice in the Settings > Payments section of wp-admin? If yes, should we follow the same design with a CTA on the right?

moon0326:
Thank you for the review 👍 
> We're not using `<?=` anywhere else in calypso bridge, nor in wpcomsh. It would be better to keep it consistent.

That's a good idea! Updated in https://github.com/Automattic/wc-calypso-bridge/pull/936/commits/82099c1ef205222eaabb874488615e7e4108b78e

includes/free-trial/class-wc-calypso-bridge-free-trial-orders-notice.php
```diff
@@ -0,0 +1,72 @@
...
+			return;
+		}
+
+		add_action('admin_notices', function() {
+			$screen = get_current_screen();
+			if ( 'edit-shop_order' === $screen->id ) {
+				?>
+				<div class="free-trial-orders-notice notice notice-info">
+                    <div>
+                        <h3><?php echo __('Start selling to everyone', 'wc-calypso-bridge')?></h3>
```
PanosSynetos, line 43: Thanks for fixing the shorthand echos. Could you please add `;` at the end of these commands?

PanosSynetos (changes requested):
Minor change suggested :) Apart from that, the code looks good - I cannot say about the notice's style and text. We need feedback from the design team

@FlorianBru is AFK today

FlorianBru:
Hey there 👋 

So when it comes to the order page the banner looks good. The only thing which is different from the design is the blue right border. In the design we still use the calypso pink. Was there a reason to switch to blue? @moon0326 

>I have a similar admin_notice on my https://github.com/Automattic/wc-calypso-bridge/pull/930 PR related to the free trial restrictions.
Design feedback would be great here. I actually don't see the notice in Settings in the Figma design.

I assume we are talking about this banner?
image

If that is true, I would suggest sticking with the banner Veronica designed for the payment section. Because if we would use the banner with the button here as well, the 'Finish setup' button would be less highlighted. @xristos3490

includes/free-trial/class-wc-calypso-bridge-free-trial-orders-notice.php
moon0326: 🤦 Sorry about that. Updated in https://github.com/Automattic/wc-calypso-bridge/pull/936/commits/6fdd9fa02a76064ec8a65eca76b659a44a1d6bbb

moon0326:
> The only thing which is different from the design is the blue right border. In the design we still use the calypso pink. Was there a reason to switch to blue? @moon0326

Thank you for catching it. That was my mistake 🤦 . I've updated it in [here](https://github.com/Automattic/wc-calypso-bridge/pull/936/commits/a9c5396570521bfba2274f37fbee6bd596f066c2) to use the current theme color.

includes/free-trial/class-wc-calypso-bridge-free-trial-orders-notice.php
```diff
@@ -0,0 +1,72 @@
+<?php
+
+/**
+ * Class WC_Calypso_Bridge_Free_Trial_Orders_Notice.
+ *
+ * @since   1.9.16
+ * @version 1.9.16
```
PanosSynetos, line 7: What about the versions? We're already in 2.0.3

PanosSynetos (changes requested):
PHP Side looks good, I cannot say anything about the designs if they are ok or not.

Just added a comment regarding versions - In general, has @Automattic/ghidorah updated the versions to 2.0.x for everything that has been released so far?

ilyasfoo:
Thanks for working on this, @moon0326! I didn't manage to test this PR, but I'm leaving a comment to suggest adding a changelog to `readme.txt`. Thanks!

includes/free-trial/class-wc-calypso-bridge-free-trial-orders-notice.php
moon0326: Updated in https://github.com/Automattic/wc-calypso-bridge/pull/936/commits/b3e521c1c70a417679a0493224f9c2cf703f3f92

Regarding version numbers, I think we should have a git hook to check class versions 🤔

moon0326: Hi @PanosSynetos ✋ 

Could you take a look at it one more time please? I've updated the version.

moon0326: Hi @PanosSynetos ✋ 

Could you please take a look at the change request? Versions have been updated.

PanosSynetos: Good :) Thanks and sorry I missed your ping yesterday

moon0326: Thank you 😄

PR #937, opened 2023-02-08
WooExpress: Hide the Launch task for sites on eCommerce trial

valterlorran:
The users on the eCommerce trial should not be allowed to launch the site. This PR hides the Launch task for users on the trial plan.

The back-end changes are being handled here: D100473-code

### Testing Instructions

* Sandbox the public-api.
* Sandbox the store
* Create a site with a business plan
* Use the WoA Developer Blog Manager to convert your site to a WoA dev site.
* Transfer your site to atomic.
* Navigate to `https://wordpress.com/hosting-config/{your-site}`, activate the hosting configurations, and create the ssh access.
* Use the credentials shown to access your site's VM.
* Disable the store in your sandbox so your site can be considered free.
* Navigate to the developer console
* Make a POST request to `WP.COM API v1.1 /sites/:siteSlug/ecommerce-trial/add/ecommerce-trial-bundle-monthly` to add the eCommerce Trial plan to your site.
* Add the changes in `includes/tasks/class-wc-calypso-task-launch-site.php` to your WoA VM -> `/wp-content/mu-plugins/wpcomsh/vendor/automattic/wc-calypso-bridge/includes/tasks/class-wc-calypso-task-launch-site.php` (I manually copied the changes and applied on the VM file)
* Navigate to: https://{YOUR_DOMAIN}/wp-admin/admin.php?page=wc-admin
* You should not see the Launch task anymore

Depends on https://github.com/Automattic/wc-calypso-bridge/pull/926
Related to https://github.com/Automattic/wp-calypso/issues/73015

### DIFF
--- a/includes/tasks/class-wc-calypso-task-launch-site.php
+++ b/includes/tasks/class-wc-calypso-task-launch-site.php
@@ -88,4 +88,15 @@ class LaunchSite extends Task {
 	public function is_complete() {
 		return 'launched' === get_option( 'launch-status' );
 	}
+
+	/**
+	 * Checks if the task is active. The task should not be active
+	 * if the site is on the eCommerce trial.
+	 * 
+	 * @return bool
+	 */
+	public function can_view() {
+		$ecommerce_trial_purchase = wp_list_filter( wpcom_get_site_purchases(), array( 'billing_product_slug' => 'wp-bundle-ecommerce-trial' ) );
+		return empty( $ecommerce_trial_purchase );
+	}
 }
--- a/includes/tasks/class-wc-calypso-task-launch-site.php
+++ b/includes/tasks/class-wc-calypso-task-launch-site.php
@@ -95,7 +95,6 @@ class LaunchSite extends Task {
 	 * @return bool
 	 */
 	public function can_view() {
-		$ecommerce_trial_purchase = wp_list_filter( wpcom_get_site_purchases(), array( 'billing_product_slug' => 'wp-bundle-ecommerce-trial' ) );
-		return empty( $ecommerce_trial_purchase );
+		return ! wc_calypso_bridge_is_ecommerce_trial_plan();
 	}
 }
--- a/includes/tasks/class-wc-calypso-task-launch-site.php
+++ b/includes/tasks/class-wc-calypso-task-launch-site.php
@@ -90,8 +90,7 @@ class LaunchSite extends Task {
 	}
 
 	/**
-	 * Checks if the task is active. The task should not be active
-	 * if the site is on the eCommerce trial.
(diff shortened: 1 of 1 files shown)


xristos3490:
#926 introduces a new global function named `wc_calypso_bridge_is_ecommerce_trial_plan()`. Can you block this until we merge the refactor (early next week) and use the same plan-detection functions across The Bridge?

@Automattic/ghidorah Also, what's your take on this? Are there any plans for introducing a dropdown at the bottom of this list with "hidden/disabled" setup tasks in the near future?

valterlorran:
@nelsonec87 I'm going to check it again.

@xristos3490 Yes, thanks for letting me know. We are also discussing whether we should hide the task or maybe add an upsell nudge on the Launch task page.

daledupreez:
My preference is to keep the task enabled, but do two things when we're on an eCommerce trial site:
1. Change the wording and CTA to indicate that you need to upgrade your store before you can launch, and
2. Change the CTA to direct the user to `https://wordpress.com/plans/:siteSlug` instead of the site settings.

daledupreez:
Also we should make sure we have exactly one team working on this specific change -- it may be that this is called out in the Figma already.

ilyasfoo:
> Also, what's your take on this? Are there any plans for introducing a dropdown at the bottom of this list with "hidden/disabled" setup tasks in the near future?

@xristos3490 Thanks for the ping! We should be good with this since the dropdown would be independently implemented

ilyasfoo:
> Also we should make sure we have exactly one team working on this specific change -- it may be that this is called out in the Figma already.

@daledupreez Currently, we're not handling this specific change since the origin of the task isn't in Core. We'd be happy if Serenity is helping out with it! 🙏

daledupreez:
I confirmed this with the designers earlier today: we want to ensure that users can finish the task list while they're still on the free trial, and that would not be possible if we have the Launch task visible. So we should go ahead with removing/disabling the task altogether.

xristos3490:
Hey folks! 

I've merged the plan detection (+refactor) work we released today into this branch! 

I was wondering about the most robust way of checking for the ecommerce free trial plan?

We currently use: https://github.com/Automattic/wc-calypso-bridge/blob/master/class-wc-calypso-bridge-dotcom-features.php#L124-L130 

Thoughts?

valterlorran:
@xristos3490 I used the `wc_calypso_bridge_is_ecommerce_trial_plan` function as you suggested. :)

xristos3490 (approved):
> I used the wc_calypso_bridge_is_ecommerce_trial_plan function as you suggested.

Thank you for keeping it consistent! 

Looks good!

PR #938, opened 2023-02-08
Add TypeScript support back

moon0326:
Typescript support was removed from [here](https://github.com/Automattic/wc-calypso-bridge/commit/47ec89a90cb3192a0a51fb0ad27e2bf7993a13c4)

This PR brings back TypeScript support.

### Testing Instructions.

- Run `npm start`
- Change an existing `.js` file to `.tsx`, for example `src/payments-welcome/index.tsx`
- See that it builds ok
- Make an intentional type error, for example `const PaymentsWelcomePage = () : number => (`
- See that a build error occurs.


adrianduffell (approved):
This appears to be working well! Thanks for re-instating it @moon0326 🚀 

I think we can merge it for now to help with our upcoming work.

SomewhereWarm team, I pinged you in the reviewers in case you had a strong reason to remove it?

xristos3490:
Thanks, @adrianduffell! I was under the impression that `@wordpress/scripts` supports typescript out of the box. Is there a specific reason why we are manually adding this babel-loader?

chihsuan:
> Thanks, @adrianduffell! I was under the impression that `@wordpress/scripts` supports typescript out of the box. Is there a specific reason why we are manually adding this babel-loader?

@xristos3490 👋  Just jumping in, I don't have the full context, but I guess it's because [TypesSript was added to  `@wordpress/scripts`](https://github.com/WordPress/gutenberg/pull/36260) after https://github.com/Automattic/wc-calypso-bridge/pull/699. I agree that we don't need to add `babel-loader` here. 😄 


@moon0326 @adrianduffell  I did a test locally. It works perfectly for me without the `babel-loader` and `babel.config.json`.

Here is a [link](https://github.com/WordPress/gutenberg/blob/b019306a4dd98908b2d1686f13ab7404fa7ddfd9/packages/scripts/config/webpack.config.js#L146) to the  `@wordpress/scripts` default webpack config.

adrianduffell:
Great finds, @xristos3490 @chihsuan !

Sounds like we'd be best using the built-in TS support in `@wordpress/scripts` and trimming things down a bit here.

If it works out well it seems we should bring this approach to woocommerce-admin too 🤔

moon0326:
Thank you everyone for looking into it!

I've removed TS setup and updated @wordpress/scripts to ^20.0.0

TS works well.

@adrianduffell @chihsuan @xristos3490 Could you give it another shot, please?

chihsuan (approved):
@moon0326 LGTM! Thanks for adding the TypeScript support back. ❤️ 

We might also need to update the `package-lock.json` using node 12.

xristos3490 (approved):
Works great! 💯 

> We might also need to update the package-lock.json using node 12.

Yup, we do use node 12 for now.

moon0326:
> @moon0326 LGTM! Thanks for adding the TypeScript support back. ❤️
> 
> We might also need to update the `package-lock.json` using node 12.

Thank you for catching it! Updated in https://github.com/Automattic/wc-calypso-bridge/pull/938/commits/80076b38b34d8e1c58e3b2e8d4092302a6c39dd3

PR #952, opened 2023-02-16
Fix conflict between woocommerce navigation and nav unification

ilyasfoo:
Fixes https://github.com/Automattic/wp-calypso/issues/65522

This PR adds a CSS that fixes empty column when WooCommerce navigation is enabled.

### Testing instructions

- In an ecom or business plan site, go to `wp-admin/options` and set `woocommerce_navigation_enabled` to `yes`
- Go to WooCommerce > Home
- Set browser size to at least 1024px
- Observe there's no empty column on the right hand side

image

### DIFF
--- a/assets/css/nav-unification.css
+++ b/assets/css/nav-unification.css
@@ -32,6 +32,10 @@
 		width: calc( 100% - 272px );
 	}
 
+	.jetpack-connected.is-nav-unification.has-woocommerce-navigation .woocommerce-layout__header {
+		width: 100%;
+	}
+
 	body.jetpack-connected.folded .woocommerce-layout__header {
 		width: calc( 100% - 36px );
 	}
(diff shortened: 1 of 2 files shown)


xristos3490 (approved):
Thanks for fixing this quickly! Looks good!

xristos3490:
Just one hint. This only affects business sites, not ecom. We include a hard-coded option filter to disable navigation feature when in ecom here: https://github.com/Automattic/wc-calypso-bridge/blob/master/includes/class-wc-calypso-bridge-woocommerce-admin-features.php#L83-L93

ilyasfoo:
>Just one hint. This only affects business sites, not ecom. We include a hard-coded option filter to disable navigation feature when in ecom here:

Ah yes, thanks for the note! I forgot about wc navigation being ecom-plan specific.

PR #959, opened 2023-02-16
Fatal error when using the bridge as a regular plugin

xristos3490:
### Specification

This PR aims to solve an issue with running The Bridge locally as a regular plugin. 

### Testing instructions

- Install The Bridge as a regular plugin by copying its contents into the`wp-content/plugins/wc-calypso-bridge` folder.
- Activate the plugin and notice a fatal error.
- Apply this fix.
- Ensure that the plugin can be successfully activated.


PanosSynetos (approved):
👌🚢

Issue #955, opened 2023-02-16
Free trial extension: Override products task completion logic


ilyasfoo:
Since WPCOM has a theme that adds products by default, we should override "Add products" task [completion logic](https://href.li/?https://github.com/woocommerce/woocommerce/blob/993f1f487995ad39dab81ffc72c83abcf7e1cfad/plugins/woocommerce/src/Admin/Features/OnboardingTasks/Tasks/Products.php#L164-L165) so that it wouldn't be unintentionally marked as completed.

@daledupreez has provided a sample (untested) code to filter out sample products pdtkmj-182-p2#comment-2053:

```php
$all_product_count = wp_count_posts( 'product' );
$headstart_product_ids = get_posts( array(
    'post_type' => 'product',
    'fields' => 'ids',
    'meta_key' => '_headstart_post',
) );
 
// Return number of products excluding the demo products
return $all_product_count - count( $headstart_product_ids );
```

I recommend this to be done as a follow-up to https://github.com/Automattic/wc-calypso-bridge/issues/945 since there would be some overlaps on how the task is going to be loaded by filter.

The new products are added for new sites with Tsubaki theme only, Dale suggested to setup an ecommerce site or WoA pdtkmj-182-p2#comment-2058 to help with testing. Ping him if you need help.

## Acceptance Criteria

1. When a user creates a site with Tsubaki theme, the "Add products" task should not be marked as completed

PR #963, opened 2023-02-17
Replace product task with custom completion logic

ilyasfoo:

Closes https://github.com/Automattic/wc-calypso-bridge/issues/955.

This PR adds a replacement product class that uses a custom completion logic. Applies to all plans.

However, this PR does not address if the user modifies the sample product, it'll simply not be counted as a published product.

1. Make sure there are no products in your site
2. Import products [CSV from headstart](https://gist.githubusercontent.com/ilyasfoo/1907a9d5e5b8a72fefe2af6e8d05ef03/raw/c3f3e5e2a52169480b9aeab6e1ec8e672fbb73db/headstart-product-export.csv) 
3. Go to WooCommerce > Home, observe that product task is not marked as completed
5. Add a new product and publish it
6. Go to WooCommerce > Home, observe that product task is marked as completed


includes/tasks/class-wc-calypso-task-headstart-products.php
```diff
@@ -0,0 +1,42 @@
...
+		return self::has_products();
+	}
+
+	/**
+	 * Check if the store has any published products,
+	 * but exclude sample products generated by Headstart themes.
+	 *
+	 * @return bool
+	 */
+	public static function has_products() {
```
chihsuan, line 27: @ilyasfoo I think we could just use one query to get the same outcome. What do you think?

```php
public static function has_products() {
	$product_query = array(
		'post_type' => 'product',
		'post_status' => 'publish',
		'fields' => 'ids',
		'meta_query' => array(
			array(
				'key' => '_headstart_post',
				'compare' => 'NOT EXISTS'
			)
		)
	);
	$products = get_posts($product_query);
	$product_count = count($products);

	return $product_count > 0;
}
```

chihsuan (approved):
Looks good and tested well. 💯

daledupreez (commented):
Without testing, this looks perfectly reasonable to me. Let me know if you'd like me to test as well, @ilyasfoo.

In terms of next steps, might it make sense to create a follow-up task to cleanly handle the case where users modify these products? I am happy to create the issue -- I am thinking that the implementation could also pull the `post_date_gmt` and `post_modified_gmt` fields in the query and remove any products with a post->modified gap larger than 60 seconds (or some similarly generous window).

But I strongly think that belongs in a follow-up PR so we can get this one shipped.

includes/tasks/class-wc-calypso-task-headstart-products.php
ilyasfoo: Actually, you have a point here. `get_posts` can be set to return only the post ids, which means the data return size would be negligible.

In contrast, your solution would only require a single round trip to db which is better.  I'm leaning towards making the change

ilyasfoo:
> Without testing, this looks perfectly reasonable to me. Let me know if you'd like me to test as well, @ilyasfoo.

@daledupreez Thanks, no need for another test 👍 

> In terms of next steps, might it make sense to create a follow-up task to cleanly handle the case where users modify these products? I am happy to create the issue

Yes that sounds good, thanks!

daledupreez:
I've created a follow-up issue in #973.

chihsuan (approved):
Thanks for making the change. LGTM! 💜

PR #974, opened 2023-02-21
Remove absolute path prefix from My Home and Customer menu URLs

daledupreez:

This PR modifies the _My Home_ and _Customers_ menu items to use URLs without a leading `/`, and to simply use `admin.php`. This should fix some situations where the _My Home_ menu item was incorrectly directing users to `/admin.php?page=wc-admin` instead of `/wp-admin/admin.php?page=wc-admin`.

Fixes Automattic/wp-calypso#73537

1. Add the updated code from this patch to a free eCommerce trial site where the filesystem is writable
   - Refer to these instructions for setting up such a site: peapX7-1D4-p2
   - For upload purposes, you'd need to update the modified file (`includes/class-wc-calypso-bridge-ecommerce-admin-menu.php`) in the following directory on the server: `~/htdocs/wp-content/mu-plugins/wpcomsh/vendor/automattic/wc-calypso-bridge/includes/`
3. With that in place, navigate to `/wp-admin/admin.php?page=wc-admin` on the site.
4. Use the left nav to navigate to _Extensions_ -> _Manage_
5. Mouse over the _My Home_ item in the left nav -- confirm that the URL has the full path to `/wp-admin/admin.php?page=wc-admin`

### Other information:

### FOR PR REVIEWER ONLY:

### DIFF
--- a/includes/class-wc-calypso-bridge-ecommerce-admin-menu.php
+++ b/includes/class-wc-calypso-bridge-ecommerce-admin-menu.php
@@ -111,7 +111,7 @@ class Ecommerce_Atomic_Admin_Menu extends \Automattic\Jetpack\Dashboard_Customiz
 	 * Adds My Home menu.
 	 */
 	public function add_my_home_menu() {
-		$this->update_menu( 'index.php', '/admin.php?page=wc-admin', __( 'My Home', 'jetpack' ), 'edit_posts', 'dashicons-admin-home' );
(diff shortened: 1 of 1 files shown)


xristos3490 (approved):
I couldn't replicate this one, @daledupreez! But it seems like a logical change. 

How Jetpack handles this `update_menu()` method turns out that the process eventually comes to the `add_submenu_page`, [which runs the `plugin_basename()`](https://github.com/WordPress/wordpress-develop/blob/6.1/src/wp-admin/includes/plugin.php#L1404) function on the URL, resulting in a string with trimmed slashes.

So, the result would be the same. You probably hit a race condition here, but we can approve this change overall!

Issue #972, opened 2023-02-20
Remove Avalara from Tax task


ilyasfoo:
Discussion: p1676641348447479-slack-C03Q87XT1QF

Avalara tax extension is not compatible with some bundled extensions and we should remove it from tax task until everything is fixed.

### Acceptance criteria

1. Replace the [tax fill](https://github.com/woocommerce/woocommerce/blob/trunk/plugins/woocommerce-admin/client/tasks/fills/tax/index.tsx) with another fill which automatically [navigates](https://github.com/woocommerce/woocommerce/blob/d26dd4c77208db31aa828a343777219a52a66f62/plugins/woocommerce-admin/client/tasks/fills/tax/index.tsx#L219-L221) to promoting WooCommerce Tax instead

![Image](https://user-images.githubusercontent.com/3747241/220124561-bcc31730-eeb2-485a-bae7-0263d1db19ff.png)

PR #975, opened 2023-02-21
Replace tax task to remove Avalara

ilyasfoo:

Closes https://github.com/Automattic/wc-calypso-bridge/issues/972. This PR:

1. Imports Tax task fill and replace the original fill with Avalara removed
2. Added shared variables and mixin stylesheets
3. Added some JS utils

**Test Avalara not suggested**
1. Make sure your store is located in Avalara Tax supported country, such as `US`
1. Go to WooCommerce > Home > Tax task
1. Observe that Avalara is not suggested

**Test WooCommerce Tax**
1. Click on "Yes please" and observe that you're redirected to Homescreen with Tax task completed
2. Go to WooCommerce > Settings > Tax
3. Observe that you have `Enable automated taxes` selected in the `Automated taxes` dropdown
1. Go to WooCommerce > Home > Tax task
2. Click on `No thanks, I'll set up manually` and observe that you're redirected to tax settings screen
1. Go to WooCommerce > Home > Tax task
2. Click on `I don't charge sales tax`
2. Observe that you're redirected to Homescreen with Tax task completed

**Test manual configuration**
1. Go to WooCommerce > Settings > General
5. Remove all address fields, set store country to "Malaysia - Johor" and save
6. Go to WooCommerce > Home > Tax task
7. Fill in all store address fields in `Set store location` and click on continue
8. Click on `Configure` and observe that you're redirected to tax settings
9. Go to WooCommerce > Settings > General
10. Observe that the store address values are saved from the values inserted in the previous form

#### WooCommerce Tax:
image

#### Manual configuration:
image


src/free-trial/tax/components/partner-card.scss
```diff
@@ -0,0 +1,50 @@
+@import '../../../stylesheets/variables';
```
ilyasfoo, line 1: I'd prefer to import this as a global import rather than each required file. I thought of using Webpack but wasn't sure how to do it proper. Any thoughts?

chihsuan: I 100% agree! I think we can set up a sass-loader `additionalData` like [this](https://github.com/woocommerce/woocommerce/blob/trunk/packages/js/internal-style-build/index.js#L48-L55).

src/index.js
```diff
@@ -33,16 +34,19 @@ registerPlugin( 'wc-calypso-bridge', {
...
+		'wc-admin-onboarding-task-payments',
+		'wc-admin-onboarding-task-tax',
+	];
 	addAction(
 		'plugins.pluginRegistered',
 		'wc-calypso-bridge',
 		function ( _settings, name ) {
-			if ( name === 'wc-admin-onboarding-task-payments' ) {
-				unregisterPlugin( 'wc-admin-onboarding-task-payments' );
+			if ( oldTaskNames.indexOf( name ) !== -1 ) {
```
chihsuan, line 48: I think we can use `includes` here, though they are very similar. 😄 

```suggestion
			if ( oldTaskNames.includes( name ) ) {
```

src/index.js
```diff
@@ -61,6 +65,23 @@ if ( !! window.wcCalypsoBridge.isEcommercePlanTrial ) {
 			</WooOnboardingTask>
 		),
 	} );
+
+	registerPlugin( 'wc-calypso-bridge-task-tax', {
+		// @ts-expect-error @types/wordpress__plugins need to be updated
```
chihsuan, line 70: We can remove this line since this file is not TS. 🙂

chihsuan (approved):
Only left a few comments. Looks good and tested well. 👍 

Pre-approving!

src/free-trial/tax/components/partner-card.scss
ilyasfoo: Thanks, @chihsuan! I was trying this out but it didn't turn out too great as I keep on importing new packages. I'm thinking to do this in a follow-up 🙏

chihsuan: Yeah, sounds great!

rjchow:
https://user-images.githubusercontent.com/27843274/220558679-9c95b2bb-59b8-4d02-b54f-45fdb8ddafdb.mov

Seeing a small visual artifact here but I'm not really sure what `variant=tertiary` should look like when `disabled=true` - maybe it should just not change?

In another PR in wc core earlier I noticed this oddity that any variant other than `primary` looks weird when isBusy so I just forced it to primary whenever `isBusy=true` , but I don't think that would look good here

ilyasfoo:
@rjchow Oof sorry I didn't see your comment before merging 🤦 

In my opinion, it looks just fine to have it grayed-out (I see it as disabled). Not sure if everyone would have the same opinion though. Still, I believe this would apply to all `wordpress/components` buttons?

chihsuan:
> In my opinion, it looks just fine to have it grayed-out (I see it as disabled). Not sure if everyone would have the same opinion though.

I think it's fine, too. But perhaps we can see if any feedback from @FlorianBru.

FlorianBru:
Just a few things comments based on the video shared by @rjchow 

- Why do the text links show a border during the hover event? Imo they should be text links rather than buttons.
- Do we need this button loading animation?
- Why are the text links greyed out?
- Is this layout based on an existing layout or did we create this from scratch? This also includes the illustration.

PR #939, opened 2023-02-08
Add task completion task for free trial

moon0326:
Fixes #928 

### Changes

1. Copied over [task completed component](https://github.com/woocommerce/woocommerce/blob/trunk/plugins/woocommerce-admin/client/two-column-tasks/completed-header.tsx) from the Core
2. Used task completion [slotfill](https://github.com/woocommerce/woocommerce/pull/36487) to add the modified component.

### Test Instructions

1. Checkout this branch
2. Change [this line](https://gist.github.com/moon0326/cac46c70a2cee81b61faef517fef7178#file-a-woocommerce-calypso-bridge-helper-php-L34) to false to force free trial mode
3. On your local WooCommerce env, add `taskList.isComplete = true` right after this [line](https://github.com/woocommerce/woocommerce/blob/trunk/plugins/woocommerce-admin/client/tasks/tasks.tsx#L129) to force tasklist completion. Make sure to build the assets to reflect the change.
4. Navigate to `WooComemrce -> Home`
5. You should see the updated task completion task for free trial.
6. Check the URL of `Sign up for a plan` and make sure it has your domain.
7. Test `How was your experience?` by clicking one of the emoticons.
8. Click `Share` and make sure it makes a request to save the selection.
9. Click the eclipse menu (triple dot) and click `Hide this`
10. Confirm the component is hidden now.
11. Delete `woocommerce_task_list_hidden_lists` option to bring it back.
12. Click `Show setup task list` and confirm the tasks are rendered again.
13. You can delete `woocommerce_task_list_keep_completed` to bring the completion component back. 

### Screenshots

Screen Shot 2023-02-08 at 2 44 12 PM

PR #978, opened 2023-02-22
Fix incorrect SVG size

moon0326:

After mering https://github.com/Automattic/wc-calypso-bridge/pull/939 PR, the bundle size increased to 8MB (!)

It looks like a tool I used to convert SVG to React component incorrectly added some bytes.

This PR fixes that.

Use the same test instructions from https://github.com/Automattic/wc-calypso-bridge/pull/939

### Other information:

### FOR PR REVIEWER ONLY:


rjchow (approved):
Can't really review this without having to complete all my tasks which will require setting up a new env afterwards, but the new SVG checks out so approving this since there are no other changes.

Issue #929, opened 2023-01-30
Free trial extension: Customize homescreen title and description


ilyasfoo:
Customized homescreen title and description that changes on certain steps completion

## User Story

> As a free trial user, 
> I want to get a customized homescreen title and description
> so that I'm motivated to complete tasks

## Acceptance Criteria

```
Feature: Custom copy according to how many tasks completed for free trial plan
Given the user is on a free trial plan
When the user views WooCommerce homescreen
Then show custom title and description according to how many tasks are completed
```

```
Feature: Custom copy according to how many tasks completed for a paid plan
Given the user has purchased a plan
When the user views WooCommerce homescreen
Then show custom title and description according to how many tasks are completed
```

#### The following are the free plan copies:

1. Up to 1 tasks completed:
    - Title: `Welcome to your free trial store`
    - Description: `You’re almost ready to start selling! Follow these steps to set up your test store. { n} out of 6 complete.`
1. 2 tasks completed:
    - Title: `Only a few more tasks to tick off!`
    - Description: `{ n} out of 6 complete.`
1. Up to 4 tasks completed:
    - Title: `Everything is looking great, keep it up!`
    - Description: `{n} out of 6 complete.`
1. 5 tasks completed:
    - Title: `Woo! We’ve made it to the last step! Great job`
    - Description: `{ n} out of 6 complete.`

#### The following are the purchased plan copies:

1. Up to all tasks completed:
    - Title: `Only a few more tasks to tick off!`
    - Description: `{ n} out of 8 complete.`

Figma: oRIYBgxHeLDtlVrTjINlxf-fi-646%3A29343

![Image](https://user-images.githubusercontent.com/3747241/214739803-f891f697-64f2-4dc1-b98b-8185e179f2c8.png)

Related extensibility PR: https://github.com/woocommerce/woocommerce/pull/36482

PR #987, opened 2023-02-23
Customize homescreen title and progress header

ilyasfoo:

Closes https://github.com/Automattic/wc-calypso-bridge/issues/929.

This PR adds homescreen title and header component fills.

image

#### Copy changes
1. Reset tasks so that none is completed
1. View homescreen
2. Observe the title is `Welcome to %s` where `%s` is your site title
3. Observe the copy below the title is `You’re almost ready to start selling! Follow these steps to set up your test store.
0 out of 6 complete.`
4. Complete 1 task
5. Observe the title is `Only a few more tasks to tick off!` and the copy below is `1 out of 6 complete.`
6. Complete 3 tasks, observe the title is `Everything is looking great, keep it up!`
6. Complete 5 tasks, observe the title is `Woo! We’ve made it to the last step! Great job`

#### Hide tasklist
1. Click on the burger menu on the right side of the title
2. Click on `Hide setup list`
3. Observe that the tasklist is now hidden


chihsuan:
Hi @ilyasfoo I noticed that the text overlaps the ellipsis icon in two columns layout on my test site.

Screenshot 2023-02-23 at 18 28 44

ilyasfoo:
Good catch, @chihsuan! Fixed in [ce8a354](https://github.com/Automattic/wc-calypso-bridge/pull/987/commits/ce8a3541c6fc506430b5db2e8d88fa5db6b08f05)

chihsuan (approved):
Nice work! @ilyasfoo  LGTM and tested well. 💯

src/homescreen-progress-header/progress-title.tsx
```diff
@@ -0,0 +1,87 @@
...
+		switch ( completedCount ) {
+			case 1:
+			case 2:
+				return __(
+					'Only a few more tasks to tick off!',
+					'wc-calypso-bridge'
+				);
+			case 3:
+			case 4:
+			default:
```
rjchow, line 67: Was surprised that this didn't take precedence over the bottom case - tested it to make sure it worked. Wonder if it would be better to move the default to the end as is the common practice?

rjchow (changes requested):
Small visual nit, but I think a bit of margin here at this breakpoint is necessary - I actually took care of this in the slot for the banner, on wc core side

image

src/homescreen-progress-header/progress-title.tsx
ilyasfoo: Yeah I tested this too (also asked ChatGPT if it was ok 😆 ). I'd agree if you could move it to end, I was torn between the ascending numbers and default positioning 😬

rjchow (approved):
fixt

PR #989, opened 2023-02-23
fix: enqueue wcpay customisation script on all admin pages

### LINKED ISSUE #981: WooCommerce payments customization not applied when navigating from home
### Description

When navigating to WCPay landing screen from homescreen, it's not applying the JS customizations done in https://github.com/Automattic/wc-calypso-bridge/pull/977

https://user-images.githubusercontent.com/3747241/220664736-c8b7ad1a-b815-49c8-aabe-da881e8c0d64.mp4

### To reproduce

1. In a free trial site, ensure WCPay is not set up yet
2. Go to WooCommerce > Home
3. Click on `Payments` from the navigation bar
4. Observe customizations are not applied
5. Refresh the page
6. Customizations are applied

rjchow:

The trial banner customisations script for the WCPay page was only being enqueued if the `$current_page['id']` is 'wc-payments'. 

This may seem correct but as the WC Payments page is loaded as a React router route, there is no page refresh when the page is navigated to from within the React application.

This PR enqueues the script on all wc-admin React pages by checking if `PageController::is_admin_or_embed_page()` is true. 

As the script does not do anything unless the URL param for `path` matches one of the specified payments page URLs, there should not be any noticeable overhead. Just a bit of extra JS to load.

Ideally this feature should be introduced by including a slotfill in either the WC Payments plugin, or a slotfill on the React router layout that passes in the page id - but that would require a release of wc-core to be usable.

Closes #981.

1. Set up a free trial instance with this plugin
2. Observe that the trial banner is displayed both when you navigate to the Payments page from the nav sidebar, and when you load the page directly.
3. Observe that no other page has these customisations applied.

### Other information:

### FOR PR REVIEWER ONLY:

### DIFF
--- a/assets/scripts/free-trial-wc-payments.js
+++ b/assets/scripts/free-trial-wc-payments.js
@@ -68,7 +68,8 @@
 			);
 
 			if ( stepItems.length === 3 ) {
-				const p = stepItems.querySelector( 'p' );
+				const { 2: setupCompleteStepElement } = stepItems;
+				const p = setupCompleteStepElement.querySelector( 'p' );
 				if ( p ) {
 					p.innerText = __(
 						'You’re ready to start testing the features and benefits of WooCommerce Payments',
--- a/includes/free-trial/class-wc-calypso-bridge-free-trial-wc-payments.php
+++ b/includes/free-trial/class-wc-calypso-bridge-free-trial-wc-payments.php
@@ -39,8 +39,7 @@ class WC_Calypso_Bridge_Free_Trial_WC_Payments  {
 		}
 
 		add_action('current_screen', function() {
-			$current_page = \Automattic\WooCommerce\Admin\PageController::get_instance()->get_current_page();
-			if ( isset( $current_page['id'] ) && strpos($current_page['id'], 'wc-payments') === 0 ) {
+			if ( \Automattic\WooCommerce\Admin\PageController::get_instance()->is_admin_or_embed_page() ) {
 				add_action('admin_enqueue_scripts', function() {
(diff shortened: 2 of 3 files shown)


ilyasfoo (changes requested):
Not sure if this is related to the PR, but I'm noticing this error in my console from this line https://github.com/Automattic/wc-calypso-bridge/blob/52e2c418f57846af844e71cdf343d2e5dbc07d4f/assets/scripts/free-trial-wc-payments.js#L71

image

rjchow:
> Not sure if this is related to the PR, but I'm noticing this error in my console from this line

Nice catch @ilyasfoo ! It was actually broken on trunk and I didn't notice that the copy modification wasn't taking place 😅

assets/scripts/free-trial-wc-payments.js
```diff
@@ -68,7 +68,8 @@
 			);
 
 			if ( stepItems.length === 3 ) {
-				const p = stepItems.querySelector( 'p' );
+				const { 2: setupCompleteStepElement } = stepItems;
+				const p = setupCompleteStepElement.querySelector( 'p' );
```
chihsuan, line 72: Just curious can we directly access the node element?

```suggestion
				const p = stepItems[ 2 ].querySelector( 'p' );
```

chihsuan (approved):
Tested well. Thanks for fixing the  "payment bugs". 👍

ilyasfoo (approved):
Works well, LGTM! 🚢 🚢

assets/scripts/free-trial-wc-payments.js
ilyasfoo: Could be possible with [nth-of-type](https://caniuse.com/?search=nth-of-type), but I'm not too picky at this moment

rjchow: that would work too! But I like to name things so that its clear what "element 2" is 😬

PR #991, opened 2023-02-24
Add blog name/site title as starter suggestion for Add a domain task

daledupreez:

This PR ensures that when we are on a site that has a primary site address that starts with `woo-` or `wooexpress-`, the "Add a domain" task will use the site title/blog name as the seed content for domain suggestions (assuming the title is not empty).

That is achieved as follows when we're building the task's target URL:
1. We get the `home_url()` for the current site, and checking if it starts with `woo-` or `wooexpress-` _and_ ends with `.wpcomstaging.com`. If it doesn't we return the existing URL without a suggestion.
2. We then get the `blogname` option - if it's empty, we return the existing URL.
3. We now know we have a title, and we add it as the `suggestion` URL parameter in the returned data.

Closes Automattic/wp-calypso#73393.

1. Start by setting up a free eCommerce trial site as a WoA dev site, as documented here: peapX7-1D4-p2
  - Note that the site's domain must start with `woo-` or `wooexpress-` for the new logic to work
3. Next, use SFTP to apply the changes from this patch on your dev site, where the `wc-calypso-bridge` code is located under `wp-content/mu-plugins/wpcomsh/vendor/automattic/wc-calypso-bridge/` - you can either copy up just this modified file, or you can rebuild the plugin locally and upload it as a whole. (If you do the latter, I'd recommend that you upload a zip file, `rm -r` the `wc-calypso-bridge` directory, and then `unzip` the zip file.)
4. Once the patch is applied, visit WooCommerce home for the site (`/wp-admin/admin.php?page=wc-admin`)
5. Click on the "Add a domain" task in the tasklist
6. Verify that you are taken to the domain suggestions page and the URL includes a `suggestion` URL parameter pulled from the site's blogname (which can be checked or modified via the _Site title_ setting under _Settings_ -> _General_ in the left nav.

### Other information:

### FOR PR REVIEWER ONLY:

### DIFF
--- a/includes/tasks/class-wc-calypso-task-add-domain.php
+++ b/includes/tasks/class-wc-calypso-task-add-domain.php
@@ -67,8 +67,19 @@ class AddDomain extends Task {
 	public function get_action_url() {
 		$status      = new \Automattic\Jetpack\Status();
 		$site_suffix = $status->get_site_suffix();
+		$domain_path = sprintf( "https://wordpress.com/domains/add/%s", $site_suffix );
+		$home_url    = \home_url( '', 'https' );
 
-		return sprintf( "https://wordpress.com/domains/add/%s", $site_suffix );
+		if ( ! \str_starts_with( $home_url, 'https://woo-' ) && ! \str_starts_with( $home_url, 'https://wooexpress-' ) ) {
+			return $domain_path;
+		}
+
+		$blog_name = \get_option( 'blogname' );
+		if ( empty( $blog_name ) ) {
+			return $domain_path;
+		}
+
+		return sprintf( '%s?suggestion=%s', $domain_path, rawurlencode( $blog_name ) );
 	}
 
 	/**
--- a/includes/tasks/class-wc-calypso-task-add-domain.php
+++ b/includes/tasks/class-wc-calypso-task-add-domain.php
@@ -70,6 +70,10 @@ class AddDomain extends Task {
 		$domain_path = sprintf( "https://wordpress.com/domains/add/%s", $site_suffix );
 		$home_url    = \home_url( '', 'https' );
 
+		if ( ! \str_ends_with( $home_url, '.wpcomstaging.com' ) ) {
+			return $domain_path;
+		}
+
 		if ( ! \str_starts_with( $home_url, 'https://woo-' ) && ! \str_starts_with( $home_url, 'https://wooexpress-' ) ) {
 			return $domain_path;
 		}


includes/tasks/class-wc-calypso-task-add-domain.php
```diff
@@ -67,8 +67,19 @@ public function get_time() {
 	public function get_action_url() {
 		$status      = new \Automattic\Jetpack\Status();
 		$site_suffix = $status->get_site_suffix();
+		$domain_path = sprintf( "https://wordpress.com/domains/add/%s", $site_suffix );
+		$home_url    = \home_url( '', 'https' );
 
-		return sprintf( "https://wordpress.com/domains/add/%s", $site_suffix );
+		if ( ! \str_starts_with( $home_url, 'https://woo-' ) && ! \str_starts_with( $home_url, 'https://wooexpress-' ) ) {
```
PanosSynetos, line 73: Hey @daledupreez `str_starts_with` is a PHP8 function - is there any chance that we end up with a PHP7.x installation?

I don't think it needs a `\` when calling it, as it's a built-in function that exists in the global namespace. 

Same for `home_url` and `get_option` - they do exist in the global namespace

daledupreez: > `str_starts_with` is a PHP8 function - is there any chance that we end up with a PHP7.x installation?

Good catch, @PanosSynetos! However, WordPress Core has had a polyfill available since 5.9.0 (at least per [this code](https://github.com/WordPress/wordpress-develop/blob/7c965e781f297e9f47f3abd3835c613eb41c0765/src/wp-includes/compat.php#L467-L489)). And we are running on at least WordPress 6.1 on Atomic servers.

> I don't think it needs a `\` when calling it, as it's a built-in function that exists in the global namespace.
> Same for `home_url` and `get_option` - they do exist in the global namespace

With the file namespaced, I wanted to be explicit about what I want to happen. So while these aren't _necessary_, it's very clear what they ought to be doing.

xristos3490 (approved):
LGTM! 

I have a sanity question @daledupreez. Is there a chance that we end up with a different domain other than the `*.wpcomstaging.com` in the free trial?

daledupreez:
> Is there a chance that we end up with a different domain other than the `*.wpcomstaging.com` in the free trial?

@xristos3490, if the user does nothing, then this is the only possible domain they have.

However, they can purchase a custom domain registration and use that instead, in which case they would be at `theircustomdomain.tld`, and not a subdomain of `wpcomstaging.com`.

If you're asking whether we might use a different domain other than `wpcomstaging.com` for new domains, then no, we won't.

I am not sure if that answers your question.

includes/tasks/class-wc-calypso-task-add-domain.php
```diff
@@ -67,8 +67,23 @@ public function get_time() {
 	public function get_action_url() {
 		$status      = new \Automattic\Jetpack\Status();
 		$site_suffix = $status->get_site_suffix();
+		$domain_path = sprintf( "https://wordpress.com/domains/add/%s", $site_suffix );
+		$home_url    = \home_url( '', 'https' );
 
-		return sprintf( "https://wordpress.com/domains/add/%s", $site_suffix );
+		if ( ! \str_ends_with( $home_url, '.wpcomstaging.com' ) ) {
```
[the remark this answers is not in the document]
xristos3490: Looks defensive, but it certainly doesn't hurt. 

So domain suggestions will only happen when:
- The site doesn't have a domain already, **and**;
- The site **is** originated from the WCCOM NUX

Correct? If so, looks good!

includes/tasks/class-wc-calypso-task-add-domain.php
```diff
@@ -67,8 +67,23 @@ public function get_time() {
...
+		if ( ! \str_starts_with( $home_url, 'https://woo-' ) && ! \str_starts_with( $home_url, 'https://wooexpress-' ) ) {
+			return $domain_path;
+		}
+
+		$blog_name = \get_option( 'blogname' );
+		if ( empty( $blog_name ) ) {
+			return $domain_path;
+		}
+
+		return sprintf( '%s?suggestion=%s', $domain_path, rawurlencode( $blog_name ) );
```
chihsuan, line 86: Hi @daledupreez  Since we don't limit user input format in NUX, `$blog_name` might contain non-alphanumeric, upper characters, or would be too long. Do you think we need to normalize `rawurlencode( $blog_name )` further? For example, convert it to lowercase, and replace blank spaces with `-`.

daledupreez: Good thought, @chihsuan!

The code responsible for performing domain searches should be able to handle whatever we throw at it, as users can also enter whatever they like in the input field. All we are doing with this PR is supplying specific content for the input field in the UI, so I don't think we need to be more defensive than this.

includes/tasks/class-wc-calypso-task-add-domain.php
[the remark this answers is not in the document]
daledupreez: Yes, that's pretty much the change, so we really do limit the cases where we use the site title.

includes/tasks/class-wc-calypso-task-add-domain.php
chihsuan: > The code responsible for performing domain searches should be able to handle whatever we throw at it

Nice, that’s good to hear! :slightly_smiling_face:

PR #1025, opened 2023-03-03
Add defensive WC_Calypso_Bridge_Instance->get_site_slug() helper function

daledupreez:

This PR adds a defensive `get_site_slug()` helper function to `WC_Calypso_Bridge_Instance`. We've seen fatals in production due to the `Automattic\Jetpack\Status` class not existing (though it _should_ be autoloaded by Jetpack), so this PR adds a wrapper function that should still work in cases where the Jetpack class isn't present. When that class doesn't exist, we fall back on using `home_url()` to identify the primary location of the site. While this isn't foolproof, it shouldn't break as badly.

I am honestly not sure how best to test the failing case for this effectively, as we haven't triggered the fatals on dev sites.

However, we should do the following happy path testing to verify that the changes are working when the Jetpack code is loaded:
* Apply these changes to a development site -- see peapX7-1D4-p2 for instructions on setting one up
* Verify that the following links are correct:
  - The _Add a domain_ task in the onboarding checklist
  - The site launch task (which is only visible for upgraded sites)
  - _Extensions_ -> _Discover_ - the "Upgrade now" CTA should point at `/plans/:siteSlug`
  - The Earn and Marketing Tools submenu items are _not_ visible in the Tools menu
  - _Settings_ -> _WooCommerce_ -> "Payments" tab -- verify the "upgrade to a paid plan" links to `/plans/:siteSlug` in the "Only Administrators and Store Managers can place orders..." notice
  - _Orders_ -> _Orders_ -- verify the "Upgrade now" CTA points at `/plans/:siteSlug`
  - 

### Other information:

### FOR PR REVIEWER ONLY:

### DIFF
--- a/class-wc-calypso-bridge-shared.php
+++ b/class-wc-calypso-bridge-shared.php
@@ -106,7 +106,7 @@ class WC_Calypso_Bridge_Shared {
 			filemtime( WC_CALYPSO_BRIDGE_PLUGIN_PATH . '/build/style-index.css' )
 		);
 
-		$site_suffix = wc_calypso_bridge_get_site_slug();
+		$site_suffix = WC_Calypso_Bridge_Instance()->get_site_slug();
 
 		$params      = array(
 			'isEcommercePlan'              => (bool) wc_calypso_bridge_has_ecommerce_features(),
--- a/class-wc-calypso-bridge-shared.php
+++ b/class-wc-calypso-bridge-shared.php
@@ -4,7 +4,7 @@
  *
  * @package WC_Calypso_Bridge/Classes
  * @since   1.0.0
- * @version 2.0.
+ * @version 2.0.8
  */
 
 defined( 'ABSPATH' ) || exit;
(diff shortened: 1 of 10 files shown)


xristos3490:
I've added a commit to ensure that all occurrences are replaced with this helper just to keep them consistent! 

In terms of validation in this one, I think there isn't much we can do since we cannot replicate the state where the Jetpack classes are missing. I would opt to move forward with this without specific testing steps. 

@daledupreez Thoughts?

daledupreez:
@xristos3490, thanks for cleaning up the other usages! I've just tested all the items except for the site launch option, but I think that should be OK. I vote we move forward as-is, especially as I believe you already checked the manual site slug construction.

If you agree, could you approve the PR? (As the author, I am not allowed to do that!)

PanosSynetos (approved):
👌

xristos3490:
> I believe you already checked the manual site slug construction.

Yup. It worked great! 

I was looking around and saw this case here, which does a similar string construction: https://github.com/Automattic/wpcomsh/blob/trunk/wpcom-migration-helpers/site-migration-helpers.php#L27-L38

Should we perhaps go with the `site_url()`? It seems less risky since a home URL might differ from the shop one? 🤔 

It also uses a check for `http` for whatever reason.

daledupreez:
@xristos3490, I've updated the implementation to use `site_url()` (and `untrailingslashit()`) to mimic the code you referenced.

I think this is likely the right approach for the site slug, and my testing looks OK. Would you mind checking my code as a final sanity check before I merge?

xristos3490 (approved):
Looks great! 👌

PR #1066, opened 2023-04-05
Make the free trial banner responsive

### LINKED ISSUE #1065: Free trial notice banner is not responsive
The free trial notice banner doesn't fit the screen correctly when the browser width is less than 782px.

Originally reported in p1680619720122459-slack-CNA89KY6M

### Example: 
Screen Shot 2023-04-05 at 9 12 09 AM

### Cause

On a WPCOM site, we have `admin-bar` mu-plugin that contains https://s0.wp.com/wp-content/mu-plugins/admin-bar/wpcom-admin-bar.css?ver=12.0-a.7 CSS.

The CSS file defines `--wp-admin--admin-bar--height:32px` variable, which overrides the Core's `--wp-admin--admin-bar--height` definition, ignoring the media query from the Core.

Screen Shot 2023-04-05 at 9 03 31 AM

This is how it looks without the override:

It looks better but still has problems.

Screen Shot 2023-04-05 at 9 18 34 AM

moon0326:

Closes #1065

This PR makes the free trial notice banner responsive and fixes a few minor glitches. 

1. Render the banner right after the `body` tag so that we don't have to calculate body tag `margin-top` manually.
2. Add media query and set the `--wp-admin--admin-bar--height` manually. As described in #1065, `admin-bar` mu-plugin manually sets `-wp-admin--admin-bar--height`, which makes the browser ignore the media query from the core. Setting `-wp-admin--admin-bar--height` with `!important` is a temporary solution. I'll submit an issue to `admin-bar` mu-plugin to fix the root cause.
3. Hide the free trial banner on a preview screen (screenshot below)

Since one of the problems came from `admin-bar` mu-plugin, it is better to test this PR on a real trial site with `admin-bar` mu-plugin. Please set up a free trial site, convert it to a dev blog, and copy changes in this PR before testing.

1. Go to /view/:your-site-slug
2. Make sure the banner is not rendered (screenshot 1)
3. Go to the frontend
4. Make sure the banner is rendered right after the admin bar
5. Change your browser width to various sizes and make sure the free trial banner works without breaking words.
6. Make sure the free trial banner works with a few different themes.

### Screenshots
Screen Shot 2023-04-05 at 10 37 41 AM

### FOR PR REVIEWER ONLY:

### DIFF
--- a/assets/css/free-trial-plan-picker-banner.css
+++ b/assets/css/free-trial-plan-picker-banner.css
@@ -28,6 +28,12 @@ body.has-free-trial-plan-picker .woocommerce-store-notice {
 	}
 }
 
+@media screen and (max-width: 480px) {
+	#free-trial-plan-picker-banner {
+		position: inherit;
+	}
+}
+
 body.hide-free-trial-plan-picker #free-trial-plan-picker-banner {
 	display: none;
 }
--- a/assets/css/free-trial-plan-picker-banner.css
+++ b/assets/css/free-trial-plan-picker-banner.css
@@ -4,7 +4,6 @@
 	font-size: 0.8125em;
 	position: sticky;
 	z-index: 99999;
-	width: 100%;
 	padding: 5px;
 	text-align: center;
 	top: calc( var( --wp-admin--admin-bar--height, 0px ) ) !important;
--- a/assets/css/free-trial-plan-picker-banner.css
+++ b/assets/css/free-trial-plan-picker-banner.css
@@ -3,7 +3,7 @@
 	color: #ffffff;
 	font-size: 0.8125em;
 	position: sticky;
-	z-index: 9999;
+	z-index: 99999;
 	width: 100%;
 	padding: 5px;
 	text-align: center;
--- a/assets/css/free-trial-plan-picker-banner.css
+++ b/assets/css/free-trial-plan-picker-banner.css
@@ -2,13 +2,11 @@
 	background-color: #c9356e;
 	color: #ffffff;
 	font-size: 0.8125em;
-	position: fixed;
+	position: sticky;
 	z-index: 9999;
 	width: 100%;
-	height: 40px;
-	display: flex;
-	justify-content: center;
-	align-items: center;
+	padding: 5px;
+	text-align: center;
 	top: calc( var( --wp-admin--admin-bar--height, 0px ) ) !important;
 }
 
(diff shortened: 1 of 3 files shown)


includes/free-trial/class-wc-calypso-bridge-free-trial-plan-picker-banner.php
```diff
@@ -36,9 +36,11 @@ public function __construct() {
 
 		add_action('init', function() {
 			if ( current_user_can( 'manage_woocommerce' ) ) {
-				add_action( 'wp_footer', array( $this, 'add_plan_picker_banner' ), 10 );
+				add_action( 'wp_head', array( $this, 'add_plan_picker_banner' ), -2000 );
```
moon0326, line 39: In a normal situation, `wp_body_open` should be used instead. However, `wpcomsh` uses [wp_head](https://github.com/Automattic/wpcomsh/blob/trunk/private-site/private-site.php#L102) hook to render the launch banner (screenshot). In order to render the free trial banner before the launch bar, we also have to use `wp_head` with higher priority.

FYI, when a browser detects an HTML element in the head tag, it actually renders the element in the body correctly.

Screen Shot 2023-04-05 at 11 10 39 AM

assets/css/free-trial-plan-picker-banner.css
```diff
@@ -2,13 +2,11 @@
 	background-color: #c9356e;
 	color: #ffffff;
 	font-size: 0.8125em;
-	position: fixed;
-	z-index: 9999;
+	position: sticky;
+	z-index: 99999;
 	width: 100%;
```
ilyasfoo, line 7:
```suggestion
```

On smaller screens (tested on iPhone XR) it seems there's a white spacing on the right caused by the banner:

image

I find that removing width actually fixes it without any side effects. Let me know if this works!

image

ilyasfoo (changes requested):
Thanks for the quick fix on this, @moon0326!

I spotted a few glitches:

This may not be from this PR, but the banner floats with a top offset when scrolling down. If you confirm this isn't related, I could create a follow-up issue to address it

image

moon0326:
> This may not be from this PR, but the banner floats with a top offset when scrolling down. If you confirm this isn't related, I could create a follow-up issue to address it

Thank you!

That's actually a glitch. A fix is on the way 🚗

moon0326:
@ilyasfoo Updated in https://github.com/Automattic/wc-calypso-bridge/pull/1066/commits/5606d638794b6f30320104e9a0e9acc7d171ad8f

Thank you!

ilyasfoo (approved):
Awesome, thanks for the fix! Tested well, LGTM!

chihsuan (approved):
Looks perfect and tested well! 🚢

PR #1083, opened 2023-04-19
Add checks for domain purchases to domain task

daledupreez:

This PR explores a more robust mechanism for checking if the site has added a custom domain, as the current logic only checks for the site's primary address being different. In most cases, that's sufficient, but in the case where it takes a while for a domain to be registered and for the DNS details to be propagated, there can be a lengthy delay of 10-30 minutes before the site gets switched across to the new domain. (For some TLDs, that delay can be even longer.)

The implementation logic has now been updated to perform three levels of checks:
1. If the site includes `.wpcomstaging.com` in the URL, we mark the task as complete
   - It's worth noting that this is a minor refinement of the current logic, which only checks for `wpcomstaging`
2. We check if the `wpcom_get_site_purchases()` function exists. If it doesn't we mark the task as incomplete, as we can't perform the next step.
3. We call `wpcom_get_site_purchases()`, and filter the results so we're only looking at purchases where `product_type` is either `domain_map` or `domain_reg` for domain mappings (AKA connected domains) and domain registrations. If any such purchases exist, we can assume the customer has set up a domain, and we can mark the task as complete. 

* Set up an eCommerce site as a development site
* Connect a domain to the site, but ensure that this domain is not the primary domain, i.e. we keep the `wpcomstaging.com` URL
* Visit WooCommerce Home
* Verify the task is incomplete
* Apply this patch
* Reload WooCommerce Home
* Verify the task is marked as complete

### Other information:

### FOR PR REVIEWER ONLY:

### DIFF
--- a/includes/tasks/class-wc-calypso-task-add-domain.php
+++ b/includes/tasks/class-wc-calypso-task-add-domain.php
@@ -100,7 +100,25 @@ class AddDomain extends Task {
 	 * @return bool
 	 */
 	public function is_complete() {
-		// Determine if a custom domain is used by ensuring that the default atomic url wpcomstating is not part of the `siteurl` option.
-		return false === strpos( get_option( 'siteurl' ), 'wpcomstaging' );
+		// Determine if a custom domain is used by ensuring that the default atomic url .wpcomstaging.com is not part of the `siteurl` option.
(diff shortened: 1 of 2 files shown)


chihsuan (approved):
Looks good! Confirmed the task was marked as completed immediately with this patch. 👍

PR #1088, opened 2023-04-21
Update node to 16 and various dependencies to current versions

daledupreez:

This PR tries to update various dependencies used to build the client-side code for `wc-calypso-bridge`, mainly so we can be somewhat more current, possibly as a starting point to more automation in the build process.

This PR includes the following major changes:
 * Bump `node` to `16` by modifying `.nvmrc`
 * Bump `@woocommerce/dependency-extraction-webpack-plugin` from `^1.7.0` to `^2.2.0`
    - It looks like our webpack config wasn't subject to the change that triggered the version 2 bump: https://github.com/woocommerce/woocommerce/blob/e88152ff3c1e984dc6524c7f6d6097ff85821c19/packages/js/dependency-extraction-webpack-plugin/PREVIOUS_CHANGELOG.md?plain=1#L8
 * Bump `@woocommerce/eslint-plugin` from `1.1.0` to `^2.2.0`
 * Bump `@wordpress/prettier-config` from `^1.1.1` to `^2.14.0`
 * Bump `@wordpress/scripts` from `^20.0.2` to `^24.6.0`
   - The current version is at 26.x, but it looks like 25.0.0 switched to React 18, and I am not sure if we can do that yet: https://github.com/WordPress/gutenberg/blob/a47a09ee5fa6f19f45ab70b20f9ebe0db3fda6ed/packages/scripts/CHANGELOG.md?plain=1#L44-L48
 * Bump `wp-prettier` from `^2.2.1-beta-1` to `^2.8.5` (though we use it as `prettier`)
 * Completely removed `package-lock.json` and regenerated the file from scratch -- it was also using an older lockfile definition, which slowed down installs and definitely caused me additional warnings

The PR also includes the following minor changes:
 * Shift `@automattic/color-studio` to allow newer versions using `^2.5.0`
 * Bump `@woocommerce/onboarding` to `^3.3.0`
 * Bump `dompurify` to `^2.4.5`
 * Bump `gridicons` to `^3.4.2`

I also decided not to update TypeScript from `^4.9.5` to `5.x`, as I am not familiar enough with TypeScript to know if we have any likely issues. This change should also be a relatively simple update in a follow-up  if we want to tackle that.

1. Checkout this branch locally - `git checkout try/update-node-and-dependencies`
2. Run `nvm use` - if you don't have version 16 of node installed, you may need to run `nvm install 16` 
3. Run `npm install`
4. Verify that the install completes successfully -- there will still be some warnings, especially for resolving the version of `react`
5. Run `npm run build`
6. Verify that the build completes successfully
7. _I am not sure how we generally test the behaviour of the generated Javascript package. I need someone to help me with that._

### Other information:

### FOR PR REVIEWER ONLY:


chihsuan (approved):
Thanks @daledupreez, LGTM, and tested well on my atomic dev site! 🚢 

I noticed that the JS lint rules and unit test behaviors seem to have changed, but since there are already some errors in the master branch, I think we can fix it in a follow-up PR or when we touch those files again.

> The current version is at 26.x, but it looks like 25.0.0 switched to React 18, and I am not sure if we can do that yet:

I've also tested with the latest wp.scripts, and it works!

> I also decided not to update TypeScript from ^4.9.5 to 5.x, as I am not familiar enough with TypeScript to know if we have any likely issues. This change should also be a relatively simple update in a follow-up if we want to tackle that.

We're still using TS 4 in WC Core. Perhaps we can update it to 5 when we update TS in Core. 🙌

daledupreez:
The unit tests look completely broken, so I think we should ensure they are actually runnable.

chihsuan:
Hey @daledupreez I've fixed unit tests in https://github.com/Automattic/wc-calypso-bridge/pull/1094. I'd appreciate it if you could take a look when you get a chance. Thank you! 🙂

PR #1127, opened 2023-05-08
Use the correct text domain for i18n functions

d-alleyne:

Relates to https://github.com/Automattic/i18n-issues/issues/594

Some strings in the WordPress Admin appear untranslated (i.e. in English) when setting up an eCommerce site for the first time. On examining the codebase, many i18n function calls use text domains different from `wc-calypso-bridge`. This commit ensures:
- all strings from this repository use the same text domain
- the linter (phpcs) enforces the use of the `wc-calypso-bridge` domain

#### How to test the changes in this PR
1.
2.
3.

### Other information:

### FOR PR REVIEWER ONLY:

-   [x] I have reviewed that everything is sanitized/escaped appropriately for any SQL or XSS injection possibilities. I made sure Linting is not ignored or disabled.


daledupreez (commented):
I have some comments up from inspection, but I think we also need input from @Automattic/somewherewarm and @Automattic/ghidorah, as they worked on a number of the features where we're trying to use strings from other domains/projects.

PanosSynetos (changes requested):
I agree with Dale that NOT all the text domains need to change to wc-calypso-bridge - Those that come directly from WooCommerce or Jetpack and are not altered, should stay as is.

There are a lot of changes to go through and rollback, I'll try and wrap it before my AFK; given that I wrap up all my other commitments.

includes/class-wc-calypso-bridge-ecommerce-admin-menu.php
```diff
@@ -176,7 +176,7 @@ public static function menu_order( $menu_order ) {
 	 * Adds My Home menu.
 	 */
 	public function add_my_home_menu() {
-		$this->update_menu( 'index.php', 'admin.php?page=wc-admin', __( 'My Home', 'jetpack' ), 'edit_posts', 'dashicons-admin-home' );
+		$this->update_menu( 'index.php', 'admin.php?page=wc-admin', __( 'My Home', 'wc-calypso-bridge' ), 'edit_posts', 'dashicons-admin-home' );
```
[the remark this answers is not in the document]
d-alleyne: @daledupreez Thank you for taking the time to provide feedback. Instead of immediately committing these changes, I'll wait for input from other teams. 

I suspect that I'll need to add `phpcs:ignore WordPress.WP.I18n` comments based on their feedback

daledupreez (commented):
The updated changes look much better to me, though I don't know how to effectively test the PHPCS workflow.

Unfortunately, I can't test this locally as I have PHP 8.2 installed, and the i18n sniff triggers errors.

daledupreez (commented):
OK, I managed to work around the issue, but I do get a lot of mismatched text domains returned. It's not clear if that's expected or not.

d-alleyne:
@daledupreez 
> It's not clear if that's expected or not.

The `Get changed files` section of Github workflow does the equivalent locally:
```bash
git diff --name-only origin/master fix/use-correct-textdomain | grep '\.php$' | xargs
```
The resulting string will restrict PHPCS to only run on 3 PHP files that were changed in this PR. Observe the new stages in the most recent [workflow run](https://github.com/Automattic/wc-calypso-bridge/actions/runs/5802221431/job/15728155622?pr=1127#step:10:2)

For this commit, you can run PHPCS locally like:
`vendor/bin/phpcs --warning-severity=0 --sniffs=WordPress.WP.I18n includes/class-wc-calypso-bridge-free-trial-payment-restrictions.php includes/class-wc-calypso-bridge-woocommerce-admin-features.php includes/tasks/class-wc-calypso-task-add-domain.php`

In future commits, when other files are changed, the developer can decide whether to use `phpcs:ignore` (for example, when the same string is present in the source project) or whether the `wc-calypso-bridge` text domain is needed.

bin/lint-php-i18n.sh
```diff
@@ -0,0 +1,4 @@
+#!/usr/bin/env bash
+DEFAULT_BRANCH=$(git symbolic-ref refs/remotes/origin/HEAD | sed 's@^refs/remotes/origin/@@')
+CHANGED_PHP_FILES=$(git diff --name-only origin/$DEFAULT_BRANCH | grep '\.php$' | xargs)
```
daledupreez, line 3: I am seeing some weirdness in my testing as a result of some super-subtle, implicit assumptions in this code:
1. The changes here may include deleted files, which can't be found by `phpcs`. We can work around that by adding `--diff-filter=d` to exclude deleted files.
```suggestion
CHANGED_PHP_FILES=$(git diff --name-only --diff-filter=d origin/$DEFAULT_BRANCH | grep '\.php$' | xargs)
```
2. This code will only work when the current branch has been rebased off of `origin/master`. I am hitting some unexpected files in the command because the current code was branched off from master at some point in the past, but our diff is against the _current_ state of `origin/master`. I think we need to use an interim step to find the right starting point for our diff. I think something like the following should work (based on a quick inspection of https://github.com/sirbrillig/phpcs-changed/blob/trunk/PhpcsChanged/UnixShell.php, which implements a Git-aware `phpcs-changed` command):
```suggestion
INITIAL_COMMIT_FOR_BRANCH=$(git merge-base origin/$DEFAULT_BRANCH HEAD)
CHANGED_PHP_FILES=$(git diff --name-only --diff-filter=d $INITIAL_COMMIT_FOR_BRANCH | grep '\.php$' | xargs)
```
We could also just add `@sirbrillig/phpcs-changed` as a dev dependency to avoid reinventing this wheel!

daledupreez (changes requested):
I tripped over something subtle with the Git tracking that I think we need to fix.

bin/lint-php-i18n.sh
d-alleyne: Thanks for the insight! I was unaware of all of these bugs. Will read up on `@sirbrillig/phpcs-changed`

d-alleyne: Good idea.

daledupreez (approved):
These simpler changes look good to me!

Issue #1096, opened 2023-04-25
Deploy Automation: Automate the deploy process


markbiek:
As we've had several imcomplete/broken releases recently, we want to start trying to automate the release process.

The plan is to take a similar approach as we did on the [premium-theme-tools](https://github.com/Automattic/premium-theme-tools) repo. We'll have individual functions for each part of the process and then a command-runner script that does the actual work

This is a high-level issue to track the automation status. We'll be splitting the work into individual issues which will be linked here.

_**When making updates to the automation scripts, make sure the changelog is updated to reflect those changes!**_

**Prioritized Tasks**

- [x] #1100 
- [x] #1101 
- [x] #1097 
- [x] #1098
- [x] https://github.com/Automattic/wc-calypso-bridge/issues/1147 
- [x] #1099 
- [ ] https://github.com/Automattic/i18n-issues/issues/594
- [x] https://github.com/Automattic/wc-calypso-bridge/issues/1134

markbiek:
@valterlorran found this nice package for doing some nice cli interfaces. https://www.npmjs.com/package/inquirer

markbiek:
Quick status update: At this point, the entire release prep and release creation process is automated! There is still some lingering work related to translation files but the release process can be run as-is without it.

markbiek:
I think we can close this. The main automation is complete and usable.

PR #1130, opened 2023-05-11
i18n: Add a mechanism to update the files in the languages/ folder

d-alleyne:

Related to https://github.com/Automattic/i18n-issues/issues/594,  https://github.com/Automattic/wc-calypso-bridge/issues/1096.

This library [loads `.mo` files in the `languages` directory](https://github.com/Automattic/wc-calypso-bridge/blob/01dfb42a44d4c52130c19eb5d1aece4cdf6bdf51/class-wc-calypso-bridge.php#L203) to show translated strings to the user. However, this folder has not been updated in many years. This leads to instances where new features are untranslated to non-English users. 

This PR adds a `makepot` entry into `package.json` which generates a `.pot` file that contains the latest strings, as well as downloading the latest translations from GlotPress. The functionality is similar the system found in [wpcomsh](https://github.com/Automattic/wpcomsh/tree/trunk#how-the-translation-system-works), and we're looking to replicate that process here.

1. Checkout this commit
2. Run `npm run makepot`
3. Verify that the `languages/` folder has `wc-calypso-bridge.pot` as well as updated `.po` and `.mo` files

### Other information:

### FOR PR REVIEWER ONLY:


package.json
```diff
@@ -20,7 +20,8 @@
 		"test:e2e": "wp-scripts test-e2e",
 		"test:unit": "wp-scripts test-unit-js",
 		"create-release": "node ./scripts/release.js",
-		"bump-version": "node ./scripts/commands/bump-version.js"
+		"bump-version": "node ./scripts/commands/bump-version.js",
+		"makepot": "wp i18n make-pot . languages/wc-calypso-bridge.pot && bash ./bin/update-translations.sh"
```
markbiek, line 24: I think this should follow the other build/release commands and automatically do a `git commit` if `update-translations.sh` runs successfully.

Eventually, we're going to want users to be able to run a single command which bumps the version, updates the changelog, updates the pot files, and then creates a PR for all those changes.

markbiek: Also, will `wp i18n make-pot` output a useful error message if it fails for some reason? I'd like to make sure the user gets some feedback if something goes wrong and the `update-translations.sh` step doesn't run.

bin/update-translations.sh
```diff
@@ -0,0 +1,25 @@
...
+	if [[ "$LANG_WP_LOCALE" == "de_DE" ]] && [[ "$LANG_SLUG" == "formal" ]]; then
+		LANG_WP_LOCALE="de_DE_formal"
+	fi
+
+	LANG_FILENAME=languages/wc-calypso-bridge-$LANG_WP_LOCALE
+
+	echo "Downloading $LANG_NAME ($LANG_LOCALE/$LANG_SLUG → $LANG_WP_LOCALE)"
+	curl -so $LANG_FILENAME.po https://translate.wordpress.com/projects/woocommerce/wc-calypso-bridge/$LANG_LOCALE/$LANG_SLUG/export-translations/?format=po
+	[[ -e $LANG_FILENAME.po ]] && msgfmt $LANG_FILENAME.po -o $LANG_FILENAME.mo
+done
```
markbiek, line 25: Can we add some error checking here? Part of the reason for automating the release process is catching errors that people often missed when going through the process manually.

Some possible things to check for:

1. The first `curl` call doesn't return anything to be translated.
2. Downloading one of the existing files fails.

markbiek (changes requested):

package.json
d-alleyne: I have converted the script to node, and it has very descriptive log messages

scripts/commands/update-translations.js
```diff
@@ -0,0 +1,84 @@
...
+	});
+}
+
+async function runScript() {
+	const git = gitFactory();
+
+	try {
+		info('Executing wp i18n make-pot command...');
+		await execPromise('wp i18n make-pot . languages/wc-calypso-bridge.pot --ignore-domain');
+		success('wp i18n make-pot command executed successfully.');
```
markbiek, line 32: How do we know that the `wp i18n make-pot` command executed successfully? Maybe we should check for the existence of the language file?

scripts/commands/update-translations.js
```diff
@@ -0,0 +1,84 @@
...
+				file.close();
+				resolve();
+			});
+		}).on('error', (err) => {
+			reject(err);
+		});
+	});
+}
+
+async function runScript() {
```
markbiek, line 26:
```suggestion
async function updateTranslations() {
```

scripts/commands/update-translations.js
```diff
@@ -0,0 +1,84 @@
...
+		success('Language files committed successfully.');
+
+		console.log('Script execution completed.');
+	} catch (err) {
+		error('Error executing the script: ' + err.message);
+		process.exit(1);
+	}
+}
+
+runScript();
```
markbiek, line 95:
```suggestion
updateTranslations();
```

scripts/commands/update-translations.js
```diff
@@ -0,0 +1,84 @@
...
+					data += chunk;
+				});
+				response.on('end', () => {
+					resolve(data);
+				});
+			}).on('error', (err) => {
+				reject(err);
+			});
+		});
+		success('Language data retrieved successfully.');
```
markbiek, line 54: We should probably do some verification here to make sure the data returned is what we're expecting.

markbiek (commented):
Everything is working nicely! Thanks for switching this over to node.

I left a couple of very minor comments. Also, can you please format this with prettier so it matches the rest of the code?

scripts/commands/update-translations.js
d-alleyne: Good idea!

scripts/commands/update-translations.js
d-alleyne: Done.

scripts/commands/update-translations.js
```diff
@@ -0,0 +1,170 @@
...
+}
+
+async function updateTranslations() {
+	const git = gitFactory();
+
+	try {
+		info( 'Executing wp i18n make-pot command...' );
+
+		const makePotResult = await execPromise(
+			`wp i18n make-pot . ${ POT_FILE_PATH } --ignore-domain`
```
dlind1, line 52: Currently `.gitignore` has `languages/*` in it, which would be removed.

scripts/commands/update-translations.js
```diff
@@ -0,0 +1,170 @@
...
+
+		const status = await git.status();
+
+		const isPotFileUpdated =
+			status.modified.includes( POT_FILE_PATH ) ||
+			status.created.includes( POT_FILE_PATH );
+
+		if ( ! isPotFileUpdated ) {
+			warning(
+				`${ POT_FILE_PATH } is up to date. No need to update translations.`
```
dlind1, line 70: As you've already suggested, it'd make sense not to exit here since we might have some new translations in GlotPress.

scripts/commands/update-translations.js
```diff
@@ -0,0 +1,170 @@
...
+				success( `${ LANG_FILENAME } compiled successfully.` );
+			} catch ( err ) {
+				error(
+					`Error downloading or compiling ${ LANG_FILENAME }: ${ err.message }`
+				);
+				process.exit( 1 );
+			}
+		}
+
+		info( 'Staging language files...' );
```
dlind1, line 155: It could happen that there won't be any changes here so adding/committing would fail. It would be better to do something like:
```
if (isPotFileUpdated) {
    info ( 'Committing new POT file...' );
    commit the file
}
if (otherLanguageFilesUpdated) {
    info ( 'Committing new translation files...' );
    commit the files
}
```

dlind1: Would `npm run makepot` be executed as part of a release process, or should anyone be able to run it? If it's the latter, then committing to the current branch might not be the best idea. Perhaps this script shouldn't commit anything at all and it'd be done in tag.js instead?

d-alleyne: I believe the goal is to make `npm run makepot` be part of the release process. @markbiek ?

scripts/commands/update-translations.js
d-alleyne: Done

scripts/commands/update-translations.js
d-alleyne: Done

markbiek (approved):
LGTM!

dlind1 (approved):
LGTM!
`wp i18n make-pot` would fail if translators comments are missing, but that sounds like a plus.

PR #1162, opened 2023-05-26
Add defensive code to OBW skip filter

daledupreez:

This PR adds some defensive checks to the code we use to add the `skipped` flag to the `woocommerce_onboarding_profile` option, as we had the following fatal reported in pe5sF9-1xN-p2#comment-2202:
```
PHP Fatal error:  Uncaught TypeError: Cannot access offset of type string on string in 
 
/wordpress/plugins/wpcomsh/3.12.11/vendor/automattic/
wc-calypso-bridge/includes/class-wc-calypso-bridge-skip-obw.php:80
```

Per the report, this should be reproducible on these sites by running the following:
1. Open WP-CLI for a Woo Express trial site
2. Run `wp option get active_plugins`
3. Verify that you don't get a fatal

**NOTE:** We're still investigating this, so it may be possible that the primary driver for this issue is bad data in the option, so some sites may not exhibit the behaviour.

### Other information:

### FOR PR REVIEWER ONLY:

### DIFF
--- a/includes/class-wc-calypso-bridge-skip-obw.php
+++ b/includes/class-wc-calypso-bridge-skip-obw.php
@@ -77,7 +77,13 @@ class WC_Calypso_Bridge_Skip_OBW {
 	 */
 	public function add_skipped_state( $option_value ) {
 		$value = $option_value ?? array();
-		$value['skipped'] = true;
+		if ( is_array( $value ) ) {
+			$value['skipped'] = true;
+		} elseif ( empty( $value ) ) {
+			$value = array(
+				'skipped' => true,
+			);
+		}
 		return $value;
 	}
 }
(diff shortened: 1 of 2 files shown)


ilyasfoo (commented):
Thanks for working on this, @daledupreez! I was able to confirm that the site does not crash when the option `woocommerce_onboarding_profile` has a string value.

I think it's worth to replace the corrupted option data with array and skipped value since it's not an invalid use of the option anyways. Otherwise, users on affected sites will suddenly get redirected to OBW.

daledupreez:
@ilyasfoo, I've updated the logic to return an array when we don't have a valid array to start with.

ilyasfoo (approved):
Awesome, thanks @daledupreez! Tested well 🚢 

Also, I just noticed that no new entry in changelog. Pre-approving!

PR #1172, opened 2023-06-01
Suppress WooCommerce Subscriptions move/duplicated site messages

### LINKED ISSUE #1048: Suppress subscriptions staging message
On new WooExpress sites, the users see that "it looks like this site has moved or is a duplicate site"

image

This doesn't make sense as it's a new site just created. We need to dig into this and discuss with Helix to help

Probably this affects a domain change too.

See related P2 comments peapX7-1V3-p2#comment-2200

PanosSynetos:

Closes #1048  .

- #1048

**Before syncing this PR to your site, make sure your store has a USA address, so that the notice pops up.**

Once you see the notice, run `wp option get wc_subscriptions_siteurl` . In case it returns something, delete it `wp option delete wc_subscriptions_siteurl`

The reason we delete it, is because new sites don't have this option, that's why the message appears

Checkout this PR and sync calypso-bridge.

- [x] When you visit the site, the staging notice should be gone
- [x] Executing `wp option get wc_subscriptions_siteurl` will return something similar to `https://devpanos.wpc_[wc_subscriptions_siteurl]_omstaging.com`
- [x] Executing `wp option get wc_calypso_bridge_one_time_operation_set_wc_subscriptions_siteurl_add_domain` should return `init` as this job will run forever until a domain is purchased
- [x] To test that the add domain job is complete, edit `class-wc-calypso-bridge-setup.php` and "fake" the site_url on line 553, something like `$site_url = 'panosisawesome.com';`
Execute `wp option get wc_calypso_bridge_one_time_operation_set_wc_subscriptions_siteurl_add_domain` and should return `complete`.
- [x] No staging messages should appear

### Other information:

### FOR PR REVIEWER ONLY:

### DIFF
--- a/includes/class-wc-calypso-bridge-setup.php
+++ b/includes/class-wc-calypso-bridge-setup.php
@@ -4,7 +4,7 @@
  *
  * @package WC_Calypso_Bridge/Classes
  * @since   1.0.0
- * @version x.x.x
+ * @version 2.1.7
  */
 
 use Automattic\WooCommerce\Admin\WCAdminHelper;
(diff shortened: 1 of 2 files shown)


xristos3490 (approved):
Tested well! Great work, @PanosSynetos!

Unrelated to this work: At some point, it will be necessary to refactor the one-time-operation controller into a separate controller to improve readability. However, this should be addressed later when we begin to abstract everything for the bundled version of Woo Express.

PR #1231, opened 2023-07-20
Prepare for release 2.2.3

d-alleyne:

Preparation for release of 2.2.3

### Changelog

```
= 2.2.3 =
* Enable the creation of translation files during release

```


scripts/commands/update-translations.js
```diff
@@ -52,11 +52,7 @@ export default async function updateTranslations() {
 			`wp i18n make-pot . ${ POT_FILE_PATH } --ignore-domain`
 		);
 
-		if ( makePotResult.stderr ) {
```
markbiek, line 55: I would love it if we can come up with some way to tell if the `wp i18n make-pot` command succeeded. Can we check the return code of the command or verify the output or something?

scripts/commands/update-translations.js
```diff
@@ -52,11 +52,7 @@ export default async function updateTranslations() {
...
-			error( `wp i18n make-pot failure: ${ makePotResult.stderr }` );
-			return false;
-		}
-
+		// If execPromise doesn't reject the promise, the command was successful
```
markbiek, line 55: Is this accurate? Can you provide more details?

d-alleyne: Here is what occurrs when the command returns an error
non-existent source directory

If you run the command in a directory that contains no files, there is no error - the POT file that is created simply has a header
```msgid ""
msgstr ""
"Project-Id-Version: WooCommerce Calypso Bridge 2.2.3\n"
"Report-Msgid-Bugs-To: https://wordpress.org/support/plugin/wc-calypso-bridge\n"
"Last-Translator: FULL NAME <EMAIL@ADDRESS>\n"
"Language-Team: LANGUAGE <LL@li.org>\n"
"MIME-Version: 1.0\n"
"Content-Type: text/plain; charset=UTF-8\n"
"Content-Transfer-Encoding: 8bit\n"
"POT-Creation-Date: 2023-07-20T18:54:30+00:00\n"
"PO-Revision-Date: YEAR-MO-DA HO:MI+ZONE\n"
"X-Generator: WP-CLI 2.6.0\n"```

File: scripts/commands/update-translations.js (hunk 1 above)
Comment (d-alleyne) [reply scripts/commands/update-translations.js:55]: Here is the output of the command the code update
code update

Comment (markbiek) [reply scripts/commands/update-translations.js:55]: Rather than always displaying that output, what if we only output it in the case where the `execPromise` has failed and we've fallen into the exception handler?

Something like this (untested)
```
} catch ( err ) {
    if ( makePotResult && makePotResult.stderr ) {
      warning( makePotResult.stderr );
    }
    return error( 'Error executing the script: ' + err.message );
}
```

File: scripts/commands/update-translations.js [hunk 3]
```diff
@@ -52,9 +52,12 @@ export default async function updateTranslations() {
 			`wp i18n make-pot . ${ POT_FILE_PATH } --ignore-domain`
 		);
 
+		// If execPromise doesn't reject the promise, the command was successful
```
markbiek, line 55: I would be tempted to move this comment to just before the `execPromise` call. It might also be noting in the comment that, if `execPromise` fails, an exception will be thrown.

scripts/commands/update-translations.js
d-alleyne: If `execPromise` has failed, then `makePotResult` doesn't exist. This can achieve similar:
```
} catch ( err ) {
	if ( err.stderr ) {
		warning( err.stderr )
	}
	return error( 'Error executing the script: ' + err.message );
}
```
Using a non-existent directory in the `make-pot` command produces the following result:

Screenshot 2023-07-21 at 1 01 21 PM

PanosSynetos:
Hey @d-alleyne @markbiek 

I needed to release a new version, and 2.2.3 was in the way, so I had to create a 2.2.4 version and release it.

- #1240 

I wanted to ask, is there a reason that the development and reviewing are happening on a release PR, instead of a usual PR? 

Thanks!

d-alleyne:
Sorry about that. The PR was created using the instructions found in the [Automated Deployment](https://github.com/Automattic/wc-calypso-bridge/blob/master/README.md#automated-deployment) method, rather than the [Manual Deployment](https://github.com/Automattic/wc-calypso-bridge/blob/master/README.md#manual-deployment).

PanosSynetos:
Thanks for the explanation Damien - The instructions you've quoted are for deployment purposes.

As we've chatted in Slack, the fix is a simple one and ready for release, that's why you followed this process.

Even for a simple fix a new PR should be created and then, once tested and merged, then a deployment PR is opened.

I'd recommend to close this PR, create a new one in a new branch with the same changes and ask someone from Serenity to review it, as Mark is AFK

d-alleyne:
work continued [here](https://github.com/Automattic/wc-calypso-bridge/pull/1242)

PR #1245, opened 2023-07-27
Prepare for release 2.2.5

d-alleyne:

Preparation for release of 2.2.5

### Changelog

```
= 2.2.5 =
* Enable downloading updated translation files

```


daledupreez (approved):
The changes look good to me!

@d-alleyne, could you confirm what expectations there are for the developers in the repo to maintain/update translations?

daledupreez:
Thanks for clarifying, @d-alleyne!

My only concern with that process/requirement is that I don't believe the current repo configuration installs `wp` (or makes it available on the path) -- I think we need to prioritise fixing that before we require a workflow that will fail if this unspecified dependency isn't met.

PR #1272, opened 2023-08-15
Add text domain linter

d-alleyne:

Relates to https://github.com/Automattic/i18n-issues/issues/594 and https://github.com/Automattic/wc-calypso-bridge/pull/1127

This PR is a follow-up to #1127. Users will now detect linting errors prior to committing the code to the repo. Catching linting errors on the local machine provides immediate feedback, thus leading to faster development cycles.

1. Checkout the repo and create a new branch (e.g. git checkout -b <new branch>)
2. Edit [line 32](https://github.com/Automattic/wc-calypso-bridge/blob/9f08796b234c81913749e97ce03e563182ed9c0c/includes/tasks/class-wc-calypso-task-add-domain.php#L32) of `includes/tasks/class-wc-calypso-task-add-domain.php`, and change the text domain to 'jetpack'
3. Attempt to commit the change (`git commit -a`)
4. Verify that an error occurs
5. Undo the change

### Other information:

### FOR PR REVIEWER ONLY:


bin/lint-php-changed.sh
```diff
@@ -0,0 +1,9 @@
+#!/usr/bin/env bash
+DEFAULT_BRANCH=$(git symbolic-ref refs/remotes/origin/HEAD | sed 's@^refs/remotes/origin/@@')
+INITIAL_COMMIT_FOR_BRANCH=$(git merge-base origin/$DEFAULT_BRANCH HEAD)
+CHANGED_PHP_FILES=$(git diff --name-only --diff-filter=d $INITIAL_COMMIT_FOR_BRANCH | grep '\.php$' | xargs)
+if [ -z "$CHANGED_PHP_FILES" ]; then
+	echo "No PHP files changed. Exiting."
+	exit 0
+fi
+vendor/bin/phpcs-changed --warning-severity=0 --extensions=php,html --git --git-staged $CHANGED_PHP_FILES
```
daledupreez, line 9: I have a concern about the exact way this code is set up, mainly because the call to `phpcs-changed` is passing in `--git-staged`, but the logic above is finding all PHP files that have been changed on the current branch. 

I also notice that we're including `html` in the `--extensions` list despite limiting the linter to files that end in `.php` above.

If we're wanting to run this only for staged files, then it should be enough to have the changed files computed via the `--staged` argument to `git diff`. If we only want this for staged files, then it may be worth moving my suggestion directly into `.husky/pre-commit`. But if we want the code to work more broadly, we should consider pushing the changed file detection into separate scripts for basic approaches. e.g. for pre-push, we would probably want to consider all changes on the branch.
```suggestion
CHANGED_STAGED_PHP_FILES=$(git diff --staged --name-only --diff-filter=d | grep '\.php$' | xargs)
if [ -z "$CHANGED_STAGED_PHP_FILES" ]; then
	echo "No modified PHP files are staged. Exiting."
	exit 0
fi
vendor/bin/phpcs-changed --warning-severity=0 --extensions=php --git --git-staged $CHANGED_STAGED_PHP_FILES
```

d-alleyne: I observed that there are numerous other linters in the `package.json` file, which leads me to believe that keeping this code outside of `.husky/pre-commit` would be a better approach. If the team decides to do so, other linters (such as `npm run lint:js`) can be easily incorporated as additional lines within the `.husky/pre-commit` file.

daledupreez: My bigger concern is that there is nothing in the naming to suggest that it will only operate on staged files. In general, linters will run on the files you give them, but this particular script will only lint staged files. Can you update the script name and alias to reflect that important distinction? Maybe `bin/lint-php-staged.sh` and `lint:php:staged`?

daledupreez (approved):
Tested and LGTM! Thanks for sticking with me on the rename, I think it's going to be much better in the long run for developers.

PR #1278, opened 2023-08-22
Convert plugins page to a WC page

### LINKED ISSUE #1275: Add telemetry for the new /plugins page
The new /plugins page doesn't have any telemetry so we're not able to understand how it is impacting conversions.

We should just replicate how we track the Extensions page:

- Add a wcadmin_plugins_view event (On the extensions page we have https://mc.a8c.com/tracks/live/?eventname=wcadmin_extensions_view)
- Fire the wcadmin_page_view event too
- Implement tracking on the various CTAs on the page.

moon0326:

Closes #1275  .

This PR converts `/plugins` (WP Core page) to a WC page to record tracks automatically and make the UI consistent.

I wasn't able to fully test the visual changes locally since we miss some CSS that are only available on a WPCOM site. Please use your WoA site and transfer the files in order to test.

1. Prepare a WoA site.
2. Checkout this branch in your WoA site.
3. Go to your WoA site.
4. Confirm `Plugins` menu is rendered.
5. Click `Plugins` menu.
6. Confirm the page is rendered correctly.
7. Open your browser console and execute `window.localStorage.setItem('debug', 'wc-admin:*')` and make sure to change log level to verbose.
8. Click buttons and links and make sure tracks are fired.
9. Confirm `wcadmin_page_view` track gets fired.
10. Click `Network` tab and search for `t.gif?path=plugins-upgrade` and click it.
11. Confirm `_en` parameter value is `wcadmin_page_view`

### Other information:

### FOR PR REVIEWER ONLY:


src/index.js
```diff
@@ -177,6 +181,16 @@ if ( !! window.wcCalypsoBridge.isEcommercePlan ) {
 					}
 					return page;
 				} );
+
+				pages.push( {
+					container: Plugins,
+					path: '/plugins-upgrade',
+					breadcrumbs: [ 'Plugins', 'Upgrade' ],
```
ilyasfoo, line 188: I'm unsure if we need `Upgrade`, but I think we need the menu to display `Plugins`

image

ilyasfoo (changes requested):
Thanks for working on this, @moon0326! Minor comments but otherwise testing well!

Also, please add changelog in the unreleased section for this PR 🙏

src/index.js
moon0326: Thank you for checking typos 🙇 

I'll update the breadcrumb to `Plugins` 👍

moon0326: @ilyasfoo Updated in https://github.com/Automattic/wc-calypso-bridge/pull/1278/commits/ec4a373bb96500dab748c5bf61ae01d604c21308

src/index.js
```diff
@@ -177,6 +182,16 @@ if ( !! window.wcCalypsoBridge.isEcommercePlan ) {
 					}
 					return page;
 				} );
+
+				pages.push( {
+					container: Plugins,
+					path: '/plugins-upgrade',
+					breadcrumbs: [ __( 'Plugins' ), __( 'Plugins' ) ],
```
ilyasfoo, line 189: Missing translation domain

```suggestion
					breadcrumbs: [ __( 'Plugins', 'wc-calypso-bridge' ), __( 'Plugins', 'wc-calypso-bridge' ) ],
```

assets/css/ecommerce.css
```diff
@@ -37,13 +37,13 @@ body.woocommerce_page_wc-bridge-landing-page #screen-meta-links {
 }
 
 .wc-bridge-landing-page__title {
```
xristos3490, line 39: How about introducing a more specific rule instead of using `!important` here? Such as the `.woocommerce.wc-bridge-landing-page .wc-bridge-landing-page__title`?

xristos3490 (commented):
Nicely done @moon0326! 

I notice one thing different from your validation steps. That's the `_en` param which is set to `wcadmin_extensions_view` instead of `wcadmin_page_view` See 

I've also added a minor comment on the CSS part.

assets/css/ecommerce.css
ilyasfoo: Sounds good! Made changes in https://github.com/Automattic/wc-calypso-bridge/pull/1278/commits/9860d7ee171ac02a1f1a1f87e4c576b68092904a

ilyasfoo:
> I notice one thing different from your validation steps. That's the _en param which is set to wcadmin_extensions_view instead of wcadmin_page_view See

Good catch, @xristos3490! I've added the track in https://github.com/Automattic/wc-calypso-bridge/pull/1278/commits/9860d7ee171ac02a1f1a1f87e4c576b68092904a

image

xristos3490 (approved):
Thanks for sorting this @ilyasfoo! 

Apart from the fact that I now see two events firing, one with `_en` `wcadmin_page_view` and one with `wcadmin_plugins_view`, it tested well! 

If there is a specific reason for the second (`wcadmin_plugins_view`) event, I think we are good! If not, I'd suggest removing the specific `wcadmin_plugins_view` event and keeping the native one. Thoughts?

ilyasfoo:
> If there is a specific reason for the second (wcadmin_plugins_view) event, I think we are good! If not, I'd suggest removing the specific wcadmin_plugins_view event and keeping the native one. Thoughts?

Tbh, I was replicating what extensions page was doing https://github.com/Automattic/wc-calypso-bridge/issues/1275#issue-1858674640, but I think you're right that it's actually redundant. 28e30af

Thanks for looking into it!

ilyasfoo (approved):
Merging

PR #1318, opened 2023-10-03
Use the wc-admin version of the marketplace under Extensions > Discover menu item

### LINKED ISSUE #1317: Use the wc-admin version of the marketplace under Extensions > Discover menu item
### Description

The WooCommerce Marketplace is actively getting a revamp, and it's transferred to the WC Admin router.

The legacy URL is: `wp-admin/admin.php?page=wc-addons`, and the new one is: `wp-admin/admin.php?page=wc-admin&path=%2Fextensions`

We must change the URL and ensure that the correct item is re-ordered and renamed "Discover".

**Notes:**
- The "My subscriptions" page (in WX, the "Extensions > Manage") needs to remain as _is_.

xristos3490:
### Specification

This PR aims to refactor the ecommerce menu structure to accommodate the new rectified marketplace version. 

#### Notes on the changes:
- A new `WC_Calypso_Bridge_Addons::get_menu_slug()` method is introduced that would return the menu slug to use based on the context. The new marketplace will only be used when (a) the feature is enabled and (b) the site is not a free trial site. The latter is required because the landing page is hosted on the legacy page. 
- The `get_menu_slug()` helper is later used in the menu controller to identify the item that needs renaming and reordering.

Closes #1317 

### Validation

Please ensure that the changeset in this https://github.com/woocommerce/woocommerce/pull/40549 is included in your WC installation before testing.

To test the changes on WX sites follow the steps below:

**Testing on Ecommerce/Essential/Performance WX sites with the feature enabled:**

- Use a WX site (not a free trial.)
- Check that the "Extensions > Discover" menu item appears twice, and both point to the legacy page _(wp-admin/admin.php?page=wc-addons)_
- Apply this patch.
- Ensure that you only see one "Extensions > Discover" item, and it points to the rectified version.
- Ensure that the "Extensions > Manage" continues to work as expected.

**Testing on Ecommerce/Essential/Performance WX sites with the feature disabled:**

- Now, on the same site, head over to "Settings > WooCommerce > Advanced > Features" and disable the "Marketplace" feature. 
- Visit the "Extensions > Discover" item and ensure the legacy template gets loaded as expected. 
- Ensure that the "Extensions > Manage" page is still working.

**Testing on WX free trials:**

- Use a WX free trial
- Visit the "Extensions > Discover" item and ensure the landing page is loaded correctly.
- Head over to "Settings > WooCommerce > Advanced > Features" and disable the "Marketplace" feature.
- Visit the "Extensions > Discover" item and ensure the landing page is loaded correctly _again_.
- Ensure that the "Extensions > Manage" page is still working.
- Manually navigate to `wp-admin/admin.php?page=wc-admin&path=%2Fextensions` and ensure you cannot access the page while in the free trial.

### DIFF
--- a/includes/class-wc-calypso-bridge-addons.php
+++ b/includes/class-wc-calypso-bridge-addons.php
@@ -5,8 +5,10 @@
  *
  * @package WC_Calypso_Bridge/Classes
  * @since   1.0.6
- * @version 2.0.0
+ * @version x.x.x
  */
+
+use Automattic\WooCommerce\Utilities\FeaturesUtil;
 
 defined( 'ABSPATH' ) || exit;
 
@@ -54,9 +56,11 @@ class WC_Calypso_Bridge_Addons {
 	 * Initialize.
 	 */
 	public function init() {
+
+		// Hide the default marketplace.
 		add_filter( 'woocommerce_show_addons_page', '__return_false' );
+		// Handle the addons legacy page.
 		add_action( 'admin_menu', array( $this, 'addons_menu' ), 70 );
-		add_action( 'woocommerce_loaded', array( $this, 'load_modified_addons_menu' ) );
 
 		// Add admin body class for the free trial landing page.
 		if ( wc_calypso_bridge_is_ecommerce_trial_plan() ) {
@@ -72,11 +76,8 @@ class WC_Calypso_Bridge_Addons {
 		}
 	}
 
-	/**
-	 * Load the class for the modified Addons menu.
-	 */
-	public function load_modified_addons_menu() {
-		require 'class-wc-modified-admin-addons.php';
+	public function get_menu_slug() {
+		return class_exists( '\Automattic\WooCommerce\Utilities\FeaturesUtil' ) && FeaturesUtil::feature_is_enabled( 'marketplace' ) && ! wc_calypso_bridge_is_ecommerce_trial_plan() ? 'wc-admin&path=/extensions' : 'wc-addons';
 	}
 
 	/**
--- a/includes/class-wc-calypso-bridge-addons.php
+++ b/includes/class-wc-calypso-bridge-addons.php
@@ -8,8 +8,6 @@
  * @version x.x.x
  */
 
-use Automattic\WooCommerce\Utilities\FeaturesUtil;
-
 defined( 'ABSPATH' ) || exit;
 
 /**
@@ -77,7 +75,7 @@ class WC_Calypso_Bridge_Addons {
 	}
 
 	public function get_menu_slug() {
-		return class_exists( '\Automattic\WooCommerce\Utilities\FeaturesUtil' ) && FeaturesUtil::feature_is_enabled( 'marketplace' ) && ! wc_calypso_bridge_is_ecommerce_trial_plan() ? 'wc-admin&path=/extensions' : 'wc-addons';
+		return class_exists( '\Automattic\WooCommerce\Utilities\FeaturesUtil' ) && \Automattic\WooCommerce\Utilities\FeaturesUtil::feature_is_enabled( 'marketplace' ) && ! wc_calypso_bridge_is_ecommerce_trial_plan() ? 'wc-admin&path=/extensions' : 'wc-addons';
 	}
 
 	/**
--- a/includes/class-wc-calypso-bridge-ecommerce-admin-menu.php
+++ b/includes/class-wc-calypso-bridge-ecommerce-admin-menu.php
@@ -295,7 +295,7 @@ class Ecommerce_Atomic_Admin_Menu extends \Automattic\Jetpack\Dashboard_Customiz
 			);
 		}
 
-		if ( class_exists( '\Automattic\WooCommerce\Utilities\FeaturesUtil' ) && \Automattic\WooCommerce\Utilities\FeaturesUtil::feature_is_enabled( 'analytics' ) ) {
+		if ( class_exists( '\Automattic\WooCommerce\Admin\Features\Features' ) && \Automattic\WooCommerce\Admin\Features\Features::is_enabled( 'analytics' ) ) {
 			// Move Customers to root menu.
 			$this->hide_submenu_page( 'woocommerce', 'wc-admin&path=/customers' );
 			add_menu_page( __( 'Customers', 'woocommerce' ), __( 'Customers', 'woocommerce' ), 'manage_woocommerce', 'admin.php?page=wc-admin&path=/customers', null, 'dashicons-money', 100 );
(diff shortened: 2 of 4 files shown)


xristos3490:
I'm marking this as blocked till the https://github.com/woocommerce/woocommerce/pull/40549 gets merged and released

includes/class-wc-calypso-bridge-addons.php
```diff
@@ -72,11 +74,8 @@ public function init() {
 		}
 	}
 
-	/**
-	 * Load the class for the modified Addons menu.
-	 */
-	public function load_modified_addons_menu() {
-		require 'class-wc-modified-admin-addons.php';
+	public function get_menu_slug() {
+		return class_exists( '\Automattic\WooCommerce\Utilities\FeaturesUtil' ) && \Automattic\WooCommerce\Utilities\FeaturesUtil::feature_is_enabled( 'marketplace' ) && ! wc_calypso_bridge_is_ecommerce_trial_plan() ? 'wc-admin&path=/extensions' : 'wc-addons';
```
andfinally, line 78: We'll need to remember to change this when we eventually make the new style marketplace permanent.

includes/class-wc-calypso-bridge-ecommerce-admin-menu.php
```diff
@@ -292,7 +295,7 @@ public function add_woocommerce_menu() {
 			);
 		}
 
-		if ( class_exists( '\Automattic\WooCommerce\Admin\Features\Features' ) && \Automattic\WooCommerce\Admin\Features\Features::is_enabled( 'analytics' ) ) {
+		if ( class_exists( '\Automattic\WooCommerce\Utilities\FeaturesUtil' ) && \Automattic\WooCommerce\Utilities\FeaturesUtil::feature_is_enabled( 'analytics' ) ) {
```
andfinally, line 298: I'm not 100% clear if the new (p7bje6-4kh-p2) and [old](https://github.com/woocommerce/woocommerce/blob/9ea83d6382580d106e818dd0cff7c48e8e173359/plugins/woocommerce-admin/docs/features/feature-flags.md) feature declarations are compatible with each other. I don't see any examples of the `analytics` feature using the `FeaturesUtil` method, only `Features::is_enabled` calls.

includes/class-wc-calypso-bridge-ecommerce-admin-menu.php
```diff
@@ -75,6 +77,7 @@ protected function handle_free_trial_menu() {
 
 			// Hide Extensions > Manage.
 			$this->hide_submenu_page( 'woocommerce', 'admin.php?page=wc-addons&section=helper' );
+			$this->hide_submenu_page( 'woocommerce', 'wc-admin&path=/extensions' );
```
andfinally, line 80: Maybe we can spell out in the comment `Hide Extensions > Manage and the new Extensions page.`

andfinally (commented):
Looks good to me, thank you! 🙌  Tested the `woocommerce_show_addons_page` filter works with the changes in https://github.com/woocommerce/woocommerce/pull/40549/. I've now merged that change, and I've asked the release team to include it in the next weekly release. They say it'll have to go out in next week's release on 10 October.

includes/class-wc-calypso-bridge-addons.php
xristos3490: Noted! I think this would be quite relevant to how we decide to proceed with this feature checkbox. As far as I understand, we might want to:
- Permanently fix the marketplace in place and lose the checkbox, or;
- Handle the fallback which currently shows the legacy one, and instead hide it completely when the feature is disabled.

In terms of dotCom, we could force-enable the feature in all Woo Express sites (or even on all plans.)

includes/class-wc-calypso-bridge-ecommerce-admin-menu.php
xristos3490: Great catch! Indeed, this isn't working with the `analytics` feature! Reverted https://github.com/Automattic/wc-calypso-bridge/pull/1318/commits/976722331efd87836d53437ae352a1374876d295. Thanks 🙌

includes/class-wc-calypso-bridge-ecommerce-admin-menu.php
xristos3490: Done in https://github.com/Automattic/wc-calypso-bridge/pull/1318/commits/d4354a1134850435aea931121a16fe828e5dc800

PanosSynetos (approved):
Manually tested all the scenarios and we're almost good :)

The only thing I noticed, is that on the Trial testing instruction @xristos3490 you mention to test

> Ensure that the "Extensions > Manage" page is still working.

I guess that this step was a copy paste leftover, as in trial sites, we don't have a manage menu item.

Anywoo, I visited `wp-admin/admin.php?page=wc-addons&section=helper` which showed the `Take your store to the next level, with extensions` page.

andfinally:
Thanks for the tweaks! My Woo PR is merged and should be included in this week's weekly release.

I'm just wondering about the best way to test these changes. I guess I should install Woo core nightly release in a local WP instance and follow the steps in [Development Environment](https://github.com/Automattic/wc-calypso-bridge#development-environment).

PanosSynetos:
Hey @andfinally - I don't think calypso bridge would work on a local WP instance. The recommended way would be to have a WoA dev site and this guide pdqkbG-2Ma-p2

If it's too much to create two dev sites (one upgraded, and one free trial) I can invite you on my dev sites and test things out.

The only thing you won't be able to test is, the duplicate menu item - but we've established that it happens without this PR applied :D

andfinally (approved):
Thanks for the tweaks, and thanks Panos for the loan of your environments! 💯 Looks good and works as expected – like Panos I also don't get the "Extensions > Manage" page on the trial site, which I'm sure is intentional.

FYI we'll have some more changes coming fairly soon:

* Firstly, we're adding themes to the marketplace. This should just work on Woo Express.
* After that, we'll be releasing a rebuilt My Subscriptions (a.k.a. Manage) page with a new URL: `admin.php?page=wc-admin&path=%2Fextensions&tab=my-subscriptions`. This will replace the current page at `admin.php?page=wc-addons&section=helper`.

xristos3490:
Thanks for the review, everyone! 

> I guess that this step was a copy paste leftover, as in trial sites, we don't have a manage menu item.
> like Panos I also don't get the "Extensions > Manage" page on the trial site, which I'm sure is intentional.

Yes, nice catch! I actually didn't write the "what's expected" here, which is to see the landing page instead of 404. The step ensures that the rewrite rule for the page still exists.

> After that, we'll be releasing a rebuilt My Subscriptions (a.k.a. Manage) page with a new URL: admin.php?page=wc-admin&path=%2Fextensions&tab=my-subscriptions. This will replace the current page at admin.php?page=wc-addons&section=helper.

Good to know – That would basically retire the `wc-addons` wp-admin page and require some changes here. We can sync the releases to mitigate the risks!

PR #1377, opened 2023-12-04
Removed the empty Product Data > Get more options tab

jimjasson:
### Changes proposed in this Pull Request:

This PR removes the **Product Data > Get more options** tab, as it is empty -- we do not offer any Marketplace suggestions in Woo Express/Ecommerce sites.

Before the fix: 

After the fix: 

Closes #1333 

- [ ] This PR is a very minor change/addition and does not require testing instructions (if checked you can ignore/remove the next section).

### How to test the changes in this Pull Request:

- [x] Have you followed the [Writing high-quality testing instructions guide](https://github.com/woocommerce/woocommerce/wiki/Writing-high-quality-testing-instructions)?

1. Open a new site -- the plan doesn't change this behavior.
2. Create a new product.
3. Notice the "Get more options" tab under **Product Data**.
4. Click on it and make sure it is empty. 
5. Apply the patch. 
6. Refresh the page -- the "Get more options" tab should now be removed. 

### Other information:

-   [x] Have you added an explanation of what your changes do and why you'd like us to include them?
-   [ ] Have you written new tests for your changes, as applicable?

### FOR PR REVIEWER ONLY:

-   [x] I have reviewed that everything is sanitized/escaped appropriately for any SQL or XSS injection possibilities. I made sure Linting is not ignored or disabled.

PR #1395, opened 2023-12-13
Revert #1377 and fix init priority to properly hide the empty get more options tab

### LINKED ISSUE #1394: Revert PR 1377 empty Product Data > Get more options tab and improve solution
As discussed with @manospsyx and @jimjasson , let's revert

- #1377 

and improve the solution, by making sure the `woocommerce_allow_marketplace_suggestions` used elsewhere, works properly

PanosSynetos:

Discussed with @jimjasson about this, and he correctly pointed out that there is race condition.

WooCommerce is hooking class `WC_Marketplace_Suggestions` on `init` on priority 10, and we also hook with priority 10.

We need to lower the priority on our class

Closes #1394  .
- #1394 

See testing instructions 👇 
- https://github.com/Automattic/wc-calypso-bridge/pull/1377

### Future discussion

Should we treat the same `init` with lower priorities? If so, let's open a new PR to address them

image

### DIFF
--- a/includes/class-wc-calypso-bridge-filters.php
+++ b/includes/class-wc-calypso-bridge-filters.php
@@ -4,7 +4,7 @@
  *
  * @package WC_Calypso_Bridge/Classes
  * @since   1.1.6
- * @version 2.2.26
+ * @version x.x.x
  */
 
 defined( 'ABSPATH' ) || exit;
@@ -65,11 +65,6 @@ class WC_Calypso_Bridge_Filters {
 		 * Filter recommended themes
 		 */
 		add_filter( '__experimental_woocommerce_rest_get_recommended_themes', array( $this, 'woocommerce_filter_get_recommended_themes'), 10, 3);
-
-		/**
-		 * Remove the empty Product Data > Get more options tab.
-		 */
-		add_filter( 'woocommerce_product_data_tabs', array( $this, 'remove_marketplace_suggestions_product_tab' ) );
 	}
 
 	/**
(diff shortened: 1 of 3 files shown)


jimjasson (approved):
I agree 100% with this fix as core also hooks at `init` with priority `10` to add the marketplace suggestions -- we need to hook in sooner to remove them.

jimjasson:
> Should we treat the same init with lower priorities? If so, let's open a new PR to address them

I think that this depends! For example, in this case we were able to identify that core hooks into `init` with priority `10` to add the marketplace suggestions and we were hooking in the same hook with the same priority to remove them. Is this the case for other things we do too? Like the `update_option` you mention in the example -- does core introduce this option at the same hook/priority where we are updating it?

PanosSynetos:
The `update_option` was misleading in my screenshot, sorry - It's not an `add_action`

I'm only referring to the `add_action`. I believe that when we use this practice to add an `add_action` in the constructor, hooked on `init`, it needs to be really early in priority.

Then the rest of the `add_action` , `add_filter` inside the init function, should have their own priority (ideally really late, PHP_INT_MAX)

I'm not saying that's the only way going forward, it's a discussion. Looking forward for @xristos3490  point of view (even @manospsyx if/when he gets a chance)


```php
...
		add_action( 'init', array( $this, 'init' ), 1 );
	}

	public function init() {
...
		add_filter( 'woocommerce_data_stores', array( $this, 'filter_notes_data_store' ), 100 );
		add_action( 'admin_head', array( $this, 'hide_alerts_on_non_settings_pages' ) );
		add_filter( 'woocommerce_helper_suppress_connect_notice', '__return_true' );
		add_filter( 'woocommerce_show_admin_notice', '__return_false' );
		add_filter( 'woocommerce_allow_marketplace_suggestions', '__return_false' );
	}
```

xristos3490 (approved):
Changes are looking safe. Approved!🤞 

Regarding this subject:
> Should we treat the same init with lower priorities? If so, let's open a new PR to address them

I'd opt to keep them as is, until there is a reason to change them.

PR #1397, opened 2023-12-15
Render introductory offer banner on Woo Home

### LINKED ISSUE #1320: [Woo Express $1 offer] Homescreen banner
## Issue description

Image

Figma: jFG3dxNDOdMkw8AC1aGWBL-fi-458_1884

Homescreen banner is customized for $1 offers.

## Acceptance criteria

```
Feature: Homescreen Banner

Scenario: Display new homescreen banner for eligible users
  Given the user is eligible for the $1 offer
  And the new homescreen banner is not dismissed by the user
  When the user navigates to the homescreen
  Then the new homescreen banner for the $1 offer is displayed

Scenario: Display old homescreen banner for non-eligible users
  Given the user is not eligible for the $1 offer
  And the old homescreen banner is not dismissed by the user
  When the user navigates to the homescreen
  Then the old homescreen banner is displayed

Scenario: Dismiss the new homescreen banner
  Given the new homescreen banner for the $1 offer is displayed
  When the user clicks the dismiss option on the new banner
  Then the new banner should be dismissed
  And the dismiss action used should be unique for the new banner

Scenario: Navigate to the plans page
  Given the new homescreen banner for the $1 offer is displayed
  When the user clicks the "Upgrade now" button on the banner
  Then navigate to plans page (ex: https://wordpress.com/plans/sitenamehere)
```

moon0326:

Closes #1320

This PR includes code to render a banner with introductory offer copy when an introductory offer is available.

- Adds banner component https://github.com/Automattic/wc-calypso-bridge/commit/931c617b2518fdbe964064713fe0319fd64869a1
- Queries plans data from WPCOM https://github.com/Automattic/wc-calypso-bridge/commit/87b173b45da722ab7ee8fbe0bedef6d75876c7b2
 

1. Set up a new trial site with SSH access.
2. Sync this branch to your trial site.
3. Navigate to WooCommerce -> Home.
4. Confirm the introductory offer banner is rendered.

Screenshot 2023-12-14 at 6 06 34 PM

### Other information:

### FOR PR REVIEWER ONLY:

### DIFF
--- a/class-wc-calypso-bridge-shared.php
+++ b/class-wc-calypso-bridge-shared.php
@@ -109,6 +109,7 @@ class WC_Calypso_Bridge_Shared {
 		$site_suffix = WC_Calypso_Bridge_Instance()->get_site_slug();
 		$is_ecom_trial = (bool) wc_calypso_bridge_is_ecommerce_trial_plan();
 
+
 		$params      = array(
 			'isEcommercePlan'              => (bool) wc_calypso_bridge_has_ecommerce_features(),
 			'isEcommercePlanTrial'         => $is_ecom_trial, // This is true for ecommerce trial only.
@@ -125,9 +126,12 @@ class WC_Calypso_Bridge_Shared {
 		);
 
 		if ( $is_ecom_trial ) {
-			$introductory_offers = WC_Calypso_Bridge_Introductory_offers::get_introductory_offers_for_current_blog();
-			if ( count( $introductory_offers ) ) {
-				$params['introductoryOffer'] = WC_Calypso_Bridge_Introductory_offers::extract_offer_data_for_js( current( $introductory_offers ) );
+			$woo_express_introductory_offers = WC_Calypso_Bridge_Woo_Express_Introductory_offers::get_offers_for_current_blog();
+			if ( count( $woo_express_introductory_offers ) ) {
+				$woo_express_introductory_offers = array_filter( $woo_express_introductory_offers, function( $offer ) {
+					return in_array( $offer['product_slug'], WC_Calypso_Bridge_Woo_Express_Introductory_offers::WOO_EXPRESS_PRODUCT_SLUGS );
+				} );
+				$params['wooExpressIntroductoryOffer'] = WC_Calypso_Bridge_Woo_Express_Introductory_offers::extract_offer_data_for_js( current( $woo_express_introductory_offers ) );
 			}
 		}
 
--- a/class-wc-calypso-bridge-shared.php
+++ b/class-wc-calypso-bridge-shared.php
@@ -126,14 +126,14 @@ class WC_Calypso_Bridge_Shared {
 		);
 
 		if ( $is_ecom_trial ) {
-			$woo_express_introductory_offers = WC_Calypso_Bridge_Woo_Express_Introductory_offers::get_offers_for_current_blog();
+			$woo_express_introductory_offers = WC_Calypso_Bridge_Woo_Express_Introductory_offers::get_offers_for_current_blog( function( $offer ) {
+				return in_array( $offer['product_slug'], WC_Calypso_Bridge_Woo_Express_Introductory_offers::WOO_EXPRESS_PRODUCT_SLUGS );
+			} );
 			if ( count( $woo_express_introductory_offers ) ) {
-				$woo_express_introductory_offers = array_filter( $woo_express_introductory_offers, function( $offer ) {
-					return in_array( $offer['product_slug'], WC_Calypso_Bridge_Woo_Express_Introductory_offers::WOO_EXPRESS_PRODUCT_SLUGS );
-				} );
 				$params['wooExpressIntroductoryOffer'] = WC_Calypso_Bridge_Woo_Express_Introductory_offers::extract_offer_data_for_js( current( $woo_express_introductory_offers ) );
 			}
 		}
+		
 
 		if ( wc_calypso_bridge_has_ecommerce_features() ) {
 			$params['showEcommerceNavigationModal'] = ! WC_Calypso_Bridge_Helper_Functions::is_wc_admin_installed_gte( WC_Calypso_Bridge::RELEASE_DATE_ECOMMERCE_NAVIGATION );
--- a/class-wc-calypso-bridge-shared.php
+++ b/class-wc-calypso-bridge-shared.php
@@ -125,7 +125,7 @@ class WC_Calypso_Bridge_Shared {
 		);
 
 		if ( $is_ecom_trial ) {
-			$introductory_offers = WC_Calypso_Bridge_Introductory_offers::get_current_offers_from_current_blog();
+			$introductory_offers = WC_Calypso_Bridge_Introductory_offers::get_introductory_offers_for_current_blog();
 			if ( count( $introductory_offers ) ) {
 				$params['introductoryOffer'] = WC_Calypso_Bridge_Introductory_offers::extract_offer_data_for_js( current( $introductory_offers ) );
 			}
--- a/class-wc-calypso-bridge-shared.php
+++ b/class-wc-calypso-bridge-shared.php
@@ -107,10 +107,11 @@ class WC_Calypso_Bridge_Shared {
 		);
 
 		$site_suffix = WC_Calypso_Bridge_Instance()->get_site_slug();
+		$is_ecom_trial = (bool) wc_calypso_bridge_is_ecommerce_trial_plan();
 
 		$params      = array(
 			'isEcommercePlan'              => (bool) wc_calypso_bridge_has_ecommerce_features(),
-			'isEcommercePlanTrial'         => (bool) wc_calypso_bridge_is_ecommerce_trial_plan(), // This is true for ecommerce trial only.
+			'isEcommercePlanTrial'         => $is_ecom_trial, // This is true for ecommerce trial only.
 			'isWooExpress'                 => (bool) wc_calypso_bridge_is_woo_express_plan(),
 			'isWooNavigationEnabled'       => (bool) apply_filters( 'ecommerce_new_woo_atomic_navigation_enabled', 'yes' === get_option( 'wooexpress_navigation_enabled', 'yes' ) ) && class_exists('\Jetpack') && \Jetpack::is_module_active( 'sso' ),
 			'isWooPage'                    => $is_woo_page,
@@ -122,6 +123,13 @@ class WC_Calypso_Bridge_Shared {
 			'hasViewedPayments'            => get_option( 'wc_calypso_bridge_payments_view_welcome_timestamp', false ) !== false,
 			'version'                      => WC_CALYPSO_BRIDGE_CURRENT_VERSION,
 		);
+
+		if ( $is_ecom_trial ) {
+			$introductory_offers = WC_Calypso_Bridge_Introductory_offers::get_current_offers_from_current_blog();
+			if ( count( $introductory_offers ) ) {
+				$params['introductoryOffer'] = WC_Calypso_Bridge_Introductory_offers::extract_offer_data_for_js( current( $introductory_offers ) );
+			}
+		}
 
 		if ( wc_calypso_bridge_has_ecommerce_features() ) {
 			$params['showEcommerceNavigationModal'] = ! WC_Calypso_Bridge_Helper_Functions::is_wc_admin_installed_gte( WC_Calypso_Bridge::RELEASE_DATE_ECOMMERCE_NAVIGATION );
--- a/includes/class-wc-calypso-bridge-introductory-offers.php
+++ b/includes/class-wc-calypso-bridge-introductory-offers.php
@@ -3,8 +3,8 @@
  * Contains the logic for querying plans with introductory offers.
  *
  * @package WC_Calypso_Bridge/Classes
- * @since   1.0.6
- * @version 2.0.8
+ * @since   x.x.x
+ * @version x.x.x
  */
 
 use Automattic\Jetpack\Connection\Client;
--- a/includes/class-wc-calypso-bridge-introductory-offers.php
+++ b/includes/class-wc-calypso-bridge-introductory-offers.php
@@ -22,7 +22,7 @@ class WC_Calypso_Bridge_Introductory_offers {
 	 *
 	 * @return array|mixed
 	 */
-	public static function get_current_offers_from_current_blog() {
+	public static function get_introductory_offers_for_current_blog() {
 		return static::get_introductory_offers( Jetpack_Options::get_option( 'id' ) );
 	}
 
--- a/includes/class-wc-calypso-bridge-introductory-offers.php
+++ b/includes/class-wc-calypso-bridge-introductory-offers.php
@@ -18,7 +18,7 @@ class WC_Calypso_Bridge_Introductory_offers {
 	CONST TRANSIENT_PREFIX = 'wc-calypso-bridge-introductory-plans-';
 
 	/**
-	 * return introductory offers for the current blog.
+	 * Return introductory offers for the current blog.
 	 *
 	 * @return array|mixed
 	 */
--- a/includes/class-wc-calypso-bridge-introductory-offers.php
+++ b/includes/class-wc-calypso-bridge-introductory-offers.php
@@ -15,6 +15,8 @@ use Automattic\Jetpack\Connection\Manager;
  */
 class WC_Calypso_Bridge_Introductory_offers {
 
+	CONST TRANSIENT_PREFIX = 'wc-calypso-bridge-introductory-plans-';
+
 	/**
 	 * return introductory offers for the current blog.
 	 *
@@ -53,6 +55,11 @@ class WC_Calypso_Bridge_Introductory_offers {
 			return [];
 		}
 
+		$cached_offers = get_transient( static::TRANSIENT_PREFIX . $blog_id );
+		if ( $cached_offers ) {
+			return $cached_offers;
+		}
+
 		$response = Client::wpcom_json_api_request_as_blog(
 			'/sites/'.$blog_id.'/introductory-offers',
 			'1.3',
@@ -61,13 +68,17 @@ class WC_Calypso_Bridge_Introductory_offers {
 			'rest'
 		);
 
+		$offers = [];
+
 		if ( ! is_wp_error( $response ) && isset( $response[ 'http_response' ] ) && $response[ 'http_response' ] instanceof WP_HTTP_Requests_Response && 200 === $response[ 'http_response' ]->get_status() ) {
-			$offers = json_decode( $response['http_response']->get_data(), true );
-			if ( is_array( $offers ) && count( $offers ) ) {
-				return $offers;
+			$data = json_decode( $response['http_response']->get_data(), true );
+			if ( is_array( $data ) && count( $data ) ) {
+				 $offers = $data;
 			}
 		}
 
-		return [];
+		set_transient( static::TRANSIENT_PREFIX . $blog_id, $offers, 120 );
+
+		return $offers;
 	}
 }
--- a/includes/class-wc-calypso-bridge-introductory-offers.php
+++ b/includes/class-wc-calypso-bridge-introductory-offers.php
@@ -1,0 +1,73 @@
+<?php
+/**
+ * Contains the logic for querying plans with introductory offers.
+ *
+ * @package WC_Calypso_Bridge/Classes
+ * @since   1.0.6
+ * @version 2.0.8
+ */
+
+use Automattic\Jetpack\Connection\Client;
+use Automattic\Jetpack\Connection\Manager;
+
+/**
+ * WC_Calypso_Bridge_Introductory_offers
+ */
+class WC_Calypso_Bridge_Introductory_offers {
+
+	/**
+	 * return introductory offers for the current blog.
+	 *
+	 * @return array|mixed
+	 */
+	public static function get_current_offers_from_current_blog() {
+		return static::get_introductory_offers( Jetpack_Options::get_option( 'id' ) );
+	}
+
+	/**
+	 * Format offer data for JS.
+	 *
+	 * @param $offer
+	 *
+	 * @return array
+	 */
+	public static function extract_offer_data_for_js( $offer ) {
+		return array(
+			'rawPrice' => $offer['introductory_offer_raw_price'],
+			'formattedPrice' => $offer['introductory_offer_formatted_price'],
+			'intervalUnit' => $offer['introductory_offer_interval_unit'],
+			'intervalCount' => $offer['introductory_offer_interval_count'],
+			'formattedIntervalUnit' => $offer['introductory_offer_interval_count'] === 1 ? $offer['introductory_offer_interval_unit'] : $offer['introductory_offer_interval_unit'].'s',
+		);
+	}
+
+	/**
+	 * Return introductory offers by blog I.D
+	 *
+	 * @param $blog_id
+	 *
+	 * @return array|mixed
+	 */
+	public static function get_introductory_offers( $blog_id ) {
+		if ( ! ( new Manager() )->is_user_connected() ) {
+			return [];
+		}
+
+		$response = Client::wpcom_json_api_request_as_blog(
+			'/sites/'.$blog_id.'/introductory-offers',
+			'1.3',
+			array(),
+			null,
+			'rest'
+		);
+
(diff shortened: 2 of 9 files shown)


class-wc-calypso-bridge-shared.php
```diff
@@ -123,6 +124,13 @@ public function add_scripts() {
 			'version'                      => WC_CALYPSO_BRIDGE_CURRENT_VERSION,
 		);
 
+		if ( $is_ecom_trial ) {
+			$introductory_offers = WC_Calypso_Bridge_Introductory_offers::get_current_offers_from_current_blog();
+			if ( count( $introductory_offers ) ) {
+				$params['introductoryOffer'] = WC_Calypso_Bridge_Introductory_offers::extract_offer_data_for_js( current( $introductory_offers ) );
```
moon0326, line 130: I wasn't sure if we have a logic to select an offer when there are more than 1. For now, I'm choosing the first one with current(). @ilyasfoo

includes/class-wc-calypso-bridge-introductory-offers.php
```diff
@@ -0,0 +1,84 @@
+<?php
+/**
+ * Contains the logic for querying plans with introductory offers.
+ *
+ * @package WC_Calypso_Bridge/Classes
+ * @since   1.0.6
+ * @version 2.0.8
```
ilyasfoo, line 7:
```suggestion
 * @since   x.x.x
 * @version x.x.x
```

includes/class-wc-calypso-bridge-introductory-offers.php
```diff
@@ -0,0 +1,84 @@
...
+
+	/**
+	 * Return introductory offers by blog I.D
+	 *
+	 * @param $blog_id
+	 *
+	 * @return array|mixed
+	 */
+	public static function get_introductory_offers( $blog_id ) {
+		if ( ! ( new Manager() )->is_user_connected() ) {
```
ilyasfoo, line 54: I think in addition to being connected, we need to check if the current connection is owned by the user since:

1. We're using the blog connection's token to run the request.
2. There can be multiple Jetpack users as admin.

https://github.com/Automattic/jetpack/blob/30a0b584e44affc446ca73678fffc4ed3ba531a6/projects/packages/connection/src/class-manager.php#L727

includes/class-wc-calypso-bridge-introductory-offers.php
```diff
@@ -0,0 +1,84 @@
...
+		$offers = [];
+
+		if ( ! is_wp_error( $response ) && isset( $response[ 'http_response' ] ) && $response[ 'http_response' ] instanceof WP_HTTP_Requests_Response && 200 === $response[ 'http_response' ]->get_status() ) {
+			$data = json_decode( $response['http_response']->get_data(), true );
+			if ( is_array( $data ) && count( $data ) ) {
+				 $offers = $data;
+			}
+		}
+
+		set_transient( static::TRANSIENT_PREFIX . $blog_id, $offers, 120 );
```
ilyasfoo, line 80: Good thinking to set transient for performance!

There 2 things that may impact our cache:

1. Current blog's plan change
2. Currency change (since we display the price in UI)

**Plan**: What if we also append the current plan slug to the transient name? Though it's probably not easy to determine the exact current plan slug ([example](https://github.com/Automattic/wc-calypso-bridge/blob/78bf2dcc2cb3f625428cf8315ae24e1dfeeb2472/class-wc-calypso-bridge-dotcom-features.php#L298) where we detect them via purchased features). Perhaps we can serialize all purchases and make that as a key.

**Currency**: I'm not aware of any way to get the latest user currency, but it should be somewhere.

But I digress, I don't think this is a blocker since if there's any impact, it'll probably be edge case.

class-wc-calypso-bridge-shared.php
ilyasfoo: Hmm, I think it's better to just populate the whole array into the parameter, then our frontend can decide how to manage it. 

Might be better since we're preserving `introductoryOffer`'s type (array instead of a single object) in case of a future change. What do you think?

includes/class-wc-calypso-bridge-introductory-offers.php
```diff
@@ -0,0 +1,84 @@
...
+
+/**
+ * WC_Calypso_Bridge_Introductory_offers
+ */
+class WC_Calypso_Bridge_Introductory_offers {
+
+	CONST TRANSIENT_PREFIX = 'wc-calypso-bridge-introductory-plans-';
+
+	/**
+	 * return introductory offers for the current blog.
```
ilyasfoo, line 21: Nit
```suggestion
	 * Return introductory offers for the current blog.
```

src/index.js
```diff
@@ -148,10 +149,17 @@ if ( !! window.wcCalypsoBridge.isEcommercePlanTrial ) {
 		render: WoocommercePaymentsTaskPage,
 	} );
 
-	registerPlugin( 'wc-calypso-bridge-homescreen-slotfill-banner', {
-		render: CalypsoBridgeHomescreenBanner,
-		scope: 'woocommerce-admin',
-	} );
+	if ( window.wcCalypsoBridge.introductoryOffer ) {
```
ilyasfoo, line 152: Since this particular component is specific for ecom free trial plan, I think it's good to add ecom trial check here as well.

src/index.js
```diff
@@ -148,10 +149,17 @@ if ( !! window.wcCalypsoBridge.isEcommercePlanTrial ) {
...
+		registerPlugin( 'wc-calypso-bridge-homescreen-slotfill-banner', {
+			render: CalypsoBridgeIntroductoryOfferBanner,
```
ilyasfoo, line 154: Since this component is not loaded for other than trial plans, let's make this a lazy import

src/introductory-offer-banner/index.js
```diff
@@ -0,0 +1,91 @@
...
+			[ WC_CALYPSO_BRIDGE_INTRODUCTORY_OFFER_BANNER_HIDDEN ]: 'yes',
+		} );
+	};
+
+	return (
+		! isModalHidden && (
+			<Fill name="woocommerce_homescreen_experimental_header_banner_item">
+				<Card>
+					<div className="wc-calypso-bridge-woocommerce-admin-introductory-offer-banner">
+						<div className="wc-calypso-bridge-woocommerce-admin-introductory-offer-banner__text">
```
ilyasfoo, line 45: Font weight should be `600` here

src/introductory-offer-banner/index.js
```diff
@@ -0,0 +1,91 @@
...
+								{ __(
+									'Start selling — for less!',
+									'wc-calypso-bridge'
+								) }
+							</p>
+							<p>
+								{ sprintf(
+									/* translators: First two %s do not need to be translated. Possible values for the last %s is day, days, month, and months */
+									__(
+										'Upgrade your plan for %s for your first %s %s',
```
ilyasfoo, line 56: Sorry to be a bearer of bad news, but I think the handling for the copy here is not exactly straightforward 😬 

[Here's how it's done in calypso](https://github.com/Automattic/wp-calypso/blob/5b6338f8ec5fea0d9bf70fb317f0db9ab1a221ff/client/my-sites/plans-grid/components/billing-timeframe.tsx#L96-L197), tldr is that we may need to handle different intervals unit & count and as separate translations. Hopefully we could copy and adapt some of the calypso code to perform the logic here.

src/introductory-offer-banner/index.js
```diff
@@ -0,0 +1,91 @@
...
+					WC_CALYPSO_BRIDGE_INTRODUCTORY_OFFER_BANNER_HIDDEN
+				) === 'yes' ||
+				! hasFinishedResolution( 'getOption', [
+					WC_CALYPSO_BRIDGE_INTRODUCTORY_OFFER_BANNER_HIDDEN,
+				] ),
+		};
+	} );
+
+	const dismissModal = () => {
+		updateOptions( {
```
ilyasfoo, line 35: I just thought we may want to fire a track for dismiss.

Lets add `recordEvent( 'free_trial_dismiss_homescreen_intro_offer_banner' )`. What do you think?

src/introductory-offer-banner/index.js
```diff
@@ -0,0 +1,91 @@
...
+							<a
+								href={ `https://wordpress.com/plans/${ window.wcCalypsoBridge.siteSlug }` }
+								className="wc-calypso-bridge-woocommerce-admin-introductory-offer-banner__upgrade-button components-button is-secondary"
+							>
+								{ __( 'Upgrade now', 'wc-calypso-bridge' ) }
+							</a>
+							<Button
+								className="wc-calypso-bridge-woocommerce-admin-introductory-offer-banner__dismiss-button"
+								label={ __(
+									'Dismiss this introductory offer informational banner.'
```
ilyasfoo, line 78: Good thinking to add label! Let's make it shorter:
```suggestion
									'Dismiss this banner.'
```

ilyasfoo (changes requested):
Hey, @moon0326! Thanks for working on this. I have quite a few feedback 😬 

Also, please bump the version tags in the files: 
- `class-wc-calypso-bridge.php`
- `class-wc-calypso-bridge-shared.php`

And update `readme.txt` for changelog

class-wc-calypso-bridge-shared.php
moon0326: That's a good for future proofing, but we also need to consider the data size. We're loading it for every page load. Since this is pretty easy to change, I think we should expose the whole array when we actually have a use case. What do you think?

includes/class-wc-calypso-bridge-introductory-offers.php
moon0326: That's a good approach, but that means we need to call wpcom_get_site_purchases() on every check unless I'm mistaken. I figured 2 mins is short enough for the page to be refreshed when the user decided to upgrade.

1. User sees Woo Home with cached data. 2 mins cache starts.
2. User clicks `Upgrade` and goes through the flow.
3. User comes back to Woo Home. Cache should have been expired at this point.

src/index.js
moon0326: I believe the check is already done here https://github.com/Automattic/wc-calypso-bridge/blob/78bf2dcc2cb3f625428cf8315ae24e1dfeeb2472/src/index.js#L102

class-wc-calypso-bridge-shared.php
moon0326: Do you happen to have a link to the source code that explains how Calypso selects which plan to display when there are multiple offers?

ilyasfoo: Good point, I agree that producing all offers is premature 👍 

> Do you happen to have a link to the source code that explains how Calypso selects which plan to display when there are multiple offers (Does it even have a logic for selecting only one 🤔 )?

IMO calypso don't have the same problem since the app is JS and no server is involved, so everything is already available regardless if they only need 1 plan or more.

moon0326: @ilyasfoo I'm not clear on whether we can have multiple offers with different terms while using only one banner. Does this mean we have to choose just one offer, perhaps the first one? 🤔

includes/class-wc-calypso-bridge-introductory-offers.php
ilyasfoo: > That's a good approach, but it implies that we need to call wpcom_get_site_purchases() with every check, unless I'm mistaken

I think `wpcom_get_site_purchases` is already executed in every request anyway since there are fundamental logic built around it.

Still, good point on 2 minutes buffer for upgrades, I think it's a reasonable. 👍

class-wc-calypso-bridge-shared.php
ilyasfoo: @moon0326 Gotcha. Since the introductory offer is the same for all Woo Express plans, I think this is how we should handle it:

1. Perhaps instead of `introductoryOffer`, let's rename it to `wooExpressIntroductoryOffer`
1. When deciding which plan to take, loop and check if it's any of [Woo Express plan slugs](https://github.com/Automattic/wp-calypso/blob/63c08c6989595742db1f9aaf70216663198db29b/packages/calypso-products/src/constants/wpcom.ts#L39-L42) and return the first one found

src/introductory-offer-banner/style.scss
```diff
@@ -0,0 +1,56 @@
+.wc-calypso-bridge-woocommerce-admin-introductory-offer-banner {
+    display: flex;
+    flex-direction: row;
+    align-items: center;
+    min-height: 65px;
+	background: var(--wp-admin-theme-color);
+	border-radius: 3px;
+	border-color: none;
```
ilyasfoo, line 8: Invalid css
```suggestion
	border-color: transparent;
```

class-wc-calypso-bridge-shared.php
moon0326: Updated in https://github.com/Automattic/wc-calypso-bridge/pull/1397/commits/f2f26ed8676527a782671bf5352fe97ef4ecce68

includes/class-wc-calypso-bridge-introductory-offers.php
moon0326: Updated in https://github.com/Automattic/wc-calypso-bridge/pull/1397/commits/d312cb3c21d5339fe8af4d89d4bb864e5085d71c

includes/class-wc-calypso-bridge-introductory-offers.php
moon0326: Sounds good. Let's skip it for now 👍

src/introductory-offer-banner/index.js
moon0326: Updated in [Update font weight](https://github.com/Automattic/wc-calypso-bridge/pull/1397/commits/fdc305fb16417d0c030ae5de9cddb69dcafe38ab)

src/introductory-offer-banner/index.js
moon0326: Updated in https://github.com/Automattic/wc-calypso-bridge/pull/1397/commits/6ec02a5c7557857dd1aab523d31cb02bf3091f8f

src/index.js
moon0326: Correct me if I'm wrong, but it seems like RegisterPlugin doesn't currently support lazy-loaded React components. I've tried it, but nothing gets loaded. It's probably not using the Suspense component.

Do you happen to know if we have an example of a lazy-loaded plugin?

ilyasfoo: Good question, I'm not sure if this works differently, but [PaymentGatewaySuggestions](https://github.com/Automattic/wc-calypso-bridge/blob/0233630e1a2a406600cfd6a3c1d0be81d3637a2c/src/index.js#L123) is lazy loaded and it's under a `registerPlugin` call but it's wrapped inside another component.

Perhaps we can try to wrap the lazy loaded component inside an empty component? If it doesn't work, then I think we can just skip this suggestion.

src/introductory-offer-banner/index.js
ilyasfoo: Thanks!

src/introductory-offer-banner/index.js
ilyasfoo: @moon0326  Good point, but I think we still need to handle this slightly more robust. I think we still need to still translate the period, such as `day`, `month`, `year`, along with the plural variations.

ChatGPT provided this logic 😆 :

```js
let message = '';
if (intervalCount === '1') {
    switch (formattedIntervalUnit) {
        case 'day':
            message = sprintf(__('Upgrade your plan for %s for your first day', 'wc-calypso-bridge'), formattedPrice);
            break;
        case 'month':
            message = sprintf(__('Upgrade your plan for %s for your first month', 'wc-calypso-bridge'), formattedPrice);
            break;
        case 'year':
            message = sprintf(__('Upgrade your plan for %s for your first year', 'wc-calypso-bridge'), formattedPrice);
            break;
    }
} else {
    switch (formattedIntervalUnit) {
        case 'days':
            message = sprintf(__('Upgrade your plan for %s for your first %s days', 'wc-calypso-bridge'), formattedPrice, intervalCount);
            break;
        case 'months':
            message = sprintf(__('Upgrade your plan for %s for your first %s months', 'wc-calypso-bridge'), formattedPrice, intervalCount);
            break;
        case 'years':
            message = sprintf(__('Upgrade your plan for %s for your first %s years', 'wc-calypso-bridge'), formattedPrice, intervalCount);
            break;
    }
}
```

moon0326: Ah! I see what you mean @ilyasfoo Thank you!

I'll update it soon.

moon0326: Updated in https://github.com/Automattic/wc-calypso-bridge/pull/1397/commits/c36fd20d9d80cf58f377ce58a748300b4c851d9e

@ilyasfoo

src/introductory-offer-banner/index.js
```diff
@@ -0,0 +1,87 @@
...
+						<div className="wc-calypso-bridge-woocommerce-admin-introductory-offer-banner__actions">
+							<a
+								href={ `https://wordpress.com/plans/${ window.wcCalypsoBridge.siteSlug }` }
+								className="wc-calypso-bridge-woocommerce-admin-introductory-offer-banner__upgrade-button components-button is-secondary"
+							>
+								{ __( 'Upgrade now', 'wc-calypso-bridge' ) }
+							</a>
+							<Button
+								className="wc-calypso-bridge-woocommerce-admin-introductory-offer-banner__dismiss-button"
+								label={ __( 'Dismiss this banner.' ) }
```
chihsuan, line 75:
```suggestion
								label={ __( 'Dismiss this banner.', 'wc-calypso-bridge' )  }
```

class-wc-calypso-bridge-shared.php
```diff
@@ -123,6 +125,16 @@ public function add_scripts() {
 			'version'                      => WC_CALYPSO_BRIDGE_CURRENT_VERSION,
 		);
 
+		if ( $is_ecom_trial ) {
+			$woo_express_introductory_offers = WC_Calypso_Bridge_Woo_Express_Introductory_offers::get_offers_for_current_blog();
+			if ( count( $woo_express_introductory_offers ) ) {
+				$woo_express_introductory_offers = array_filter( $woo_express_introductory_offers, function( $offer ) {
+					return in_array( $offer['product_slug'], WC_Calypso_Bridge_Woo_Express_Introductory_offers::WOO_EXPRESS_PRODUCT_SLUGS );
+				} );
+				$params['wooExpressIntroductoryOffer'] = WC_Calypso_Bridge_Woo_Express_Introductory_offers::extract_offer_data_for_js( current( $woo_express_introductory_offers ) );
```
chihsuan, line 134: It seems that in certain cases `$woo_express_introductory_offers` may be an empty array due to the filter. Since `current` will return `false` when array is empty, should we add a check for this to ensure it return an expected array? Or should we move this logic into `get_offers_for_current_blog`?

chihsuan (approved):
Nice work. Confirmed banner is shown and works as expected. Left some major suggestions.

Pre-approving! 👍

class-wc-calypso-bridge-shared.php
moon0326: Good catch @chihsuan 👍 

I've updated `get_offers_for_current_blog` to accept a filter to reduce code in `class-wc-calypso-bridge-shared.php` and avoid using count() again. The current() should always return a value now. 

https://github.com/Automattic/wc-calypso-bridge/pull/1397/commits/017b821417b541c35d7fe672cd2d17351b5f7f90

src/introductory-offer-banner/get-offer-message.ts
```diff
@@ -0,0 +1,44 @@
...
+			{
+				formattedPrice,
+				intervalUnit,
+			}
+		);
+	} else {
+		message = sprintf(
+			/* translators:  %(intervalUnit)s: one of day, month, and year, formattedPrice and intervalCount do not need to be translated. */
+			__(
+				'Upgrade your plan for %(formattedPrice)s for your first %(intervalCount)s %(intervalUnit)ss',
```
ilyasfoo, line 32: AFAIK if we do it like this, `intervalUnit` is not translated. I think it could be like this in Spanish:

`Actualice su plan por $1 por su primer 3 months`

Thus I think we need to have all the possible interval units written in our copy here.

includes/class-wc-calypso-bridge-woo-express-introductory-offers.php
```diff
@@ -0,0 +1,94 @@
...
+	 *
+	 * @return array
+	 */
+	public static function extract_offer_data_for_js( $offer ) {
+		return array(
+			'rawPrice' => $offer['introductory_offer_raw_price'],
+			'formattedPrice' => $offer['introductory_offer_formatted_price'],
+			'intervalUnit' => $offer['introductory_offer_interval_unit'],
+			'intervalCount' => $offer['introductory_offer_interval_count'],
+			'formattedIntervalUnit' => $offer['introductory_offer_interval_count'] === 1 ? $offer['introductory_offer_interval_unit'] : $offer['introductory_offer_interval_unit'].'s',
```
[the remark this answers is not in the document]
ilyasfoo: Also, it seems like we're not using this, maybe we can remove it instead?

ilyasfoo (changes requested):
Left another round of review with minor comments. Also, please add changelog to `readme.txt`. Thanks for working on this, @moon0326!

includes/class-wc-calypso-bridge-woo-express-introductory-offers.php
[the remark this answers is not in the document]
moon0326: Thanks! You're correct that we no longer use this var. Removed in [05cb4c3](https://github.com/Automattic/wc-calypso-bridge/pull/1397/commits/05cb4c3fb46a198f578d2772a5d16b982a1124e9)

src/introductory-offer-banner/get-offer-message.ts
moon0326: @ilyasfoo  Thank you for pointing it out 👍  TIL.

Listed all possible interval units in https://github.com/Automattic/wc-calypso-bridge/pull/1397/commits/8196d18bdaa6c8a33d1d3c29c611520a2fb1680f

ilyasfoo (changes requested):
@moon0326 It seems like we've missed one minor detail in the `sites/blogid/introductory-offers` API, when changing the user's currency, it doesn't seem to support it when called via Jetpack API 😞

I tested it by changing my user's currency in Store Admin, and then run the API call via site, and it shows USD regardless.

I'm not entirely sure how to handle this yet, but likely we may need to modify our API to change currency according to user.

Also 1 more thing left, please changelog in `readme.txt`

ilyasfoo (approved):
> I'm not sure if the plans API supports different currencies as WordPress doesn't have currency settings (WC does though). If it does, it has to reflect the current rate and that's going to be tricky.

AFAIK the currency is WPCOM specific user settings, so it's not related to WP nor WC.

> For intro offers endpoint, we probably need to pass locale as well and use wpcom_switch_to_locale in the backend.

Tbh I'm not sure if locale makes a difference, I just tried it and it didn't work for me. Were you able to get it to work?
 
> I'll look into it when I get back later this evening. What do you think of making currency/local changes in a follow-up PR? I don't think this is a blocker.

I've tested the introductory-offers API, when called via https://developer.wordpress.com/docs/api/console/, it does use WPCOM currency automatically. I think it's because since we use Jetpack's site token to authenticate, WPCOM doesn't automatically use user's currency:

image

I agree, we can make a follow-up PR if necessary

ilyasfoo:
@moon0326 I've dug deep into the rabbit hole and I discovered an existing API to obtain the list of introductory offers! https://github.com/Automattic/wc-calypso-bridge/pull/1397/commits/034b2681562322c38872a9b619c4a9d1cfb782b1

Since this API is using `wpcom/v2`, it seems like it supports proper Jetpack token auth, thus it's able to use user's currency out of the box!

However, since this results in using an actual user in the request, we can't test this with an Automattician user account since it will fail the proxy check and make the user ineligible for the $1 offer.

I've removed the API we've added and removed proxy check to help with testing D133565-code.

moon0326:
> @moon0326 I've dug deep into the rabbit hole and I discovered an existing API to obtain the list of introductory offers! [034b268](https://github.com/Automattic/wc-calypso-bridge/commit/034b2681562322c38872a9b619c4a9d1cfb782b1)
> 
> Since this API is using `wpcom/v2`, it seems like it supports proper Jetpack token auth, thus it's able to use user's currency out of the box!
> 
> However, since this results in using an actual user in the request, we can't test this with an Automattician user account since it will fail the proxy check when the request is made from Jetpack backend and make the user ineligible for the $1 offer.
> 
> I've removed the API we've added and removed proxy check to help with testing [D133565-code](https://href.li/?https://code.a8c.com/D133565).

"Nice! Thank you for taking the time to look into it. I should have searched more deeply

PR #1398, opened 2023-12-19
Track calypso_wooexpress_one_dollar_offer when a user clicks on My Ho…

### LINKED ISSUE #1322: [Woo Express $1 offer] Homescreen track
## Issue description

Related to https://github.com/Automattic/wp-calypso/pull/82099, we need to add the track `calypso_wooexpress_one_dollar_offer` with the prop `location: homescreen` when a user visits homescreen.

We have a prior art for adding calypso tracks: https://github.com/Automattic/wc-calypso-bridge/blob/ef3b3ae7c16caff0aaae0a0b6e1329c351bc7a08/src/track-menu-item.ts#L31-L33

## Acceptance criteria

1. When a user is eligible for $1 offer and visits the homescreen, fire the tracks `calypso_wooexpress_one_dollar_offer` with props `location: homescreen`.

moon0326:
…me and has an active intro offer

Closes #1322  .

1. Sign up for a new trial site.
2. Make your trial site SSH-able by following peapX7-1D4-p2
3. Clone this branch and run `npm run start` to build the changes.
4. SSH into your trial site and copy over `/build` directory to `~/htdocs/wp-content/mu-plugins/wpcomsh/vendor/automattic/wc-calypso-bridge/build`
5. Access the site and open browser inspector then paste `window.localStorage.setItem( 'debug', 'calypso:analytics' )`
6. Change inspector log level to verbose
7. Go to WooCommerce Home.
8. Confirm `calypso_wooexpress_one_dollar_offer` track has been fired.

Note: This won't work if a user clicks on `My Home` from a page where the `wcCalypsoBridge` JS variable is not available, such as on the My Mailboxes or Upgrades pages.

### Other information:

### FOR PR REVIEWER ONLY:


ilyasfoo (changes requested):
Thanks for working on this, @moon0326!

Sorry this wasn't clear in the issue, we need the track to be fired when users `views` the homescreen. This is because we need to use the track to be fired as soon as the user comes from free trial setup.

moon0326:
> Thanks for working on this, @moon0326!
> 
> Sorry this wasn't clear in the issue, we need the track to be fired when users `views` the homescreen. This is because we need to use the track to be fired as soon as the user comes from free trial setup.

Thanks for the clarification. We probably need a different approach in that case. I'll make an update soon.

moon0326:
> Thanks for working on this, @moon0326!
> 
> Sorry this wasn't clear in the issue, we need the track to be fired when users `views` the homescreen. This is because we need to use the track to be fired as soon as the user comes from free trial setup.

I've decided to call the track in the intro offer banner as it is only displayed when the offer is available. 

Updated in https://github.com/Automattic/wc-calypso-bridge/pull/1398/commits/35998fa912e3a940377bd2f2b9cdb8290998b5f8

I've also updated test introduction.

ilyasfoo (changes requested):
> Thanks for the clarification. We may need to take a different approach in that case. I'll make an update soon.
>
> Do we want to fire the track only once?

Good question, I think we should track it every time, since that's the behaviour for $1 offer track for plans and checkout screen.

> I've decided to call the track in the intro offer banner as it is only displayed when the offer is available.
>
> Updated in https://github.com/Automattic/wc-calypso-bridge/commit/35998fa912e3a940377bd2f2b9cdb8290998b5f8

I think we should make the behavior consistent with other $1 tracks, so we should track the view, regardless of banner status 🙏 

Also, please update changelog, thanks! 😊

psealock:
>Change inspector log level to verbose

Can you remind me how to do this again?

moon0326:
> > Change inspector log level to verbose
> 
> Can you remind me how to do this again?

Sure thing! 

Screen Shot 2024-01-08 at 4 03 08 PM

ilyasfoo (changes requested):
Thanks for making the changes, @moon0326!

It seems like the SlotFill is being rendered in all wc-admin pages 😓
 
image

I think we'll also need to check the admin path to make sure it's homescreen

moon0326:
> Thanks for making the changes, @moon0326!
> 
> It seems like the SlotFill is being rendered in all wc-admin pages 😓
> 
> image
> I think we'll also need to check the admin path to make sure it's homescreen

Thank you for catching it! That's unexpected for sure. I thought we only render the banner on homescreen :( I'll make a fix and look into why the component is being rendered on other pages later.

src/index.js
```diff
@@ -146,16 +146,18 @@ if ( !! window.wcCalypsoBridge.isEcommercePlanTrial ) {
...
-		registerPlugin( 'wc-calypso-bridge-homescreen-slotfill-banner', {
-			render: CalypsoBridgeIntroductoryOfferBanner,
-			scope: 'woocommerce-admin',
-		} );
-	} else {
-		registerPlugin( 'wc-calypso-bridge-homescreen-slotfill-banner', {
-			render: CalypsoBridgeHomescreenBanner,
-			scope: 'woocommerce-admin',
-		} );
+	if ( window.location.search === '?page=wc-admin' ) {
```
ilyasfoo, line 149: @moon0326 I think this is not reliable since there might be other URL parameters used.

I'd suggest to use `getPath` from `@woocommerce/navigation` where it's more reliable

ilyasfoo (changes requested):
> Thank you for catching it! That's unexpected for sure. I thought we only render the banner on homescreen :( I'll make a fix and look into why the component is being rendered on other pages later.

@moon0326 Thanks! I have a minor comment on the URL parameter detection

src/index.js
moon0326: @ilyasfoo getPath() returns `location.pathname` and it's hard to determine Woo Home with the returned value AFAIK. Do you see situations where this code won't work?

moon0326: FYI, I've also tried using [determine_current_page](https://github.com/woocommerce/woocommerce/blob/trunk/plugins/woocommerce/src/Admin/PageController.php#L119) and [get_current_page](https://github.com/woocommerce/woocommerce/blob/trunk/plugins/woocommerce/src/Admin/PageController.php#L219) from PageController class. Both report `homescreen` when you're on a task page.

ilyasfoo: Thanks, @moon0326! Unfortunately, I found another problem with the way we conditionally load the SlotFill 😅 

It seems if we navigate to another page first, such as `/wp-admin/admin.php?page=wc-admin&path=%2Fanalytics%2Foverview` and then navigate to homescreen using the menu bar, no banner will be loaded.

Sorry for going round and round on this, but let's not check for homescreen in the `registerPlugin` layer, but instead in the component before firing the tracks.

moon0326: > Sorry for going round and round on this, but let's not check for homescreen in the registerPlugin layer, but instead in the component before firing the tracks.

No worries. Thank you for catching these cases 🤦 

I'll run a thorough test again.

moon0326: @ilyasfoo Fixed in https://github.com/Automattic/wc-calypso-bridge/pull/1398/commits/6bb67a11658027fa0989e3579f604333be43f61a 🙏

ilyasfoo (approved):
Tested well, LGTM! 🚢

PR #1403, opened 2023-12-21
Hide Payments menu item for unsupported countries

### LINKED ISSUE #1345: Hide Payments menu item if an unsupported location/country is selected
### Description

As highlighted by Happiness at least a couple times now, pre-activating WooPayments for all new Free Trial users results in a confusing experience for merchants based in countries unsupported by WooPayments, as:

- We are limiting/restricting the number of gateways to WooPayments, Stripe, PayPal.
- Merchants who have indicated that they are based in a country unsupported by WooPayments will not be onboarded to it via the Task List -- and in some cases, Stripe and PayPal will also be unavailable.
- However, the Payments menu item is always visible and merchants may attempt to go through the WooPayments KYC flow, even if it leads to a dead end.

A short term fix here, which is the objective of this task, is to conditionally hide the Payments menu item using the same Core logic that prompts users to install/configure WooPayments (Set up WooPayments) vs other gateways (Set up payments):

```
$show_menu_item     = true;
$suggestions              = Suggestions::get_suggestions();
$suggestion_plugins = array_merge(
    ...array_filter(
        array_column( $suggestions, 'plugins' ),
        function( $plugins ) {
            return is_array( $plugins );
        }
    )
);
$woocommerce_payments_ids = array_search( 'woocommerce-payments', $suggestion_plugins, true );
if ( false === $woocommerce_payments_ids ) {
    $show_menu_item = false;
}
```

Later on, we could review/revisit the experience of merchants who arrive to the Free Trial having chosen a country with no supported payment gateway options, to indicate that some options will be available to them once they upgrade to a paid plan:

Image

Remember that, the need to have payment restrictions in place during the Free Trial forces us to limit the payment gateways surfaced during onboarding, as payment restrictions cannot be implemented in a gateway-agnostic manner.

jimjasson:

This PR conditionally hides the Payments menu item for countries that do not support Woo Payments.

This is when Woo Payments adds the Payments menu item: https://github.com/Automattic/woocommerce-payments/blob/develop/includes/admin/class-wc-payments-admin.php#L205

Closes #1345 

1. Create a new Woo Express dev site (plan is not relevant).
2. Use a country that supports Woo Payments as the store's base (under Settings > WooCommerce > General) -- for example, US.
3. Ensure that the Payments menu item is visible. 
4. Use a country that doesn't support Woo Payments as the store's base (under Settings > WooCommerce > General) -- for example, South Africa.
5.  Ensure that the Payments menu item is visible. 
6. Apply the patch. 
7. The Payments menu item should be hidden. 
8. Use a country that supports Woo Payments as the store's base (under Settings > WooCommerce > General) -- for example, US.
9.  The Payments menu item should be visible. 

### Other information:

### FOR PR REVIEWER ONLY:

-   [x] I have reviewed that everything is sanitized/escaped appropriately for any SQL or XSS injection possibilities. I made sure Linting is not ignored or disabled.


xristos3490 (changes requested):
Looks good! 

Your idea of fetching the support flag from the onboarding tasks looks great! However, to ensure safety, I suggest using the `$this->hide_submenu_page()` method instead of completely removing menus. This way, the rewrite rules can still exist and potential issues or conflicts can be avoided

jimjasson:
`$this->hide_submenu_page()` is for hiding submenus and Payments is the primary tab, so we decided to keep this PR as is.

PanosSynetos (approved):
Manually tested, works as expected

beaulebens:
Shouldn't this be handled in WooPayments? e.g. it should only put it there in the first place if it's available?

PanosSynetos:
Hey @beaulebens - I don't know the reasoning behind this, but you do have a point. Woo Payments should be handling this. I'll ping someone from Woo Payments

PR #1472, opened 2024-05-17
i18n: Load script translations

yuliyan:

* Load translations for `wc-calypso-bridge` script.
* Build JSON translations files with `wp i18n make-json [path-to-po-file]`.
* Exclude the `/src` path when generating the POT file to reduce the number of JSON translation files generated with `make-json`, as it uses the files paths references in the POT file to create the JSON files. A JSON translation files is loaded when its corresponding JS is enqueued in WordPress and given that the source modules never enqueued directly, generating translation files for those is unnecessary.
* Add filters for `gettext` calls with `woocommerce` to attempt to translate the same string using `wc-calypso-bridge` when there's no translation in the `woocommerce` domain. This change is targeted at gettext call used in this package with `woocommerce` text domain (such as [#1](https://github.com/Automattic/wc-calypso-bridge/blob/8e73dbad01396a364a3014a91ab25154a03793da/src/task-headers/products.js#L26), [#2](https://github.com/Automattic/wc-calypso-bridge/blob/8e73dbad01396a364a3014a91ab25154a03793da/src/free-trial/tax/avalara/card.tsx#L27), [#3](https://github.com/Automattic/wc-calypso-bridge/blob/8e73dbad01396a364a3014a91ab25154a03793da/src/payment-gateway-suggestions/index.js#L215), etc.).
* Load translation data for lazy loaded modules.

Related to 799-gh-Automattic/i18n-issues

1. Checkout the branch locally.
2. Generate JSON translation files for one of the languages with `wp i18n make-json languages/[language].po`.
3. Sync the changes to your test site by following the instructions from pdDOJh-3ob-p2.
4. Open your test site and make sure set to the language you've generated the translation files for.
5. Navigate to `/wp-admin/admin.php?page=wc-admin` and confirm `wc-calypso-bridge-js-translations` translation data is loaded, i.e. the script `<script id="wc-calypso-bridge-js-translations">` that loads the data using `wp.i18n.setLocaleData()` is printed on the page.
6. Navigate to `/wp-admin/admin.php?page=wc-admin&path=%2Fmarketing` and confirm translations for lazy loaded module `Marketing` are loaded as expected. The `isEcommercePlan`, `isEcommercePlanTrial`, `isWooNavigationEnabled` params on your test site must be `true` (can be overridden in [class-wc-calypso-bridge-shared.php](https://github.com/Automattic/wc-calypso-bridge/blob/45e22acc9033d37a9e95e30603e0a06cc6fdefab/class-wc-calypso-bridge-shared.php#L114-L117))

### Other information:

### FOR PR REVIEWER ONLY:

### DIFF
--- a/src/i18n-loader.js
+++ b/src/i18n-loader.js
@@ -1,0 +1,44 @@
+/**
+ * External dependencies
+ */
+import md5 from 'crypto-js/md5';
+import { setLocaleData } from '@wordpress/i18n';
+
+const defaultLocale = 'en_US';
+
+/**
+ * Load lazy loaded modules translation data.
+ *
+ * @param {string} path
+ * @param {string} domain
+ * @return {Promise}
+ */
+export async function loadTranslations( path, domain ) {
+	const locale = window?.wcCalypsoBridge?.i18n?.locale ?? defaultLocale;
+	const baseUrl = window?.wcCalypsoBridge?.i18n?.baseUrl;
+
+	if ( locale === defaultLocale || baseUrl === undefined ) {
+		return;
+	}
+
+	const [ relativePath, version ] = path.split( '?' );
+
+	const hash = md5( relativePath ).toString();
+	const filename =
+		`${ domain }-${ locale }-${ hash }.json` +
+		( version ? `?${ version }` : '' );
+
+	const response = await fetch( `${ baseUrl }/${ filename }` );
+
+	if ( ! response.ok ) {
+		throw new Error(
+			`HTTP request failed: ${ response.status } ${ response.statusText }`
+		);
+	}
+
+	const data = await response.json();
+	const localeData =
+		data?.locale_data?.[ domain ] || data?.locale_data?.messages;
+
+	setLocaleData( localeData, 'wc-calypso-bridge' );
+}
--- a/src/i18n-loader.js
+++ b/src/i18n-loader.js
@@ -10,7 +10,7 @@ import {
 	_nx,
 	_x,
 } from '@wordpress/i18n';
-import { addFilter, applyFilters } from '@wordpress/hooks';
+import { addFilter } from '@wordpress/hooks';
 
 const defaultLocale = 'en_US';
 
--- a/src/i18n-loader.js
+++ b/src/i18n-loader.js
@@ -17,7 +17,7 @@ const defaultLocale = 'en_US';
 /**
  * Some gettext calls in this package use the `woocommerce` text domain
  * and expect that the translation data is already loaded from WooCommerce.
- * If a translation is missing, these filters will attempt to transate the string
+ * If a translation is missing, these filters will attempt to translate the string
  * using the `wc-calypso-bridge` text domain as a fallback.
  */
 addFilter(
--- a/src/i18n-loader.js
+++ b/src/i18n-loader.js
@@ -2,9 +2,84 @@
  * External dependencies
  */
 import md5 from 'crypto-js/md5';
-import { setLocaleData } from '@wordpress/i18n';
+import {
+	setLocaleData,
+	hasTranslation,
+	__,
+	_n,
+	_nx,
+	_x,
+} from '@wordpress/i18n';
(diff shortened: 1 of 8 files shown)


ilyasfoo (commented):
Thanks for working on this, @yuliyan. However, I'm having a hard time to test this. It seems like the translations aren't working when I change to `es_MX`. I confirmed to have run the command `wp i18n make-json languages/wc-calypso-bridge-es_MX.po`, and I changed my language to `Español de México`.



image

It did generate JSON files:

image

yuliyan:
Thank your for reviewing the changes, @ilyasfoo!

I'm sorry I didn't make the instructions more clear, but rendering the translated strings is a separate problem [caused by using incorrect text domain](https://github.com/Automattic/wc-calypso-bridge/blob/master/src/homescreen-progress-header/progress-header.tsx#L59) (i.e. `woocommerce` instead of `wc-calypso-bridge`).

The goal of this PR is to load the translation data into the i18n instance, i.e. to make sure `wc-calypso-bridge-js-translations` script is printed on the page.

I plan to follow up with creating a separate PR to update the text domain and another PR to commit translations files (i.e. run `npm run prepare`).

yuliyan:
@ilyasfoo,

I've previously [mentioned that some gettext calls were using incorrect text ](https://github.com/Automattic/wc-calypso-bridge/pull/1472#issuecomment-2120091605), but after further looking into it, I've noticed that some of the strings also exist in WooCommerce and updating the text domain would it from re-using the WooCommerce translations.

Instead, I took a different approach and add fallback translation lookup for gettext calls with `woocommerce` text domain that are missing tranaslations - https://github.com/Automattic/wc-calypso-bridge/pull/1472/commits/f1bceadbd63ad7006885325029e469a92f8c02bd. If you repeat your testing steps, you should now see the translated items.

I've also noticed that there were some lazy loaded modules that have their own translation files, that were previously not handled, so I've updated the PR to also handle this case in https://github.com/Automattic/wc-calypso-bridge/pull/1472/commits/9f81bc2882b65eaecebf8922bb6c1153c53e73ec.

src/i18n-loader.js
```diff
@@ -0,0 +1,119 @@
...
+} from '@wordpress/i18n';
+import { addFilter } from '@wordpress/hooks';
+
+const defaultLocale = 'en_US';
+
+/**
+ * Some gettext calls in this package use the `woocommerce` text domain
+ * and expect that the translation data is already loaded from WooCommerce.
+ * If a translation is missing, these filters will attempt to transate the string
+ * using the `wc-calypso-bridge` text domain as a fallback.
```
d-alleyne, line 21: This is an interesting workaround. I was unaware of these JS hooks.

d-alleyne (approved):
Works as described.
 `<script id="wc-calypso-bridge-js-translations">` is in the source.
I've learned that many of the untranslated strings were due to [removals in a previous commit](https://github.com/Automattic/wc-calypso-bridge/commit/4709ce6d5b2d43bf29d922c32a98c6d134f0c053#diff-5ddaeba5e468923b08c2ff94baadb67d44359ed8c2997eddc2f02b7a4b4f30f1L1121), and will be fixed in a [future commit](https://github.com/Automattic/wc-calypso-bridge/pull/1477).

src/i18n-loader.js
```diff
@@ -0,0 +1,119 @@
...
+	_x,
+} from '@wordpress/i18n';
+import { addFilter } from '@wordpress/hooks';
+
+const defaultLocale = 'en_US';
+
+/**
+ * Some gettext calls in this package use the `woocommerce` text domain
+ * and expect that the translation data is already loaded from WooCommerce.
+ * If a translation is missing, these filters will attempt to transate the string
```
ilyasfoo, line 20: Typo
```suggestion
 * If a translation is missing, these filters will attempt to translate the string
```

ilyasfoo (approved):
Awesome, thank you for the comprehensive fix! It tested well.

PR #1483, opened 2024-05-31
Fix WC_Tracks called too early crashing frontpage

ilyasfoo:

When using WooCommerce 9.0.0 and viewing homepage as an admin, PHP crashes.

p1717075183129709/1717053288.243629-slack-C011ENB20Q1

1. Set up with Entrepreneuer plan (by following this guide pdDOJh-3ob-p2, or the old free trial guide and upgrade it to Entrepreneur plan peapX7-1D4-p2)
1. Go to cli
4. Run `wp wpcomsh plugin use-unmanaged woocommerce`
5. Download and install [WooCommerce 9.0.0 beta 2](https://github.com/woocommerce/woocommerce/releases/tag/9.0.0-beta.2)
6. In your logged in browser, go to homepage in the frontend (not wp-admin)
7. Observe the homepage is loaded correctly (no white screen)

### Other information:

### FOR PR REVIEWER ONLY:

### DIFF
--- a/includes/class-wc-calypso-bridge-customize-store.php
+++ b/includes/class-wc-calypso-bridge-customize-store.php
@@ -103,7 +103,7 @@ class WC_Calypso_Bridge_Customize_Store {
 
 	public function possibly_add_track_homepage_view() {
 		if ( self::is_admin() ) {
-			if ( is_front_page() || is_home() ) {
+			if ( class_exists( 'WC_Tracks' ) && ( is_front_page() || is_home() ) ) {
 				// This is tracked via backend.
 				WC_Tracks::record_event( 'store_homepage_view' );
 			}
--- a/includes/class-wc-calypso-bridge-customize-store.php
+++ b/includes/class-wc-calypso-bridge-customize-store.php
@@ -4,7 +4,7 @@
  *
  * @package WC_Calypso_Bridge/Classes
  * @since  1.0.0
- * @version 2.3.0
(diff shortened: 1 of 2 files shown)


moon0326 (approved):
LGTM and tested well 👍 🚀  Pre-approving with one minor suggestion. 

I'm a bit hesitant to try it because we're attempting to include files from a different plugin directly, but we can include the track file if it does not exist. What do you think?

Something like this should work (not tested)
```
public function load_wc_tracker() {
	if ( class_exists( 'WC_tracks') || !defined( 'WC_ABSPATH' ) ) {
		return true;
	}

	include WC_ABSPATH . '/includes/class-wc-tracker.php';
	include WC_ABSPATH . '/includes/tracks/class-wc-tracks.php';
}
```

rjchow (approved):
LGTM Thanks!

ilyasfoo:
> I'm a bit hesitant to try it because we're attempting to include files from a different plugin directly, but we can include the track file if it does not exist. What do you think?

> Something like this should work (not tested)

Thanks for the suggestion, @moon0326! I think it's kind of overkill at this point, since we don't intend to actually include tracks that early in execution. The root cause of the bug in WooCommerce 9.0 beta was [addressed here](https://github.com/woocommerce/woocommerce/pull/47987), I intend this fix to only ensure that future potential bugs will not cause a crash.

PR #1426, opened 2024-01-12
Partner Aware Onboarding - Customizations for Square

moon0326:
### Changes proposed in this Pull Request:

Closes #1422

This PR removes Woo Payments from the following REST API endpoints when `woocommerce_onboarding_profile.partner` is `square` 

- /wp-json/wc-admin/payment-gateway-suggestions
- /wp-json/wc-admin/onboarding/free-extensions

It also adds `Get paid with Square` task.

Screen Shot 2024-01-31 at 1 58 05 PM

### How to test the changes in this Pull Request:

**Feed Removal** 
1. Checkout this branch.
2. Make sure you're in trial plan.
3. Use wp cli to update `woocommerce_onboarding_profile` option value with `a:10:{s:15:"business_choice";s:28:"im_just_starting_my_business";s:21:"selling_online_answer";N;s:17:"selling_platforms";N;s:20:"is_store_country_set";b:1;s:8:"industry";a:1:{i:0;N;}s:18:"is_agree_marketing";b:0;s:11:"store_email";s:17:"test@gmail.com";s:9:"completed";b:1;s:23:"is_plugins_page_skipped";b:1;s:7:"partner";s:6:"square";}` It includes `partner=square`
4. Open your browser's inspector and click `Network` tab.
5. Start the core profiler.
6. On the `Tell us a bit about your store` page, click `Continue` button.
7. Inspect a request to `/wp-json/wc-admin/onboarding/free-extensions`
8. Confirm `obw/core-profiler`.`plugins` does not have `woocommerce-payments`
9. Go to `WooCommerce -> Settings -> Payments`
10. Inspect a request to ` /wp-json/wc-admin/payment-gateway-suggestions`
11. Confirm the response doesn't have `woocommerce_payment` and `woocommerce_payments:with-in-person-payments`

**Get paid with Square task**
1. Install WooCommerce Square plugin (this will be installed automatically once we deploy WPCOM changes).
2. Go to `WooCommerce -> Home`
4. Confirm `Get paid with Square` task is the first task.
5. Click on it.
6. You should be redirected to Square connect wizard.

### Other information:

-   [ ] Have you added an explanation of what your changes do and why you'd like us to include them?
-   [ ] Have you written new tests for your changes, as applicable?

### FOR PR REVIEWER ONLY:

-   [ ] I have reviewed that everything is sanitized/escaped appropriately for any SQL or XSS injection possibilities. I made sure Linting is not ignored or disabled.

PR #1491, opened 2024-07-16
Update Onbarding steps for PayPal

epeicher:

Closes https://github.com/Automattic/dotcom-forge/issues/8277

This PR  adds `Get paid with PayPal` task and removes Woo Payments task when the option`woocommerce_onboarding_profile.partner` is `paypal`

CleanShot 2024-07-18 at 19 40 36@2x

**Get paid with PayPal task**
1. Create a free trial eCommerce site for Development as indicated in these instructions peapX7-1D4-p2
1. Install WooCommerce PayPal plugin (this will be installed automatically if the site is created through this flow https://wordpress.com/setup/entrepreneur/start?partnerBundle=paypal, but it will not be a Developer site).
2. Update the `woocommerce_onboarding_profile` option using the following command `wp option patch insert woocommerce_onboarding_profile partner paypal` (or `update` if the `partner` property exists).
3. Synchronize this branch with the Atomic site using the command in the parent task [here](https://github.com/Automattic/dotcom-forge/issues/8277#issuecomment-2230849198), changing `<your-domain>` with the domain of your host. **Or** you can use the following instructions to synchronize the `wc-calypso-bridge` plugin to the atomic site: Pe5pgL-2eX-p2
4. Go to `WooCommerce -> Home`
5. Confirm `Get paid with PayPal` task exists.
6. Click on it.
7. You should be redirected to PayPal connect wizard.

### Other information:

### FOR PR REVIEWER ONLY:

### DIFF
--- a/includes/free-trial/partners/class-wc-calypso-bridge-partner-paypal.php
+++ b/includes/free-trial/partners/class-wc-calypso-bridge-partner-paypal.php
@@ -108,9 +108,8 @@ class WC_Calypso_Bridge_Partner_PayPal {
             }
 
             try {
-                // TODO: Find a way to avoid hardcoding the URL.
-                // $params['paypal_connect_url'] = \WooCommerce\PayPal\Plugin::instance()->get_connection_handler()->get_connect_url();
-                $params['paypal_connect_url'] = admin_url( 'admin.php?page=wc-settings&tab=checkout&section=ppcp-gateway&ppcp-tab=ppcp-connection' );
+                $connection_tab_id = \WooCommerce\PayPalCommerce\WcGateway\Settings\Settings::CONNECTION_TAB_ID;
+                $params['paypal_connect_url'] = admin_url( 'admin.php?page=wc-settings&tab=checkout&section=ppcp-gateway&ppcp-tab=' . $connection_tab_id );
             } catch (\Throwable $e) {
                 // Fallback to the settings page
                 $params['paypal_connect_url'] = add_query_arg( array(
--- a/includes/free-trial/partners/class-wc-calypso-bridge-partner-paypal.php
+++ b/includes/free-trial/partners/class-wc-calypso-bridge-partner-paypal.php
@@ -34,10 +34,9 @@ class WC_Calypso_Bridge_Partner_PayPal {
      */
     public function __construct() {
         // Only for free trials.
-        // TODO: Uncomment this line when the free trial plan is available.
-        // if ( ! wc_calypso_bridge_is_ecommerce_trial_plan() ) {
-        //     return;
-        // }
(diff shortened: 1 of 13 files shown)


epeicher:
Hi @ilyasfoo , @moon0326 , I have created this PR based on https://github.com/Automattic/wc-calypso-bridge/pull/1426 but I have updated the testing instructions a bit as I was not able to reproduce them exactly. I have removed the `Feed Removal` section.

Could you please review the PR and let me know any comments? Please note that I am not very familiar with this codebase so be thorough in your review 😉

moon0326 (approved):
Nice work @epeicher 👍 

I think we should make a shared class for `remove_woo_payments_from_payments_suggestions_feed` and `remove_payments_note` functions when we get more payment tasks later on.

Other than that, LGTM and tested well 🚀

JulianBustamante:
Thank you, @moon0326 for reviewing.

> I think we should make a shared class for remove_woo_payments_from_payments_suggestions_feed and remove_payments_note functions when we get more payment tasks later on.

Agree! We can work on this on a follow-up task.

includes/free-trial/partners/class-wc-calypso-bridge-partner-paypal.php
```diff
@@ -0,0 +1,163 @@
...
+
+        return self::$instance;
+    }
+
+    /**
+     * Constructor.
+     */
+    public function __construct() {
+        // Only for free trials.
+        if ( ! wc_calypso_bridge_is_ecommerce_trial_plan() ) {
```
cpapazoglou, line 37: I think we should avoid using the `wc_calypso_bridge_is_ecommerce_trial_plan` condition. In the future the plan provisioned might not be a trial. Also, the behaviour should only be dictated by the `$onboarding_profile['partner']`.

epeicher: That's a key point to consider. We included that condition to avoid diverging from the original implementation, but I soon realized that it would need an update if the creating flow changes at any point, which is indeed very likely. And I totally agree with you that it should only depend on the `onboarding_profile` option.

What do you think about considering this when updating the process to be more dynamic, context here: p1721665079790669-slack-C02JPCHEKDY? Or do you think that we should instead remove this condition before shipping this iteration?

cpapazoglou: We can do it in a clean up PR.

epeicher: Task created to avoid forgetting this: 
- https://github.com/Automattic/dotcom-forge/issues/8413

Issue #1498, opened 2024-08-12
[LYS] Replace WPCOM coming soon page with LYS coming soon page


ilyasfoo:
### Issue description

P2: https://wp.me/pfwcF5-Cs

Replace WordPress.com's coming soon page with LYS coming soon pages.

We can use the filter [`a8c_show_coming_soon_page`](https://github.com/Automattic/jetpack/blob/065c5cbec110192aeef1bfb37124a2351251a00e/projects/packages/jetpack-mu-wpcom/src/features/coming-soon/coming-soon.php#L40C25-L40C50) to override WPCOM's coming soon status and apply LYS instead.

### Acceptance criteria

1. When in coming soon mode either on LYS or WPCOM, show WooCommerce's coming soon template.

PR #1500, opened 2024-08-21
Hide WPCOM coming soon page when lys feature flag is enabled

chihsuan:

Closes https://github.com/Automattic/wc-calypso-bridge/issues/1498.

Use the filter [a8c_show_coming_soon_page](https://github.com/Automattic/jetpack/blob/065c5cbec110192aeef1bfb37124a2351251a00e/projects/packages/jetpack-mu-wpcom/src/features/coming-soon/coming-soon.php#L40C25-L40C50) to override WPCOM's coming soon status so that we can show LYS coming soon page instead.

There a another task to [sync WPCOM coming soon status to LYS](https://github.com/woocommerce/team-ghidorah/issues/344). When that is done, we should see WooCommerce's coming soon template shown, when in coming soon mode either on LYS or WPCOM.

1. Use a WPCOM atomic e-commerce site and local wc-calypso-bridge
2. Enable LYS feature flag (via beta tester or you can remove [these lines](https://github.com/Automattic/wc-calypso-bridge/blob/507bde7b933c60bb2cb7efdfad66b7bd8581b39b/includes/class-wc-calypso-bridge-woocommerce-admin-features.php#L143-L147).
3. Go to `https://wordpress.com/settings/general/SITE_URL` and enable WPCOM coming soon
4. Go to `/wp-admin/admin.php?page=wc-settings&tab=site-visibility` and enable LYS coming soon
5. Open the site frontend in an incognito window and confirm that the LYS coming soon page is shown

### Other information:

### FOR PR REVIEWER ONLY:

### DIFF
--- a/includes/class-wc-calypso-bridge-coming-soon.php
+++ b/includes/class-wc-calypso-bridge-coming-soon.php
@@ -37,7 +37,7 @@ class WC_Calypso_Bridge_Coming_Soon {
 	 * Constructor.
 	 */
 	public function __construct() {
-		add_filter( 'a8c_show_coming_soon_page', array( $this, 'should_show_a8c_coming_soon_page' ), 99999, 1 );
+		add_filter( 'a8c_show_coming_soon_page', array( $this, 'should_show_a8c_coming_soon_page' ), PHP_INT_MAX, 1 );
 	}
 
 	/**
--- a/includes/class-wc-calypso-bridge-coming-soon.php
+++ b/includes/class-wc-calypso-bridge-coming-soon.php
@@ -41,18 +41,19 @@ class WC_Calypso_Bridge_Coming_Soon {
 	}
 
 	/**
-	 * Show the coming soon page if the Launch Your Store feature is enabled.
+	 * Hide the a8c coming soon page if the Launch Your Store feature is enabled.
 	 *
 	 * @param bool $should_show
 	 * @return bool
 	 */
 	public function should_show_a8c_coming_soon_page( $should_show ) {
-		if ( ! Features::is_enabled( 'launch-your-store' ) ) {
-			// Bail out early if the Launch Your Store feature is not enabled.
-			return $should_show;
+		if (
+			class_exists( '\Automattic\WooCommerce\Admin\Features\Features' ) && \Automattic\WooCommerce\Admin\Features\Features::is_enabled( 'launch-your-store' )
+		) {
+			return false;
 		}
 
-		return false;
+		return $should_show;
 	}
 }
 
--- a/includes/class-wc-calypso-bridge-coming-soon.php
+++ b/includes/class-wc-calypso-bridge-coming-soon.php
@@ -1,7 +1,5 @@
 <?php
 
-use Automattic\WooCommerce\Admin\Features\Features;
-
 defined( 'ABSPATH' ) || exit;
 
 /**
--- a/includes/class-wc-calypso-bridge-coming-soon.php
+++ b/includes/class-wc-calypso-bridge-coming-soon.php
@@ -12,7 +12,7 @@ defined( 'ABSPATH' ) || exit;
  *
  * Handle Coming Soon mode.
  */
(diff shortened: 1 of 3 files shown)


includes/class-wc-calypso-bridge-coming-soon.php
```diff
@@ -0,0 +1,59 @@
...
+		}
+
+		return self::$instance;
+	}
+
+	/**
+	 * Constructor.
+	 */
+	public function __construct() {
+		add_filter( 'a8c_show_coming_soon_page', array( $this, 'should_show_a8c_coming_soon_page' ), 99999, 1 );
```
ilyasfoo, line 40: I think typically we would use this if we intend to run it last

```suggestion
		add_filter( 'a8c_show_coming_soon_page', array( $this, 'should_show_a8c_coming_soon_page' ), PHP_INT_MAX, 1 );
```

ilyasfoo (approved):
Tests well, thanks @chihsuan! We'll also need to update the version in `class-wc-calypso-bridge.php` to `x.x.x`.

I have minor comment but LGTM!

includes/class-wc-calypso-bridge-coming-soon.php
```diff
@@ -0,0 +1,59 @@
...
+	}
+
+	/**
+	 * Show the coming soon page if the Launch Your Store feature is enabled.
+	 *
+	 * @param bool $should_show
+	 * @return bool
+	 */
+	public function should_show_a8c_coming_soon_page( $should_show ) {
+		if ( ! Features::is_enabled( 'launch-your-store' ) ) {
```
ilyasfoo, line 50: I forgot, there was some instances where `Features` class was not available during a run. I'm unsure how early `should_show_a8c_coming_soon_page` is run, but the prior art is to check for the class too: https://github.com/Automattic/wc-calypso-bridge/blob/3233649621c32b4f7f18daf68ee381f90a7df970/includes/class-wc-calypso-bridge-ecommerce-admin-menu.php#L380

What do you think?

moon0326 (commented):
LGTM and tested well 👍 🚀 

I'll let @ilyasfoo to approve it.

includes/class-wc-calypso-bridge-coming-soon.php
chihsuan: That's a good idea. Updated in  3038da3be2d72c9b5866676a94289233cbb528bc

chihsuan: I wonder if there is anything we can do to prevent any future issues like this.  🤔  We also use `Features::is_enabled` directly here, though it should be safe to use it as it is in the setup task context. In other places, we use methods/classes from WooCommerce core without checking their availability. 

https://github.com/Automattic/wc-calypso-bridge/blob/717d11285b7b770ee7dbbc6e8251d7646ade4dfc/includes/class-wc-calypso-bridge-setup-tasks.php#L147-L153

  I'd also like to know why the Features class was unavailable during the run.

ilyasfoo (approved):
Thanks @chihsuan! The change tests well.

includes/class-wc-calypso-bridge-coming-soon.php
ilyasfoo: Good question, @chihsuan! I think it's one of the pitfalls of WordPress' filters. I believe in some cases the filter ran before WooCommerce initialized.

I think a good solution might be to create a class helper that runs a class callback after WooCommerce has initialized, but I'm not sure how that'll look

ilyasfoo: To be accurate, it's very much possible that the `should_show_a8c_coming_soon_page` filter is safe. I was just trying to be guarantee a safe execution, since it's possible for other mu-plugins to do a `apply_filter( 'should_show_a8c_coming_soon_page' )` before WooCommerce is loaded

chihsuan: Good point! Agreed that it's better to be safe.

PR #1511, opened 2024-09-10
Fix logic for disabling lys task on trial sites

ilyasfoo:
Quick fix on displaying LYS task. Currently, it seems LYS task is shown in ecommerce trial plan sites.

#### Business trial plan:
image

#### Ecommerce trial plan:
image

#### Ecommerce plan:
image


ilyasfoo:
Merging without review since it's simple enough

PR #1529, opened 2024-10-23
Fix coming soon unintentionally expose the rest of the site

### LINKED ISSUE #1528: WooCommerce coming soon unintentionally expose the rest of the site
### Issue Description

When the following conditions are true, WooCommerce coming soon "Apply to store pages only" will unintentionally publish the rest of the site:

- WordPress.com logged in user is at least 30 days old
- Using a business site plan
- Installing WooCommerce via plugins menu (either upload or via plugin search)
- Going through core profiler (either skip or complete it)

#### To reproduce:

1. Set up a business site
2. In an incognito window, go to the site's frontend and ensure site is in coming soon mode
3. Activate hosting features by going to `https://wordpress.com/hosting-config/SITE_URL`
4. Go to Plugins and install WooCommerce plugin (don't use the `WooCommerce` menu from the sidebar)
5. Click on WooCommerce and observe you're redirected to core profiler
6. Skip core profiler
7. Refresh the incognito and observe it's no longer hidden under coming soon mode

### Acceptance criteria

1. When going through setup wizard, it shouldn't apply coming soon mode to store pages if WPCOM is in coming soon mode.

chihsuan:

Closes https://github.com/Automattic/wc-calypso-bridge/issues/1528.

This PR modifies the handling of WooCommerce's coming soon initialization endpoint to respect WordPress.com's site-wide Coming Soon status. Key changes:

- Added a check to detect if WordPress.com Coming Soon mode is active
- If active, we return `true` immediately to prevent WooCommerce from modifying the site's visibility
- Only set the `woocommerce_store_pages_only` option when the site is already live

#### Test with Coming Soon mode enabled

1. Set up a WordPress.com business site with Coming Soon mode enabled
2. Install WooCommerce through the plugins menu
3. Go through the core profiler (either skip or complete)
4. Verify that the site remains in Coming Soon mode after the process

#### Test with Coming Soon mode disabled

1. Set up a WordPress.com business site with Coming Soon mode disabled
2. Install WooCommerce through the plugins menu
3. Go through the core profiler (either skip or complete)
4. Verify that site is in Coming Soon mode with Store Pages only

### Other information:

### FOR PR REVIEWER ONLY:

### DIFF
--- a/includes/class-wc-calypso-bridge-coming-soon.php
+++ b/includes/class-wc-calypso-bridge-coming-soon.php
@@ -6,7 +6,7 @@ defined( 'ABSPATH' ) || exit;
  * Class WC_Calypso_Bridge_Coming_Soon
  *
  * @since   2.6.0
- * @version 2.7.1
+ * @version x.x.x
  *
  * Handle Coming Soon mode.
  */
--- a/includes/class-wc-calypso-bridge-coming-soon.php
+++ b/includes/class-wc-calypso-bridge-coming-soon.php
@@ -41,6 +41,7 @@ class WC_Calypso_Bridge_Coming_Soon {
 		add_filter( 'pre_update_option_woocommerce_coming_soon', array( $this, 'override_update_woocommerce_coming_soon' ), 10, 2 );
 		// Admin bar menu is not only shown in the admin area but also in the front end when the admin user is logged in.
 		add_action( 'admin_bar_menu', array( $this, 'remove_site_visibility_badge' ), 32 );
+		add_filter( 'rest_pre_dispatch', array( $this, 'handle_initial_coming_soon_endpoint' ), 10, 3 );
 
 		if ( is_admin() ) {
 			add_filter( 'plugins_loaded', array( $this, 'maybe_add_admin_notice' ) );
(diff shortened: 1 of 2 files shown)


chihsuan:
> This looks like a good solution, @chihsuan. I could see other hosts having a similar problem in the future if they integrate with coming soon mode. Perhaps we should move the logic to the core rest api and solve it generally? 🤔

Good idea! I think we can move the coming soon mode check and return 200 logic to the core reset api. But we probably need to keep the logic for setting `woocommerce_store_pages_only` option in the host, since only they know how to determine if the site is already live for their environment. cc @ilyasfoo if you have any thoughts on this. 🙏

ilyasfoo (approved):
Tests well, nice work @chihsuan!

ilyasfoo:
> Perhaps we should move the logic to the core rest api and solve it generally? 🤔

@adrianduffell  Good point, I think so too.

@chihsuan  I think to provide the most flexibility, we could add 2 filters for both of these just before we set them:
 
https://github.com/woocommerce/woocommerce/blob/3efa429d9862d347438ab6ba0aa7a7ebe438049e/plugins/woocommerce/src/Admin/API/LaunchYourStore.php#L136-L137

chihsuan:
> I think to provide the most flexibility, we could add 2 filters for [both of these](https://github.com/woocommerce/woocommerce/blob/3efa429d9862d347438ab6ba0aa7a7ebe438049e/plugins/woocommerce/src/Admin/API/LaunchYourStore.php#L136-L137) just before we set them. It should allow for hosts to determine the default coming soon state and store-only mode after going through core profile

Sounds good! I'll create a follow-up issue for this.

adrianduffell (approved):
Code change makes sense and LGTM! Thanks @chihsuan 🚀

PR #1597, opened 2026-02-23
i18n: Change the textdomain for strings which only exist in wc-calypso-bridge

dlind1:

Related to TRA-33.
There are several translatable strings in wc-calypso-bridge which have translations, but they're not being used because the wrong textdomain is used when displaying them. This PR fixes all the instances where `woocommerce` textdomain was used instead of `wc-calypso-bridge`, when the string only exists in wc-calypso-bridge.

1. One option would be to go to /wp-admin/admin.php?page=wc-admin on a new site with WooCommerce installed and confirm that `Add your products. Show off your products...` appears translated.
2. Another option is to review the changes and do some spot checks to confirm that the strings only appear in https://translate.wordpress.com/projects/woocommerce/wc-calypso-bridge/de/default/, but not in https://translate.wordpress.org/projects/wp-plugins/woocommerce/stable/fr/default/.

### Other information:

### FOR PR REVIEWER ONLY:

### DIFF
--- a/includes/class-wc-calypso-bridge-free-trial-payment-restrictions.php
+++ b/includes/class-wc-calypso-bridge-free-trial-payment-restrictions.php
@@ -294,7 +294,7 @@ class WC_Calypso_Bridge_Free_Trial_Payment_Restrictions {
 			?>
 			<script type="text/javascript">
 				function overrideNoPaymentMethodsMessage( translation, text, domain ) {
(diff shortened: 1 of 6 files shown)


bor0 (commented):
> This PR fixes all the instances where `woocommerce` textdomain was used instead of `wc-calypso-bridge`, when the string only exists in wc-calypso-bridge.

I noticed some strings, e.g. in class-wc-calypso-task-add-domain.php, are not changed:

```
bor0:~/dev/wc-calypso-bridge$ grep "2 minutes" includes/tasks/class-wc-calypso-task-add-domain.php
		return __( '2 minutes', 'woocommerce' );
bor0:~/dev/wc-calypso-bridge$ grep "2 minutes" ~/Downloads/woocommerce-wc-calypso-bridge-de.po 
msgid "2 minutes"
```

bor0 (approved):

PR #1599, opened 2026-03-26
Fix PHP warnings: use current() after array_filter, guard foreach against null

bor0:

Fix two PHP warnings fired ~4000 times/day on WP Cloud sites from the `automattic/wc-calypso-bridge` package.

- **`choose-domain.php`**: `array_filter()` preserves original array keys, so `$plan_purchases[0]` returns `null` when the bundle purchase is not at index 0. Replaced with `current( $plan_purchases )` which always returns the first element regardless of key.
- **`class-wc-calypso-bridge-woocommerce-admin-features.php`**: `filter_woocommerce_settings_pages_order()` can receive a non-array value from upstream. Added an `is_array()` guard before the `foreach` to prevent the warning.

Fixes WCCOM-2455.

**On trunk (before) — reproduce the warnings:**

1. On a WP Cloud site with the `wc-calypso-bridge` package vendored via `wpcomsh`, open the PHP error log.
2. Confirm you see recurring warnings:
   - `PHP Warning: Attempt to read property "subscribed_date" on null in .../class-wc-calypso-bridge-choose-domain.php on line 114`
   - `PHP Warning: Undefined array key 0 in .../class-wc-calypso-bridge-choose-domain.php on line 114`
   - `PHP Warning: foreach() argument must be of type array|object, null given in .../class-wc-calypso-bridge-woocommerce-admin-features.php on line 244`

**On this branch (after):**

1. With this version deployed, navigate to **WP Admin → WooCommerce → Settings** on a site that has a paid bundle plan purchase where the plan is not the first entry in the purchases array (e.g., a site with multiple purchases where the bundle is at a non-zero index).
2. Confirm the WooCommerce Settings page loads without PHP warnings.
3. Check the PHP error log — no new warnings should appear from either `class-wc-calypso-bridge-choose-domain.php` or `class-wc-calypso-bridge-woocommerce-admin-features.php`.
4. Navigate to **WP Admin → WooCommerce → Notes** (Inbox) and confirm the Choose Domain note still renders correctly and the `subscribed_date` logic works (note marked as unread if plan was purchased within the last 2 hours).

### Other information:

### FOR PR REVIEWER ONLY:

### DIFF
--- a/includes/notes/class-wc-calypso-bridge-choose-domain.php
+++ b/includes/notes/class-wc-calypso-bridge-choose-domain.php
@@ -111,7 +111,7 @@ class WC_Calypso_Bridge_Choose_Domain_Note {
 		}
 
 		// Calculate how many hours have passed since the paid plan purchase.
-		$subscribed_date = strtotime( get_date_from_gmt( $plan_purchases[0]->subscribed_date ) );
+		$subscribed_date = strtotime( get_date_from_gmt( current( $plan_purchases )->subscribed_date ) );
 		$current_date    = current_time( 'timestamp', true );
 		$hours_passed    = ( $current_date - $subscribed_date ) / HOUR_IN_SECONDS;
 
--- a/includes/notes/class-wc-calypso-bridge-choose-domain.php
+++ b/includes/notes/class-wc-calypso-bridge-choose-domain.php
@@ -111,7 +111,7 @@ class WC_Calypso_Bridge_Choose_Domain_Note {
 		}
 
 		// Calculate how many hours have passed since the paid plan purchase.
-		$subscribed_date = strtotime( get_date_from_gmt( current( $plan_purchases )->subscribed_date ) );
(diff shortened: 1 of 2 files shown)


includes/notes/class-wc-calypso-bridge-choose-domain.php
```diff
@@ -111,7 +111,7 @@ function ( $purchase ) {
 		}
 
 		// Calculate how many hours have passed since the paid plan purchase.
-		$subscribed_date = strtotime( get_date_from_gmt( current( $plan_purchases )->subscribed_date ) );
+		$subscribed_date = strtotime( get_date_from_gmt( reset( $plan_purchases )->subscribed_date ) );
```
bor0: In theory, even `current` won't be a problem because we use `array_filter` which generates a new array and nothing mutates between. But, I think `reset` is safer in case someone changes the code in the future

mcliwanow (approved):
approving based on code review and discussions on Slack

m1r0:
LGTM as well! 👍 

Better late than never, I guess. 🙂

PR #1559, opened 2025-05-18
Updated the default WooCommerce jetpack module list

paulopmt1:
### Changes proposed in this Pull Request:

Original Slack Thread: p1744052983694779-slack-CDLH4C1UZ
[Linear task](https://linear.app/a8c/issue/DOTCOM-13242/jetpack-block-module-deactivates-after-commerce-migration)

**TLDR**: Some Jetpack modules (like blocks) are disabled after a `Business` > `Commerce` upgrade. Because of that, those sites won't have the Jetpack blocks activated:

image

**Solution**: Here, we're updating the list of active modules we expect to have enabled after that migration. This was originally implemented [on this PR](https://github.com/Automattic/wc-calypso-bridge/pull/844), and here's more on why: peapX7-yd-p2

**Why this happens?**
It's related to the WooCommerce plugin initialization. Turns out [we have a list of Jetpack modules](https://github.com/Automattic/wc-calypso-bridge/blob/master/includes/class-wc-calypso-bridge-setup.php#L514-L544) we override when starting a WooCommerce store.

**More context**
I first tried to solve this problem with these two approaches, which partially fixed it, but fortunately, this new PR is the correct approach as we're fixing the root cause of the issue. The following PRs will be overridden after merging the current PR:

- https://github.com/Automattic/jetpack/pull/42971
- 181039-ghe-Automattic/wpcom

### How to test the changes in this Pull Request:

- [x] Have you followed the [Writing high-quality testing instructions guide](https://github.com/woocommerce/woocommerce/wiki/Writing-high-quality-testing-instructions)?

I'm unsure how to test the changes using this repository. But I'll describe how to test them on an Atomic site:

1. Create a new site using the Business plan
2. Add it as a dev site on [wpcom dev blogs](https://mc.a8c.com/atomic/wpcom-dev-blogs/)
3. Convert it to Atomic by enabling the Hosting features
4. SSH into it the new server
5. Edit the source file, at line 543, with `vim /htdocs/wp-content/mu-plugins/wpcomsh/vendor/automattic/wc-calypso-bridge/includes/class-wc-calypso-bridge-setup.php`
6. Add the three extra modules to the `$active_modules` array: `'blocks', 'blaze', 'account-protection'`
	image
8. Open that site's plans page: https://wordpress.com/plans/{blogId}
9. Buy the Commerce plan
10. Wait for the migration to complete (ensure you see WooCommerce at the /wp-admin):
	image
12. Create a new post
13. Ensure you still see Jetpack blocks:
[...]

PR #1642, opened 2026-08-14
Enable the Jetpack Stats module by default on WoW stores

kangzj:

[Linear task](https://linear.app/a8c/issue/WOOA7S-1741/site-visitor-stats-are-missing-on-wow-sites)

**TLDR**: Site visitor stats are missing on WooCommerce on WordPress.com (WoW) stores. There is no Jetpack Stats entry in wp-admin, and the `Analytics` menu is Woo-specific (products, orders, etc.), so there is nowhere to see visitor stats at all.

**Solution**: Add `stats` to the list of Jetpack modules we enable by default when a store is set up. Same approach as #1559, which added `blocks`, `blaze` and `account-protection`.

**Why this happens?**
When a brand new store initializes, [we override the whole `jetpack_active_modules` option](https://github.com/Automattic/wc-calypso-bridge/blob/master/includes/class-wc-calypso-bridge-setup.php#L514-L547) with our own list. `stats` has never been in that list, so the Stats module ends up deactivated and its wp-admin entry never appears.

As with #1559, this is easiest to verify on an Atomic site rather than from this repository:

1. Create a new site on the Business plan.
2. Add it as a dev site using the internal wpcom dev blogs tool on MC.
3. Convert it to Atomic by enabling the Hosting features.
4. SSH into the new server.
5. Edit `/htdocs/wp-content/mu-plugins/wpcomsh/vendor/automattic/wc-calypso-bridge/includes/class-wc-calypso-bridge-setup.php` and add `'stats',` to the `$active_modules` array (after `'account-protection'`).
6. Open that site's plans page: `https://wordpress.com/plans/{blogId}` and buy the Commerce plan.
7. Wait for the migration to complete (WooCommerce should be visible in /wp-admin).
8. Confirm the Stats entry is present in wp-admin and shows site visitor stats.
9. Also confirm `get_option( 'jetpack_active_modules' )` contains `stats` — e.g. via WP-CLI: `wp option get jetpack_active_modules`.

**Before**: no visitor stats anywhere in wp-admin — only the Woo-specific Analytics menu.
**After**: the Jetpack Stats screen is available on a freshly created store.

### Other information:

### FOR PR REVIEWER ONLY:

### DIFF
--- a/includes/class-wc-calypso-bridge-setup.php
+++ b/includes/class-wc-calypso-bridge-setup.php
@@ -544,6 +544,7 @@ class WC_Calypso_Bridge_Setup {
 				'blocks',
 				'blaze',
 				'account-protection',
+				'stats',
 			);
 
 			$sharing_options = array(


chihsuan (approved):
Thanks @kangzj LGTM! 🚀

PR #1643, opened 2026-08-21
Treat stock thresholds as integers in product count queries

rcrdortiz:

The `data/counts` product counts compare stock levels against the low-stock and out-of-stock thresholds. Both thresholds are stock levels, but they were read from options and interpolated into the count queries as strings, so a stored value that is not a plain integer produced a comparison that was not numeric.

This coerces both thresholds with `absint()` at the read and binds them as integers in the two `COUNT` statements, so the comparisons stay numeric regardless of what the stored option value looks like.

Behaviour on the normal path is unchanged: an unset option coerces to `0`, which matches what the previous empty-string comparison already evaluated to.

Also adds unit coverage for the product counts, which had none.

1. On a store-on-wpcom site, create three published products with **Manage stock** enabled, with stock quantities of `0`, `1` and `50`.
2. Go to **WooCommerce → Settings → Products → Inventory** and set **Out of stock threshold** to `0` and **Low stock threshold** to `2`. Save.
3. Request `GET /wp-json/wc/v3/data/counts` as a user who can manage WooCommerce. Confirm the response contains `"all": 3`, `"out-of-stock": 1` and `"low-inventory": 1`.
4. Set **Low stock threshold** to `10`, save, and repeat the request. Confirm `"low-inventory"` is now `2`, since the product with a stock of `1` and the one with `50` are both at or below the new threshold... (verify against your own fixtures if you use different quantities).
5. Set **Low stock threshold** to a value that is not a plain integer, for example `2 items`, `2.9`, or leave it empty. Save, then repeat the request.
6. Confirm the endpoint still returns a well-formed response with sensible counts, and that no database errors or PHP notices appear in the logs.

### Other information:

-   [x] Have you written new tests for your changes, as applicable?

New unit tests are added in `store-on-wpcom/tests/unit-tests/data-counts-controller.php`, covering the counts against integer, malformed and non-numeric threshold values, plus the permission boundaries on the route.

Note that the `store-on-wpcom` suite appears dormant and I was not able to run it: its `phpunit.xml` still uses the `syntaxCheck` attribute that was removed in PHPUnit 8, the existing cases use the pre-PHPUnit 8 `setUp()` signature, and the test jobs in `.github/workflows/ci.yml` are commented out. The new cases follow the conventions of the existing files in that directory, so they will need the same treatment as the rest of the suite whenever it is revived.

### DIFF
--- a/store-on-wpcom/api/class-wc-calypso-bridge-data-counts-controller.php
+++ b/store-on-wpcom/api/class-wc-calypso-bridge-data-counts-controller.php
@@ -101,11 +101,15 @@ class WC_Calypso_Bridge_Data_Counts_Controller extends WC_REST_Controller {
 				AND posts.post_status = 'publish'"
 		);
 
-		$low_stock_amount = get_option( 'woocommerce_notify_low_stock_amount' );
-		$no_stock_amount = get_option( 'woocommerce_notify_no_stock_amount' );
+		// Both thresholds are stock levels, so coerce them to integers and bind them
+		// as integers below. This keeps the comparisons numeric even when an option
+		// holds an unexpected value.
+		$low_stock_amount = absint( get_option( 'woocommerce_notify_low_stock_amount' ) );
+		$no_stock_amount = absint( get_option( 'woocommerce_notify_no_stock_amount' ) );
 
 		$counts['out-of-stock'] = (int) $wpdb->get_var(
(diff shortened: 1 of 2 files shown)


paulopmt1 (approved):
It fixes the issue as expected. Thanks @rcrdortiz!

PR #1646, opened 2026-09-01
Retire stable WooCommerce Admin feature flags

gigitux:

WooCommerce 11.1 retires stable WooCommerce Admin feature flags and introduces compatibility shims that log warnings when deprecated feature flags are used ([source code](https://github.com/woocommerce/woocommerce/blob/ac3757ce09bd1f364713e6df7985d2f682302ec9/plugins/woocommerce/src/Admin/Features/Features.php#L350-L364)). This PR introduces a version-aware compatibility boundary so the bridge continues to support its declared WooCommerce 7.3+ range without triggering deprecated APIs on newer versions.

Upstream context:

- https://github.com/woocommerce/woocommerce/pull/65472
- https://developer.woocommerce.com/2026/08/31/retiring-stable-feature-flags/

### Other information:

- [x] The changes and rationale are explained above.
- [x] No repository tests were added because the current test environment does not provide a practical way to run multiple WooCommerce versions in one process.

### FOR PR REVIEWER ONLY:

- [ ] I have reviewed that everything is sanitized/escaped appropriately for any SQL or XSS injection possibilities. I made sure Linting is not ignored or disabled.


rcrdortiz (approved):
LGTM!

includes/class-wc-calypso-bridge-helper-functions.php
```diff
@@ -13,6 +13,59 @@
...
+	 * WooCommerce versions that deprecated stable WooCommerce Admin feature flags.
+	 *
+	 * Keep the non-optional entries in sync with the retired feature
+	 * compatibility metadata in WooCommerce's Admin Features class.
+	 *
+	 * @see https://github.com/woocommerce/woocommerce/blob/trunk/plugins/woocommerce/src/Admin/Features/Features.php
+	 *
+	 * @var array<string, string>
+	 */
+	private const WC_ADMIN_STABLE_FEATURE_FLAG_DEPRECATION_VERSIONS = array(
```
Aljullu, line 25: Is there a reason to list all features, even if we are not checking them anywhere in `wc-calypso-bridge? For example, if `experimental-fashion-sample-products` is never checked, is there a point on defining it here? My concern is that this will likely get out of sync with WC core.

includes/class-wc-calypso-bridge-helper-functions.php
```diff
@@ -38,6 +91,72 @@ private function __construct() {
...
+		if (
+			null !== $deprecation_version &&
+			null !== $woocommerce_version &&
+			version_compare( $woocommerce_version, $deprecation_version, '>=' )
+		) {
+			return true;
+		}
+
+		return class_exists( '\Automattic\WooCommerce\Admin\Features\Features' ) &&
+			\Automattic\WooCommerce\Admin\Features\Features::is_enabled( $feature );
```
Aljullu, line 110: Similar to my comment above. Do we really need this fallback? I would avoid `is_wc_admin_feature_enabled()` becoming the go-to way to check if a WC feature is enabled. Instead, I would limit it to returning the appropriate values for `launch-your-store` and `customize-store`. But nothing else. Maybe I'm overthinking it, though.

Aljullu (approved):
I'm not super familiar with wc-calypso-bridge, but I took a look at the code and it makes sense to me. Left two minor comments but not a strong opinion from my side. I'm fine either way.

includes/class-wc-calypso-bridge-helper-functions.php
gigitux: Yes, because in case the store is lower than WooCommerce 11.1, we need to run the old logic: does it make sense?

includes/class-wc-calypso-bridge-helper-functions.php
gigitux: Good point. I thought it might be easier to maintain this way, but yeah, I agree with you. I’m going to keep only the features that are present in wc-calypso-bridge.

gigitux: Fixed with 34ddb914c1bb82b280d2aac86c11b590e3c1fb7e

includes/class-wc-calypso-bridge-helper-functions.php
Aljullu: Ah yes, good point. We need to keep this. Ignore this comment. :smile:

PR #1651, opened 2026-09-16
Make the legacy BACS detector log queryable

vladolaru:

The detector added in 2.11.9 writes its line with `error_log()`. The platform log index stores that message but does not index it, so nothing can query or alert on it — which was the whole point of adding the detector.

**What changed**

- The message now ends with ` in <file> on line <n>`, built from `__FILE__` and `__LINE__`. The log pipeline parses that suffix and fills the indexed `file` and `line` fields, so an alert can match this file's path.

**Why not `trigger_error()`**

- It would also populate `file`, but it raises a real PHP warning: it can print into a response on a site running with `display_errors` on, and it adds to the warning severity the platform rate-limits hardest. The suffix gets the same indexed fields while staying silent, at `severity: User`.

**Verified on a WoA site**

- Ran this code on a Business-plan dev site and sent an unauthenticated request carrying `settings.accounts`. The request got its `401`, nothing was saved, and the log entry now carries `file`, `line`, `kind: mu-plugins` and `name: wpcomsh`. Before this change the same entry had `file: "unknown"` and could not be matched by any query.

This code only runs where the plugin loads its `store-on-wpcom` branch: a WordPress.com on WP Cloud (Atomic) site on a plan without the eCommerce managed-plugins feature, such as Business. The setup steps are the same as in #1648.

On such a site, with this branch synced over `wp-content/mu-plugins/wpcomsh/vendor/automattic/wc-calypso-bridge/`:

1. Send a request carrying the retired payload, with no credentials:

```sh
curl -s -o /dev/null -w "%{http_code}\n" -X POST "https://<site>/wp-json/wc/v3/payment_gateways/bacs" \
  -H "Content-Type: application/json" \
  -d '{"settings":{"accounts":[{"account_name":"Test"}]}}'
```

2. Confirm the response is `401` and that `wp option get woocommerce_bacs_accounts` is unchanged.
3. Open **Hosting → Logs → PHP errors** for the site and find the new entry. Confirm the **File** column shows `.../wc-calypso-bridge/store-on-wpcom/class-wc-calypso-bridge.php` and **Line** is populated. Before this change both were empty.
4. Confirm the account name from step 1 does not appear anywhere in the log entry.
5. Save a bank account normally under **Payments → Take offline payments → Direct bank transfer** and confirm no new log entry appears. Ordinary admin saves must stay silent.

### Other information:

-   [ ] Have you written new tests for your changes, as applicable? (Logging change only; verified manually on a WoA site.)

### FOR PR REVIEWER ONLY:

🤖 Generated with [Claude Code](https://claude.com/claude-code)


store-on-wpcom/class-wc-calypso-bridge.php
```diff
@@ -93,6 +93,12 @@ public function includes() {
 	 * never logged. Remove this method and its filter once the observation
 	 * window has passed with no hits.
 	 *
+	 * The message ends with ` in <file> on line <n>` on purpose. The platform log
+	 * index stores the message but does not index it, so a message alone cannot
+	 * be queried or alerted on. It does index the file and line it parses out of
+	 * that suffix, which is what lets an alert match this file's path. Keep the
+	 * suffix last, and keep `__FILE__` and `__LINE__` rather than a literal.
```
oaratovskyi: The reason for the suffix is useful, but this took me a moment to parse. Could we keep the key constraint and trim it to:

```suggestion
	 * The platform only indexes the file and line parsed from this trailing suffix.
	 * Keep it last and use `__FILE__` and `__LINE__` so alerts can match this file.
```

oaratovskyi (approved):
Nice fix, Vlad. I verified the detector behavior locally and checked the exact PR head. Everything looks good. I left one non-blocking suggestion to shorten the explanatory comment.

store-on-wpcom/class-wc-calypso-bridge.php
vladolaru: Trimmed, with one change: I kept "does not index the message" explicit rather than implied. That's the constraint that makes the suffix load-bearing — without it the line still reads fine to a human but becomes unmatchable by any query, which is how the 2.11.9 version got shipped.

